PR #34, opened 2017-04-21
Add Agolia search

adambutler:
kapture 2017-04-21 at 15 06 18

This PR is evaluating Agolia search for NDP with support for searching the Nexmo knowledge base as well. Findings:

- Implementation was super easy. I've only done `autocomplete` type search but adding full page search should not be a problem.
- Algolia has a [ZenDesk Help Center Connector](https://community.algolia.com/zendesk/) making it simple to have it automatically sync as often as every 5 min.
- Loading data was really easy, it just takes an object/hash from my search Rake task.
- I can't find a way of aggregating results between two Indices but I expect this to be possible.
- I can't find a way of excluding results based on an objects key. For example if `outdated` is `true` do not return the result.

Notes:

Many further improvements could be made including full text search for NDP content, this PR represents an early-stage evaluation of the tool and can be used as a resource for deciding if we should invest more time developing search with Algolia.


adambutler:
@leggetter @tommorris This is up on staging now, feedback appreciated.

Recent changes:

- Uses serverside Algolia implementation
- Replaced Algolia autocomplete.js with a custom React search component that is applied as a progressive enhancement.
- Added full page search.
- Added support for ignoring a list of facets. Being used to remove internal categories and sections.
- Removes the Elastic search dependency

I'm happy with the implementation "as is", however, we should work to ensure all documents have descriptions to improve the results. We could also look into improving the ranking from Algolia side with synonyms & facet ranking.

leggetter:
Makes it very easy to find information and is a huge improvement on the existing search over on docs.nexmo.com.

Feedback:

## Search UI on smaller widths should be updated

screen shot 2017-07-05 at 09 08 50

## Size

The search functionality feels a little apologetic since it's so small.

1. Size of input - could it be bigger. If we want people to use it then make it more visible. We should do this since it's useful.
2. Font-size - both in the input and in the search result dialog is very small. What does it look like when made a bit larger.
3. Size of dialog - I'm specifically searching for content so I don't mind if more of the page I'm on is covered with the search results. Can we make the area used for the search results bigger?

## Search dialog flashes as you search

The full dialog expands and shinks (to show the "working" cog) as you type. This results in a flashing effect. Is there something that we can do which removes this effect?

## Reduce prominence of knowledgebase 

I understand that some users will find the KB useful. But, as a developer who is in Nexmo Developer, I am mainly interested in Nexmo Developer content. I don't mind knowing that there are also some matching results in the KB, but I'd be happy to perform another action (a click?) in order to reveal them.

The fact the KB results are on the left, and I read left-to-right, initially confused me. I understand that I'm typing on the right of the screen to perform my search and that the Nexmo Developer results appear directly underneath. But it was still a little confusing initially. How about Nexmo Developer at the top and Knowledgebase at the bottom? Or a tabbed UI? I've seen Google search use "Group 1", "Group 2" and "Both" e.g. Tab: Nexmo Developer, Tab: Knowledgebase, Tab: Both.

_Note: If agreed, this could be an improvement for a future release_

adambutler:
> Size of input - could it be bigger. If we want people to use it then make it more visible. We should do this since it's useful.

People know to find search in the top right. I don't feel that making it larger would improve discoverability and would cause issues with the search boxes visual harmony with the subnavigation.

> Font-size - both in the input and in the search result dialog is very small. What does it look like when made a bit larger.
> Size of dialog - I'm specifically searching for content so I don't mind if more of the page I'm on is covered with the search results. Can we make the area used for the search results bigger?

This I agree with, the search text was indeed too small and after experimenting with the size of the results I've increased the area of the results as well as increased the font-size across all elements within quick search.

> Search dialog flashes as you search

I've fixed this so that continued queries rendering are deferred until they are loaded and replaced in place. Furthermore I've added some debouncing to limit subsequent requests to the server.

> Reduce prominence of knowledgebase

We can enrich the search more with filters, I agree that this should be opened as a new issue.

leggetter:
@adambutler could you please deploy to staging for a review? I think some ENV vars are missing on the review apps so I can't see the functionality.

adambutler:
@leggetter 👍 It's up

leggetter:
Is there any way of showing a very small snippet of the matching text within the text results? For example, here I'm looking for "machine detection" and although I do get results (and these are correct results 🙌) I'm missing the context of why these are the results.

screen shot 2017-07-06 at 10 29 49

I'll like a hint that this is why it's the right result:

screen shot 2017-07-06 at 10 31 42

adambutler:
@leggetter Added highlighting support

leggetter:
@adambutler this is a massive improvement over the existing search on docs. Fantastic stuff!

PR #134, opened 2017-06-19
Add a smart local numbers tutorial

cbetta:
Adds a tutorial for the "800 Number Replacement" use case.

@leggetter would love your feedback if the content is in line with what you expected.

@adambutler would love your feedback on if I created this the right way, as I'm still new to the system.

The full code for the tutorial is being uploaded as soon as @leggetter gives me a repo to work with.

### DIFF
--- a/_tutorials/smart-local-numbers.md
+++ b/_tutorials/smart-local-numbers.md
@@ -1,0 +1,229 @@
+---
+title: Smart local numbers
+products: voice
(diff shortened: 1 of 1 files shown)


cbetta:
Full code sample is now up on https://github.com/Nexmo/800-replacement

adambutler:
@cbetta Looks great in general. I've made a few corrections and minor tweaks in d774341

PR #139, opened 2017-06-21
Add support for past events

adambutler:
## Description

Adds support for viewing past events.

## Deploy Notes

None


tommorris:
Tested locally, looks good. :+1:

PR #140, opened 2017-06-22
Move SIP to its own sub-product under voice

adambutler:
## Description

Changes menu structure to present SIP separate to Voice API

- [x] Move Content
- [ ] ~Make SIP overview follow overview template~ Replaced by #201
- [x] Add redirects 

## Deploy Notes

None


adambutler:
@tommorris Can you take a look at the content here. We'd need to make this more "Overview like". Can we split any of it out to guides / building-blocks?

I'll sort out the redirects once we've got everything in place.

adambutler:
@tommorris I do not think this structural change should be blocked by content changes any further. I have therefor opened #201 to review the content separately so that this can go into production.

Please can you review.

PR #167, opened 2017-07-05
Add anchor link support

### LINKED ISSUE #154: Add support for modified headings
## Description

Currently, if a heading is changed its id will also change since it is the content of the heading slugified. For example.

```md
# Original Heading
```

... will produce the html: ...

```html
<h1 id="original-heading">Original Heading</h1>
```

If this heading changes, so does the `id` and any anchor links will break along with it.

This could be fixed one of two ways:

1. Have explicit and static ids that don't change
2. Keep implicit ids that do change and maintain a list of previous headings.

I would favour soultion #2 for a number of reasons:

- As evidenced in `www-docs` headings are likely to change over time leaving `id`s to be misrepresentative of the content.
- With solution #1 any copy issues introduced into `id`s would be impossible to fix without breaking permalinks.
- Navigation generation based off of headings would be made more complicated and would require parsing the document as html.

I would therefor propose that we have a yml file that maps old headings to new ones and though the pipeline would automatically augment the title with any anchors that were previously used.

Here is an example to illustrate how this would work:

1. Starting markdown file and output html:

    ```md
    # Original Heading
    ```

    ```html
    <h1 id="original-heading">Original Heading</h1>
    ```

2. Change title:

    ```md
    # New Heading
    ```

3. Add mapping to YAML:

    ```yml
    /some/path:
      Original Heading: New Heading
    ```

4. Generated HTML:

    ```html
    <a name="old-heading"></a>
    <h1 id="new-heading">New Heading</h1>
    ```

adambutler:
## Description

Fixes #154

image

## Deploy Notes

None

### DIFF
--- a/app/filters/anchor_filter.rb
+++ b/app/filters/anchor_filter.rb
@@ -1,0 +1,9 @@
+class AnchorFilter < Banzai::Filter
+  def call(input)
+    input.gsub(/^⚓\ (.+?)\n/) do
+      <<~HEREDOC
+        <a name="#{$1.parameterize}"></a>
+      HEREDOC
+    end
+  end
+end
--- a/app/pipelines/markdown_pipeline.rb
+++ b/app/pipelines/markdown_pipeline.rb
@@ -7,6 +7,7 @@ class MarkdownPipeline < Banzai::Pipeline
       FrontmatterFilter,
       InlineEscapeFilter,
       BlockEscapeFilter,
+      AnchorFilter,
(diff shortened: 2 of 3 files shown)


tommorris:
I'm slightly reluctant at using weird Unicode symbols but... okay. Let's do this.

adambutler:
Due to it's explicitness this will be east to find-and-replace if we decide on an alternative implementation in the future.

PR #178, opened 2017-07-07
Removing code examples from Voice API reference

### LINKED ISSUE #159: Remove code examples from Voice API Reference
## Description

https://developer.nexmo.com/api/voice presently contains bad code examples. They should at least be removed. Maybe they could simply be updated to HTTP payload examples (see https://github.com/Nexmo/nexmo-developer/pull/152#issuecomment-312620982)?

This also removes a body of work from https://github.com/Nexmo/nexmo-developer/issues/48

tommorris:
Fixes #159

### DIFF
--- a/_api/voice.md
+++ b/_api/voice.md
@@ -14,10 +14,6 @@ You use the following requests to create, terminate, record and retrieve informa
 
 [POST] `https://api.nexmo.com/v1/calls`
 
-```tabbed_examples
-source: '/_examples/api/voice/calls/post'
-```
-
 This request contains:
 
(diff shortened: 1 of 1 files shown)


leggetter:
@tommorris could we incorporate the restructuring of the API reference page into this PR? Or would you rather that be a separate issue?

https://github.com/Nexmo/www-docs/pull/260

adambutler:
@leggetter Tom and I have spoken about this. We've decided it would be easiest to re-apply the `www-docs` changes on top of our changes. This will be done as a separate PR.

PR #194, opened 2017-07-12
Language routes

### LINKED ISSUE #155: Link to a page and have a specific language shown in the code examples and related text
## Description

It should be possible to link to a page and have a specific language shown in the code examples and related text e.g.

* cURL - https://developer.nexmo.com/messaging/sms/building-blocks/receiving-an-sms
* PHP -  https://developer.nexmo.com/messaging/sms/building-blocks/receiving-an-sms/php
* Node.JS - https://developer.nexmo.com/messaging/sms/building-blocks/receiving-an-sms/nodejs
* Java - https://developer.nexmo.com/messaging/sms/building-blocks/receiving-an-sms/java
* C# - https://developer.nexmo.com/messaging/sms/building-blocks/receiving-an-sms/c-sharp
* Python - https://developer.nexmo.com/messaging/sms/building-blocks/receiving-an-sms/python

Pages should also be discoverable and appear language specific (unique) to search engines.

adambutler:
## Description

Adds language specific routes for documentation pages.

- [x] Clicking on a tab pushes a change into the address bar.
- [x] Clicking back restores previous url and tab
- [x] When no language is specified in the address the most recent user selected language is used
- [x] When language is set in URL the appropriate code examples are rendered without JavaScript.
- [x] Canonical page is referenced when language is provided in URL

Fixes #155 

## Deploy Notes

None


leggetter:
Going directly to a language URL doesn't appear to render the selected language https://nexmo-developer-staging-pr-194.herokuapp.com/messaging/sms/building-blocks/send-an-sms/ruby

If I were to cURL this page I'd also expect static HTML back with the selected language rendered.

adambutler:
@leggetter Fixed. The environment led to a JS error as the Search box was not present due to the lack of necessary configuration variables.

I've built this such that the page is rendered with the correct language without the need for JavaScript, only the route specific language is rendered visually in the DOM. Other languages are hidden with a `display: none;` rule accurately indicating to search engines that this content is not visible on an initial render.

leggetter:
> Other languages are hidden with a display: none; rule accurately indicating to search engines that this content is not visible on an initial render.

Sounds good, thanks! Let's see the results of this.

adambutler:
cURL is the default but not necessarily the example that is loaded when you land on the page without a language. The selected tab is determined as such:

1. Is there a language in the URL, if so use that.
2. Has the user previously selected a language, if so use the most recent.
3. Fallback to the first tab. Given the tab weighting this is usually cURL.

Changing this behaviour as you proposed would prevent you from explicitly linking to a cURL examples.

leggetter:
> 1. Is there a language in the URL, if so use that.

✅ 

> 2. Has the user previously selected a language, if so use the most recent.

In this case and there is no language in the URL should the URL be dynamically updated?

> 3. Fallback to the first tab. Given the tab weighting this is usually cURL.
> Changing this behaviour as you proposed would prevent you from explicitly linking to a cURL examples.

Without changing the behaviour we'll have many duplicate pages on the site (no language in URL + cURL x all code examples).

Should no language in the URL just mean cURL is selected and that resolves the direct link to cURL? Is that in combination with auto updating the URL if the user has a preferred language set okay or will it be confusing?

adambutler:
> In this case and there is no language in the URL should the URL be dynamically updated?

I don't think so, consider a scenario where the tabbed content is not consistent throughout the document. This could happen for number of reasons including:

- Having a *Request* section with languages (cURL, JS, Ruby etc) followed by a response section with JSON and XML.
- Having a page listing various library calls relating to a subject but certain actions are not supported by one or more languages.

Furthermore somebody might navigate to a page that doesn't even have a tab of the same language as the users most recently selected language. All these examples result in the same problem that the behaviour is made more confusing.

The current implementation will only set the URL on a users action making the behaviour very clear.

> Without changing the behaviour we'll have many duplicate pages on the site (no language in URL + cURL x all code examples).

I've added the canonical link to pages when the language is explicitly set to prevent this being an issue:

```html
<link rel="canonical" href="/same/route/without/language">
```

leggetter:
> Having a Request section with languages (cURL, JS, Ruby etc) followed by a response section with JSON and XML.
> Having a page listing various library calls relating to a subject but certain actions are not supported by one or more languages.

IMO every example we have should be in all of our main 6 languages - the tabs should always be present. If we don't have an example in that language it should be a priority for us to add the example.

XML and JSON shouldn't be counted as languages. They are data (interchange) formats.

> <link rel="canonical" href="/same/route/without/language">

Perfect!

adambutler:
Changed implementation so that `ncco`, `xml` and `json` are not "linkable" but user preference is retained.

adambutler:
@leggetter This is now ready to review again, deployed [here](https://nexmo-developer-staging-pr-194.herokuapp.com/)

Issue #188, opened 2017-07-10
Missing curl examples for verify


ChrisGuzman:
This page https://developer.nexmo.com/verify/guides/verify-a-user has a curl example for Send but not for check or cancel

## Description

Please briefly describe what the issue is

## What did you expect to happen?

Delete section if this is not a bug.

## What actually happened?

Delete section if this is not a bug.

## Steps to reproduce:

Delete section if this is not a bug.

## Related resources

Route links & Documents paths that might be helpful. Delete section if not appropriate.

PR #205, opened 2017-07-17
Content tweaks

### LINKED ISSUE #189: Need to make it clear that 2 hours is the maximum length of a call recording.
## Description
The NCCO Record reference doesn't specify that 2 hours is the maximum call recording length.  I understand that the max call length is 2 hours but we need to make it clear in the record section since customers may be searching for this info specifically and may not read the `limit `parameter description in the Connect action.

https://developer.nexmo.com/voice/guides/ncco-reference#record

adambutler:
## Description

Fixes #189 & #188.

Also adds `ToolTipFilter` which allows one to create tooltips.

## Deploy Notes

None

### DIFF
--- a/_api/verify.md
+++ b/_api/verify.md
@@ -25,7 +25,7 @@ To use Verify Request you:
 ### Request
 
 ```
-[POST] https://api.nexmo.com/verify/:format
+[GET] https://api.nexmo.com/verify/:format
 ```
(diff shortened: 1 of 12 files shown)


tommorris:
I updated the PR description to include the addition of a new filter. Reviewing now.

PR #210, opened 2017-07-18
Update outbound-calls.md

jonlingard:
the top level flow of a VAPI Outbound call (https://docs.nexmo.com/voice/voice-api/calls) .. steps 1 thru 5... says (step 3) that Nexmo GET the NCCO from the answer_url BEFORE (step 4) placing the outbound call
this is wrong right? the NCCO fetch happens after the answer event

i've edited the text above but the graphic needs updating to match

## Description

## Deploy Notes

### DIFF
--- a/_documentation/voice/voice-api/guides/outbound-calls.md
+++ b/_documentation/voice/voice-api/guides/outbound-calls.md
@@ -11,16 +11,16 @@ The workflow to make a Call using the Voice API is:
 
 1. You make an [authenticated](/concepts/guides/applications#minting) Call request to the Voice API.
 2. Nexmo accepts the request and sends response information.  
-3. Nexmo makes a GET request to retrieve an NCCO from your *answer_url* webhook endpoint.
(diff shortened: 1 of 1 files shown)


adambutler:
@jonlingard Thanks for brining this to our attention. I've double checked the flow and you are indeed correct that the `GET` request is not made to the `answer_url` until there is an actual answer.

I wanted however to update the graph image to clarify this further, this allows us to write the flow as such:

1. You make an authenticated Call request to the Voice API.
2. Nexmo accepts the request and sends response information.
3. Nexmo places a call awaits an answer.
4. Nexmo makes a GET request to retrieve an NCCO from your answer_url webhook endpoint and follows the actions in your NCCO.
5. Nexmo sends POST request to your event_url webhook endpoint when the Call status changes. For more advanced systems, depending on the input from your user, you can return an NCCO to customize their experience.

I've therefor opened a new PR at #211 with these tweaks in place, feel free to have a look over it and let me know if you have any other thoughts.

PR #211, opened 2017-07-18
Fixes flow for outbound calls

adambutler:
## Description

Fixes flow as reported by @jonlingard at #210 

## Deploy Notes

None


_documentation/voice/voice-api/guides/outbound-calls.md
```diff
@@ -11,8 +11,8 @@ The workflow to make a Call using the Voice API is:
 
 1. You make an [authenticated](/concepts/guides/applications#minting) Call request to the Voice API.
 2. Nexmo accepts the request and sends response information.  
-3. Nexmo makes a GET request to retrieve an NCCO from your *answer_url* webhook endpoint.
-4. Nexmo places a call and follows the actions in your NCCO.
+3. Nexmo places a call awaits an answer.
+4. Nexmo makes a GET request to retrieve an NCCO from your *answer_url* webhook endpoint and follows the actions in your NCCO.
```
leggetter: Should we clarify that this is "when the call is answered"?

adambutler:
@leggetter Made a copy tweak clarifying the flow.

You can see the diff by clicking the Rich Diff button.

fixes_flow_for_outbound_calls_by_adambutler_ _pull_request__211_ _nexmo_nexmo-developer

I know what you mean about it being easier to edit/maintain however I'm of two minds about this:

If explaining a subject is complicated enough that it can't be done with words alone should that graph not deserve the attention of being designed to clearly explain the flow.

FYI: I wanted to check what a JS Sequence Diagram equivalent would look like, I notice it doesn't support bi-directional arrows.

I'd propose we should get this minor change out of the way and in the future have a further discussion about how we handle graphs and find something that compromises between finesse and adaptability.

leggetter:
RE JS sequence diagram. You tend to use a `<----` style arrow to indicate a response to an outbound.

IMO the 3 should come before the 4 from top to bottom:

screen shot 2017-07-19 at 12 58 50

PR #219, opened 2017-07-19
Slack

adambutler:
## Description

Adds a page to invite users to `nexmo-community`

image

## Deploy Notes

Depends on `SLACK_SUBDOMAIN` and `SLACK_TOKEN`, `RECAPTCHA_ENABLED`, `RECAPTCHA_SITE_KEY`, `RECAPTCHA_SECRET_KEY` variables.


adambutler:
@leggetter Looks like you hit submit before that image finished uploading.

leggetter:
@adambutler updated. Not sure what happened there. I had the page open for ages.

adambutler:
@leggetter 

1. Absolutely agree, we should add a link from 'community' as well. 

2. Personally, I don't see that this would end up being a significant issue. If out slack is open to anybody anyway they could just as easily register themselves and then look at who else is on the Slack community anyway.

3. I hate CAPCHA's but you make a very good point. How about I implement recaptcha and have it toggleable by a config variable, we'll then just enable it if we need to?

4. Yes, on it!

adambutler:
image

Changes implement as per my last comment.

adambutler:
@leggetter Ready for review again

adambutler:
🚀

PR #223, opened 2017-07-20
Many responsive tweaks

adambutler:
## Description

Lots of initial responsive tweaks, plenty more work could be done to improve mobile responsiveness but this change is a basic improvement on what we currently have implemented.

> Note: My GIF-cast tool only uses colours from the fist frame to keep the file size down some colours in the preview below are not accurate:

## Deploy Notes

None


leggetter:
Much better on mobile (small width) 👍

A couple of widths that could probably do with being tweaked so that the search doesn't sit in a strange position.

screen shot 2017-07-20 at 15 44 02

screen shot 2017-07-20 at 15 44 10

adambutler:
@leggetter 👐  Fixed

leggetter:
Brilliant and slightly silly 👍🙌

PR #246, opened 2017-07-27
Add 'cancelled' call status

judy2k:
## Description

The list of call statuses was missing 'cancelled'. @sammachin has confirmed that this is a valid status.


judy2k:
Closing because there's a duplicate PR that's probably more comprehensive from @sammachin

PR #271, opened 2017-08-10
Change licence to be shorthand MIT licence that GH expects

adambutler:
## Description

Changed licence to be shorthand version as suggested by GitHubs Templates.

> Note: Instigated by advice to DevRel by John S on 27th June 2017

## Deploy Notes

None


adambutler:
As discussed we are following by example with the `choosealicence.com` site repo which is [CC 3.0](https://creativecommons.org/licenses/by/3.0/us/deed.en_US) and MIT. See [here](https://github.com/github/choosealicense.com/tree/fe965f16f6f690d40a738e75a46fad3f41f5daa3).

They have a `LICENCE.md` file with only the MIT Licence shorthand.

In their README under the Licence heading, they have copy explaining the licence covering content vs source code.

I've taken this same structure, the only difference is that I have replaced their [CC 3.0](https://creativecommons.org/licenses/by/3.0/us/deed.en_US) for [CC 4.0](https://creativecommons.org/licenses/by/4.0/) as per our current licence.

PR #273, opened 2017-08-10
Remove the need for opening <?php tags

### LINKED ISSUE #186: Enable code highlighting without requiring opening PHP tags
## Description

At the moment code snippets require `<?php` tags. For content that isn't at the beginning of the file, this is misleading. We should support highlighting code even if the opening tag is not present

## Related resources

* https://github.com/Nexmo/nexmo-developer/issues/181

adambutler:
## Description

Fixes the need for opening `<?php` tags for PHP syntax highlighting.

Fixes  #186

## Deploy Notes

None


_examples/api/application/create-an-application/PHP
```diff
@@ -1,5 +1,3 @@
-<?php
```
mheap: If we're showing a complete example that's a standalone file, I think we should keep the opening `<?php` to show that it is a complete file. This request was mainly for tutorials where we discuss a small part of code

PR #253, opened 2017-07-30
Performance/tweaks 1

adambutler:
## Description

Various performance tweaks and dependency updates.

- Reduce dependencies
- Remove unused CSS
- Remove unused Foundation JS
- Move components into Webpack where possible
- Combine JS assets
- Add `defer` loading to pack script
- Update Foundation
- Update Rails
- Update Webpacker

**Audit report before:**

screen shot 2017-07-30 at 10 20 45

**Audit report after:**

screen shot 2017-07-30 at 10 09 48

## Deploy Notes

None

PR #276, opened 2017-08-15
Implement GTM

adambutler:
## Description

Adds Google Tag Manager to Nexmo Developer.

## Deploy Notes

Merge: @adambutler


adambutler:
Before:

screen shot 2017-08-15 at 10 57 55

After:

screen shot 2017-08-15 at 10 58 15

adambutler:
Changes were made to GTM to reduce the amount of scripts loaded for Nexmo Developer. It has marginally improved the Audit score.

Since this change has an impact on Analytics to ensure that metrics are easily split this PR will be merged on 1st September 2017

image

leggetter:
What are the render blocking scripts?

adambutler:
These are the scripts that are non-modular including [application.js](https://github.com/Nexmo/nexmo-developer/blob/gtm/app/assets/javascripts/application.js) and GTM.

What remains in application.js are scripts that couldn't be moved into webpack when I did the performance tweaks @ #253. For example, because jquery binds to `window` and other libraries depend on this. Also `smooth-scroll` doesn't expose itself as a module.

I did just notice that `webfont.js` was still in here from before `js-sequence-diagrams` was moved. I've now removed ths.

These numbers are actually a little misleading since it takes any render blocking script as a file and then measures the full time from the first script call to the very last call even if that last call happened post-render:

nexmo_developer
nexmo_developer

PR #296, opened 2017-08-22
Fixed JSON files

JamieChapman8462:
## Description

Noticed that when pasting some of the JSON examples on the site they were invalid, so made the relevant changes to the JSON files.

## Deploy Notes

N/A.


adambutler:
👍  Looks great!

🎉 Thank you very much for your contribution and for helping to improve Nexmo.

PR #299, opened 2017-08-24
Events ical feed

adambutler:
## Description

Adds events ical endpoint.

Review app events: https://nexmo-developer-staging-pr-299.herokuapp.com/community

Subscription endpoint: webcal://nexmo-developer-staging-pr-299.herokuapp.com/feeds/events.ics

## Deploy Notes

None


leggetter:
Looks great!

Any way of adding a title or something so that it has a friendly name when you add it to GCal?
screen shot 2017-08-24 at 16 34 34

Are the test events supposed to be in 2020? Just checking in case there's some sort of error here.
screen shot 2017-08-24 at 16 35 13

app/controllers/feeds_controller.rb
```diff
@@ -0,0 +1,17 @@
+class FeedsController < ApplicationController
+  def events
+    calendar = Icalendar::Calendar.new
+
+    Event.all.each do |event|
+      calendar.event do |e|
+        e.dtstart = Icalendar::Values::Date.new(event.starts_at)
+        e.dtend = Icalendar::Values::Date.new(event.ends_at + 1.day)
```
leggetter: Interested to know why `+ 1.day`?

adambutler: Just weirdness in the iCalendar format (RFC-5545).

We are using `Date` instead of `DateTime`. Take a one day event for example:

If I was to have an event that ran from `2017-01-01` to `2017-01-01` then iCalendar format would treat this as an event at `00:00` on `2017-01-01` with no duration.

Changing this to range from `2017-01-01` to `2017-01-02` will cause it to be interpreted as an "All day" event on `2017-01-01`.

adambutler:
> Are the test events supposed to be in 2020? 

Yep. That is how I have the test data setup in my [seeds file](https://github.com/Nexmo/nexmo-developer/blob/master/db/seeds.rb)

PR #300, opened 2017-08-25
Update ncco.md

jonlingard:
per VAPI-616 you can have multiple Stream actions before an Input. Is the same true of Talk too?

## Description

## Deploy Notes


adambutler:
Hey @jonlingard, we've been having a little trouble with our CI getting stuck with external contributions.

The issue has been resolved but in this case I had to open a fresh PR to trigger the newly introduced webhook. You can find the new PR at #310

PR #303, opened 2017-08-27
Change search to use Lambda

adambutler:
## Description

Changes search to use a Lambda instead of `/search.json` on Nexmo Developer.

The benefit of this change is that multiple properties can consume Algolia Search with server-side filtering without depending on one-another but rather a server with a single responsibility.

## Deploy Notes

Tested on my own personal AWS account. Will depend on <https://github.com/Nexmo/nexmo-search-lambda> being deployed to Nexmo's AWS account.


leggetter:
This is introducing a different infrastructure provider. Is there any way we could serve this from Heroku in order to use the same provider?

adambutler:
👍 Firstly, this is just a Node app so could easily be adapted for Heroku.

The issue with doing so would be that the application would require a paid dyno to provide a responsive interface, something I considered to be overkill for something that is so basic.

The benefit of using Lambda is that it's only being paid for while it is being consumed and allows us to avoid issues of maintenance and scalability, exactly what this kind of service was designed to solve.

I can foresee us in the future having other AWS dependencies such as S3 and CloudFront so I don't think we should be limiting ourselves away from tooling provided by AWS but do agree that we should introduce service dependencies with caution.

If we choose not to use Lambda then I don't think the gain would be enough to warrant this change and would instead suggest that Nexmo Developer itself becomes responsible for hosting this behaviour.

adambutler:
I'm done with tweaks on this PR.

@leggetter can you review the [deployment](https://nexmo-developer-staging-pr-303.herokuapp.com/) please.

PR #308, opened 2017-08-29
Properly encoding JSON response.

### LINKED ISSUE #307: Missing json.dumps from Python sample code
## Description
Documentation page: 
https://developer.nexmo.com/voice/voice-api/guides/interactive-voice-response

The IVR guide is missing json.dumps in step 5 of the workflow. 

The last line reads `s.wfile.write(ncco)` which causes the NCCO generated to have mostly single quotes.

The last line should be

`s.wfile.write(json.dumps(ncco))`

so that the json python library outputs correctly formatted json with double quotes.

## What actually happened?

causes the NCCO generated to have mostly single quotes.

## Steps to reproduce:

Just follow the guide and copy an paste exactly what the sample code says. Replace phone numbers and answer/event url's

See attached
greenshot 2017-08-28 23 46 00

tommorris:
Fixes #307.

### DIFF
--- a/_examples/voice/guides/interactive-voice-response/2/Python
+++ b/_examples/voice/guides/interactive-voice-response/2/Python
@@ -63,4 +63,4 @@ class MyHandler(BaseHTTPServer.BaseHTTPRequestHandler):
         s.send_response(200)
         s.send_header("Content-Type", "application/json")
         s.end_headers()
-        s.wfile.write(ncco)
+        s.wfile.write(json.dumps(ncco))


adambutler:
@tommorris Has this just been changed or tested as well?

tommorris:
@adambutler I don't think it needs testing. This code all needs to be rewritten anyway (to use Flask), but #307 is right to point out the inconsistency between the two Python code samples on this page.

The existing code will return the `__repr__` of the Python dict (in Ruby terms, the `.inspect`)  rather than a JSON encoding. When we switch these code samples to Flask, I think Flask automatically handles turning dicts into JSON without having to encode them, but Python's `BaseHTTPServer` doesn't have anything like that.

PR #310, opened 2017-08-29
Update ncco.md

adambutler:
Copied from #300 

per VAPI-616 you can have multiple Stream actions before an Input. Is the same true of Talk too?


sammachin:
Yep same works for talk and stream just tested it. 
Docs edit looks fine

PR #311, opened 2017-08-29
API Sidebar fixed to link to correct IDs

tommorris:
The Rails helper function generates an ID by parameterizing the header's
text. This breaks if there are numerous headers the same (real life
examples: "Request", "Response", "Keys and Values" etc.)

Instead, we can just read the ID that's been generated by the Markdown
parser and use that instead. This should fix lots of broken links on the
site.

Raised originally on JIRA: DOCS-290.

### DIFF
--- a/app/helpers/navigation_helper.rb
+++ b/app/helpers/navigation_helper.rb
@@ -32,7 +32,7 @@ module NavigationHelper
 
       nodes << <<~HEREDOC
         <li>
-          <a href="##{heading.text.parameterize}" data-scrollspy-id="#{heading['data-id']}">
+          <a href="##{heading.attributes['id']}" data-scrollspy-id="#{heading['data-id']}">
             #{heading.text}
           </a>
       HEREDOC


aaronbassett:
As this hash will potentially appear in the URL and may be shared, should it still contain the human readable form within the hash?

`##{heading.text.parameterize}-#{heading.attributes['id']}`

tommorris:
@aaronbassett The code that generates the ID is done elsewhere. All this fixes is that the helper function that generates the _sidebar_ pulls the ID in correctly. This PR doesn't affect how the IDs are generated, merely ensures that the correct links get put in the sidebar.

The problem is, if you have a document like this:

```
## My Heading

Some text.

## My Heading

Some more text.
```

The navigation helper will generate two links, both pointing to `#my-heading` even though the second header has the generated ID `#my-heading-2`.

aaronbassett:
Ah ok, sorry for my misunderstanding :)

PR #317, opened 2017-08-31
Fix inline anchors

SolarBear:
The anchor id generated by the Markdown for the "Delivery Receipt" section is `delivery-receipt` (with a dash) while the two links inside the page use `delivery_receipt` (with an underscore).

## Description

Inline references have been fixed to use dashes instead of underscores so the links can move you to the proper section in the page.

## Deploy Notes

None.


adambutler:
👍  Looks great!

🎉  Thank you very much for your contribution and for helping to improve Nexmo Developer.

PR #327, opened 2017-09-05
Adding examples of PUT requests for VAPI

tommorris:
For some reason, this didn't make it onto GitHub before.

### DIFF
--- a/_examples/api/voice/calls/put-request/transfer.md
+++ b/_examples/api/voice/calls/put-request/transfer.md
@@ -8,7 +8,7 @@ menu_weight: 6
     "action": "transfer",
     "destination": {
       "type": "ncco",
-      "url:" ["http://example.org/new-ncco.json"]
+      "url": ["http://example.org/new-ncco.json"]
     }
 }
 ```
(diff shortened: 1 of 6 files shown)


_examples/api/voice/calls/put-request/transfer.md
```diff
@@ -0,0 +1,16 @@
...
+title: Transfer
+menu_weight: 6
+---
+
+```json
+{
+    "action": "transfer",
+    "destination": {
+      "type": "ncco",
+      "url:" ["http://example.org/new-ncco.json"]
```
adambutler: Invalid JSON, should be:

```
"url": ["http://example.org/new-ncco.json"]
```

tommorris: @adambutler fixed.

tommorris:
The content was supposed to have been added in #222.

leggetter:
@tommorris I know this has been merged - a small request: can we try to make everything linkable?

I don't know if we can link to the examples with the transfer tab selected. But maybe we could put an `### Examples` heading which would mean we could say `Go [here] and click "Transfer" to see the example`.

PR #334, opened 2017-09-07
Update send-sms code example

AlexLakatos:
## Description
- removed import
- changed req/res properties


tommorris:
Okay, this is done to improve compatibility between Node versions. (Just a note-to-future-self)

PR #338, opened 2017-09-11
Add theme support for other environments

adambutler:
## Description

If `ENV['THEME']` is present it is set on the body allowing styling overrides on a per environment basis. For example, staging could have its own header color making it clearer to the user that they are not on the primary site.

## Deploy Notes

None


app/views/layouts/api.html.erb
```diff
@@ -2,7 +2,7 @@
 <html>
   <%= render partial: 'layouts/partials/head' %>
 
-  <body>
+  <body class="<%= theme %>">
```
tommorris: What happens if `ENV['THEME']` _isn't_ set? The theme would get `nil` at this point. Do we need to protect against this by passing an empty string in?

adambutler: `theme` is a method in [`application_helper.rb`](https://github.com/Nexmo/nexmo-developer/blob/9eb755c39996473715041912d5c0ff1e9421b8c2/app/helpers/application_helper.rb#L12).

It does return `nil` if the environment variable is not set but in Ruby `nil.to_s` returns empty string.

tommorris: Ah okay, that's sensible. :+1:

PR #342, opened 2017-09-11
Restyled Quicksearch and Search page

BennyZ28:
## Description

Changed the styling of the quick search and the Search page based on this mockup

pasted_image_at_2017_09_11_12_19_pm


adambutler:
😻  Thanks for the contribution, I'll review this tomorrow morning.

app/assets/stylesheets/layout/_search.scss
```diff
@@ -2,49 +2,70 @@ $quick-search-width: 800px;
 
 .search-result {
   overflow: hidden;
+
   > a {
     display: block;
-    padding: $spacing/2 0;
+    padding: 0;
```
adambutler: I think this `padding` rule can be removed if it's `0`.

app/assets/stylesheets/layout/_search.scss
```diff
@@ -2,49 +2,70 @@ $quick-search-width: 800px;
...
 
     .quicksearch & {
-      padding: 0;
+      padding: 0 !important;
```
adambutler: As a general rule can we try and avoid using `!important` (though I'll admit I'm guilty of this occasionally).

In this specific case, it doesn't look like this rule is doing *anything* now anyway with the removal of padding on line 8.

app/assets/stylesheets/layout/_search.scss
```diff
@@ -2,49 +2,70 @@ $quick-search-width: 800px;
...
+      width: 100%;
```
adambutler: What do you think about using `display: block` here instead of `width: 100%` ?

app/assets/stylesheets/layout/_search.scss
```diff
@@ -2,49 +2,70 @@ $quick-search-width: 800px;
...
     }
   }
 
   > a > div {
-    border-left: 2px solid darken($mint, 3%);
-    padding-left: 15px;
+    border-bottom: 1px solid darken($mint, 3%);
+    padding: $spacing/2;
+    width: 100%;
```
adambutler: This `width` rule isn't doing anything as this element is a `block`.

app/assets/stylesheets/layout/_search.scss
```diff
@@ -2,49 +2,70 @@ $quick-search-width: 800px;
...
 
     h3{
       color: $link;
+      margin-top: 5px;
+      font-weight: bold;
```
adambutler: Nitpick: We tend to use values for `font-weights` in this project... To help with searchability in the future I'd suggest changing this to:

```css
font-weight: 700;
```

app/assets/stylesheets/layout/_search.scss
```diff
@@ -2,49 +2,70 @@ $quick-search-width: 800px;
...
+        margin: 5px 0 0;
       }
     }
 
     .meta {
+      background: lighten($stone-light, 25%);
+      padding: 2px 5px;
+      width: auto;
+      display: inline-block;
+      border-radius: 3px;
```
adambutler: Use `$radius` here

app/assets/stylesheets/layout/_search.scss
```diff
@@ -2,49 +2,70 @@ $quick-search-width: 800px;
...
         font-size: 0.95em;
-        margin: 0;
+        margin: 5px 0 0;
       }
     }
 
     .meta {
+      background: lighten($stone-light, 25%);
+      padding: 2px 5px;
+      width: auto;
```
adambutler: `width` property seems to have no effect on styling here.

app/assets/stylesheets/layout/_search.scss
```diff
@@ -2,49 +2,70 @@ $quick-search-width: 800px;
...
       -webkit-box-orient: vertical;
 
+      &.search-highlighted {
+        color: $gray;
+        margin-top: 5px;
+        font-style: italic;
+
+        em {
+          color: $pink;
+          font-weight: bold;
```
adambutler: Also suggest changing this value to `700`

app/assets/stylesheets/layout/_search.scss
```diff
@@ -53,15 +74,11 @@ $quick-search-width: 800px;
...
     > a > div {
-      h3 {
-        color: darken($link, 10%);
-      }
+      background: darken($mint, 3%);
 
-      p {
-        color: $text;
+      .quicksearch & {
+        background: lighten($stone-light, 25%);
```
adambutler: I'm not keen that we are using two different `:hover` background colours for quicksearch and normal search.

Can we remove the `quicksearch` block here:

```sass
.quicksearch & {
  background: lighten($stone-light, 25%);
}
```

Then for the `backkground` on line 77 use the color `$mint` since this is close visually to both:

- `lighten($stone-light, 25%);`
- `darken($mint, 3%);`

app/assets/stylesheets/layout/_search.scss
BennyZ28: So, "#subnav a" has padding and it was overriding this, but I didn't realise that got fixed with a couple of lines added around 107, so I agree, this can go now.

app/assets/stylesheets/layout/_search.scss
BennyZ28: So, "#subnav a" has padding and it was overriding this, but I didn't realise that got fixed with a couple of lines added around 107, so I agree, this can go now.

app/assets/stylesheets/layout/_search.scss
BennyZ28: I just removed the whole block because display: block is already in it's sass parent

app/assets/stylesheets/layout/_search.scss
BennyZ28: Cool, changed

app/assets/stylesheets/layout/_search.scss
BennyZ28: This is actually because the background of the quick search is white while the background of the search page is light blue, which I think doesn't help readability but is beside the scope of this changes. So, the hover background has to have a different colour to give the effect of a soft background colour change. 

Unless we use backgrounds with opacity, then it can be the same? 

Or unless the background of the search page becomes white...

adambutler: Good point 👍  it can stay as is.

adambutler:
🎉  Great job! All looks great!

I noticed we have a responsive issue between `640px` - `830px` width but this was not introduced by these changes and is present in production.

I'm happy for these changes to take effect and have opened a [seperate issue](https://github.com/Nexmo/nexmo-developer/issues/344) for the responsive bug.

_If_ you find yourself with some time to fix this please just assign it to yourself, otherwise I'll pick it up soon.

Feel free to merge whenever you are ready, it will automatically be deployed to production when you do within about 5 min.

Issue #198, opened 2017-07-14
Footer


adambutler:
## Description

Jack mentioned it would be good to bring the nexmo.com footer into Nexmo Developer. I think it's a little too loaded for a documentation site & goes against one of the values of being Developer centric and putting simplicity first.

I do however agree that the current footer has not had much thought put into it and think there are elements of the nexmo.com footer that we could incorporate such as:

- API Status
- The get to know us section
- Links to support
- LinkedIn Link (currently missing)

Possibly even:

- Contact details
- Traditional footer Link Terms, Privacy Policy etc.
- Products (if done in a way that wasn't confusing, I'd suggest having a generic link back to nexmo.com would be more appropriate)

leggetter:
Link to the Slack community invite page?

sammachin:
The idea of status being on nexmostatus.com is that it isn't tied to any of the nexmo infrastucture even the DNS for nexmo.com (it used to be status.nexmo.com) so if we have an outage of DNS (like we did last oct) its still accessable, we should be encouraging users to visit and bookmark nexmostatus.com rather than importing the status feed into developer.nexmo.com

adambutler:
![image](https://user-images.githubusercontent.com/1238468/30320182-cdbcb9fa-97a9-11e7-8e67-40071a9a1415.png)

☝️ Proposed design, feedback welcome.

BennyZ28:
Looks awesome!

adambutler:
Fixed by #343

PR #343, opened 2017-09-12
Footer redesign

adambutler:
## Description

Footer redesigned as outlined in #198 

image

## Deploy Notes

None


app/assets/stylesheets/objects/_api-status.scss
```diff
@@ -0,0 +1,45 @@
...
+    display: inline-block;
+  }
+
+  span {
+    display: block;
+    padding: 1px 5px;
+    float: left;
+  }
+
+  &-status {
```
BennyZ28: Why the single dash? Are you still following BEM here?

adambutler: This is to add the class `api-status-status`

Unlike `none`, `minor`, `major` and `critical` this `-status` isn't a modifier but rather an element itself.

PR #347, opened 2017-09-12
Update verify.md

ecarrollc:
I changed "successful" to "success" - this is the correct status response coming back on a Verify Search.  

Also, in no point on this page does it mention api_key and api_secret as required parameters.  I think each section should state those as required parameters.

## Description

## Deploy Notes


tommorris:
Thank you for your contribution, @ecarrollc!

PR #352, opened 2017-09-15
Add support for language information in feedbacks

adambutler:
## Description

This change will now associate the users selected code language in feedbacks if they have visited a page on a language specific URL or have selected a language in a `TabbedExample` component.

It also goes through the existing data to correct resource URIs that contain the language within the URLs or other artefacts such as URL parameters.

## Deploy Notes

- [ ] `rake db:migrate`

### DIFF
--- a/app/admin/feedback_feedback.rb
+++ b/app/admin/feedback_feedback.rb
@@ -22,6 +22,7 @@ ActiveAdmin.register ::Feedback::Feedback, as: 'Feedback' do
     column :owner
     column :ip
     column :comment
+    column :code_language
     actions
   end
 end
--- a/app/admin/feedback_feedback.rb
+++ b/app/admin/feedback_feedback.rb
@@ -23,6 +23,9 @@ ActiveAdmin.register ::Feedback::Feedback, as: 'Feedback' do
     column :ip
     column :comment
     column :code_language
+    column :code_language_selected_whilst_on_page
+    column :code_language_set_by_url
+    column :created_at
     actions
   end
 end
--- a/app/assets/stylesheets/_core.scss
+++ b/app/assets/stylesheets/_core.scss
@@ -139,3 +139,7 @@ hr {
 .flex {
   display: flex;
 }
+
+.hide {
+  display: none;
+}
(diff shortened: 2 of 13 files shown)


leggetter:
@adambutler - what's the best way to test this out?

adambutler:
The migration you'd have to test locally with a dump of production DB which I've done.

Anything else can be tested in a review app, will deploy one now.

leggetter:
A few questions:

If somebody lands on NDP for the first time and submits a feedback item what value would be saved?

If they select a cURL example what value would be saved?

If they are on one page and select a language (e.g. Java) and then navigate to another page where that language is not available what will be saved? _Note: it's probably still really useful to save their language preference as it flags a demand for an example we don't presently have_

leggetter:
It doesn't seem to report the language if a user navigates to a new page even if they have selected a language on a previous page and that language tab is selected on the new page.

screen shot 2017-09-15 at 11 07 48
screen shot 2017-09-15 at 11 07 58

adambutler:
**If somebody lands on NDP for the first time and submits a feedback item what value would be saved?**

- An `Author` would be created
- A `Resource` would created if the page has never had feedback otherwise if one exists it is reused.
- A `Feedback` would be created belonging to the Author and Resource

**If they select a cURL example what value would be saved?**

If the user selects cURL then provides a feedback the language would be attached as an attribute on the feedback.

**If they are on one page and select a language (e.g. Java) and then navigate to another page where that language is not available what will be saved?**

The language would not be logged against the feedback in this scenario.

adambutler:
**It doesn't seem to report the language if a user navigates to a new page even if they have selected a language on a previous page and that language tab is selected on the new page.**

No it won't and I don't think it should for a few reasons:

- The user hasn't performed an action relating to the content they are viewing that would indicate that their feedback has anything to do with the language

- The page may not have a code example or even the language.

_If_ you believe that it's important that we capture this regardless I would suggest that we have:

`code_language` - Set if the URL is language specific or the user has clicked a language or the user has a language set in their localstorage.

`code_language_selected_whilst_on_page` - Set to true only if the language was clicked during the same page view as the feedback session.

`code_language_set_by_url` - Set to true only if the page the user was giving feedback on was set because the page is language specific.

adambutler:
I've since thought about this and decided it's better to have the data and not use it than to take the risk by not collecting it. I've therefore updated the code to reflect the suggestion in my previous comment.

@leggetter Please can you re-review.

PR #369, opened 2017-09-20
Tabbed Examples Filter now throws an error if source/tabs not set

tommorris:
The point of this is to make it so an obvious error is thrown when the
filter does not get the information it needs to render the template.

The preconditions of the filter is that the YAML block contains either
tabs or source. If it doesn't contain either, we should throw an error
earlier rather than wait for a bigger problem to bubble up.


adambutler:
I've made a few tweaks here:

- Avoid `unless` by using a guard clause instead (see Style/GuardClause)
- Moved validation into its own private method
- Fixes issue where `@config` could be `nil` meaning the validation wouldn't work.
- Removed config from raised exception
- Removed `object.has_key?('foo')` in favour of `object['foo']`
- Added a spec covering both `nil` and invalid scenarios.

PR #382, opened 2017-09-25
Link target fixes

tommorris:
First crack at fixing up link targets. This is a pretty big project, but it's better to deploy a first tranche of these changes than let them linger.

Statistics from [localspider](https://github.com/tommorris/localspider):

* Before: 308 failures.
* After: 268 failures.


adambutler:
I've take a look over the first part of the diff and have some concerns.

Although I agree with your comment "better to deploy a first tranche of these changes than let them linger." I think some of these changes may further bury some of the link issues. For example:

link_target_fixes_by_tommorris_ _pull_request__382_ _nexmo_nexmo-developer

PR #384, opened 2017-09-27
Content/sms

adambutler:
## Description

Improvements and structural changes to the SMS section of Nexmo Developer. Based on the following structure:

ndp sms

## Deploy Notes

None (at the moment)


tommorris:
Okay, I like this a lot. 👍

adambutler:
Responsiveness has been improved as much as possible given the content and layout used today.

Content column needs to be at least `700px` wide (or `630px` once font-size has been reduced) & side navigation at least `250px`.

| width | Columns | Description |
| --- | --- | --- |
| `> 1382px` | `3` | Navigation, content & code coulmns shown inline |
| `<= 1382px` | `2` | Navigation static at top. Content & code coulmns shown inline |
| `<= 1132px` | `1` | Navigation static at top. Content stacked ontop of code examples |
| `<= 630px` | `1` | Same as above except content starts to scroll within it's container. |

See below for a visual of this configuration:

img_4028

leggetter:
## Overview page:

It feels like there should be an overview of the components within an application that uses the SMS API and how they interact with each other. We previously achieved this with a sequence-like diagram. This seems missing now. Something that covers an overview of the steps and something that provides a visual aid to the reader.

In-page links not working:

- Nexmo SMS API Concepts link does not work to #concepts Are we missing a concepts section? Or have these been renamed to "Prerequisites"?
- References to #references

Getting Started

- I believe we must offer examples in all supported languages and not Node only

## guides/delivery-receipts

Missing a diagram that visualises what the detailed guide explains. What are the components? How do they interact?

Generally, a much better description of delivery receipts :+1:

## concatenation-and-encoding

I love the show and tell nature of this updated page.

I'd suggest updating the text from "{chars} over {parts} parts" to "{chars} sent in {parts} message parts". "over" could be confusing as some may read it as a preposition (Extending directly upwards from) rather than an adverb (Expressing passage or trajectory across an area).

adambutler:
Changed this to include both code examples and moved the interactive example to a 'Try it' section... Thoughts?

image

leggetter:
I thought Tech.io had support for multiple runtimes? Is there any reason why we don't want to do that?

I wonder if simply having the ability to send a text from the UI to your phone would be better than runnable code if we can't do all languages?

How can we A/B test this or get feedback? Could we have a feedback request near the new "Try it" section?

adambutler:
> I thought Tech.io had support for multiple runtimes? Is there any reason why we don't want to do that?

It does but getting the Node one to work with Node alone was a fair bit of a challenge. Code examples don't make very testable code which is required for providing user feedback. This is for a number of reasons:

1. Arguments are not provided to the block as parameters but rather variables.
2. Simply including the code executes it since it's not wrapped in a class or method.
3. Node has no global context so checking bare variables requires a bit of a weird hack using a tool called [rewire](https://github.com/jhnns/rewire) to access.
4. I expect these issues would only be harder to work around in other languages.

> I wonder if simply having the ability to send a text from the UI to your phone would be better than runnable code if we can't do all languages?

Maybe. There is a prototype that allows for both outbound and inbound SMS to be demonstrated. If we do this I'd suggest launching without the 'Try it' section for now.

> How can we A/B test this or get feedback?

A/B testing would mean we'd have one or the other. I think this would be appropriate given a code example and "ability to send a text from the UI" but I'm not sure how we'd compare them. We'd be able to measure the interactive example but the code examples are taken off-site where we can't track their usage.

> Could we have a feedback request near the new "Try it" section?

Let's decide on the above first before working out if this is required.

PR #387, opened 2017-09-28
Refactor feedback

adambutler:
## Description

> Note: An artificial delay of 1000ms when taking the Gif-cast above has been added to make this clearer. Also a user would not be challenged after passing a previous challenge.

**tldr; This change refactors the feedback mechanism and fixes a race condition bug along the way.**

When the invisible capcha was implemented I set it up so that when the user has passed once then they are not challenged again. This was done by not challenging the user if:

- The feedback exists already and is being updated
- The user has a value held against their session 

Since we validate the token on the serverside rarely but occasionally the user could change their sentiment or submit their email/comments (or more likely just confuse this button as something they need to press to send their sentiment) before the ID is returned.

This meant a new feedback would try to be generated as it was considered new due to a lack of `id` however it would have the same associated recaptcha token which was now invalidated due to being previously validated.

This change introduces better state handling into the component and will prevent issues like this.

## Deploy Notes

- [x] Test on review app


adambutler:
Available on review app.

leggetter:
Tried on the review app https://nexmo-developer-staging-pr-387.herokuapp.com/api/voice

Got error selecting thumbs up for feedback.

screen shot 2017-09-29 at 16 54 49

PR #389, opened 2017-09-29
Removing redundant Tutorial links from overview

tommorris:
There are Tutorial sections in the overviews for SMS and Voice API but not for Verify or NI. We should have them on all (preferably auto-generated) or none.

Fixes DOCS-298.

### DIFF
--- a/_documentation/messaging/sms/overview.md
+++ b/_documentation/messaging/sms/overview.md
@@ -14,13 +14,13 @@ With low latency and high deliverability, our SMS API is the most reliable way t
 
 ## Contents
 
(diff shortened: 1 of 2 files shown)


adambutler:
Not sure what this has to do with DOCS-298, also not sure I agree with this change in general.

We should have tutorial links on all product overview pages, I'll be looking at adding multiple language support to tutorials in the future and will implement auto-generated link support when I do.

tommorris:
@adambutler I'm closing this as #396 solves it more in line with what you've suggested.

PR #424, opened 2017-10-10
update e.164 link to a valid website.

valeryan:
e164.org is not a valid website that provides any information on the e.164 number plan. The website I have linked to provides extensive technical documentation on the standard. However, it may be a better choice to just link to wikipedia on e.164.


adambutler:
Thanks for spotting... we use `https://en.wikipedia.org/wiki/E.164` as the link elsewhere in the site. 

I've checked other similar links throughout the documentation and have ensured these links are all consistent.

PR #429, opened 2017-10-14
Fixed link to status codes

AverageMarcus:
## Description

A link pointing to error codes was no longer present on the page. The target had been changed to a more generic "status codes" to also include non-error codes. Updated the link to point to the new status codes target.

## Deploy Notes

N/A


adambutler:
👍  Hey @AverageMarcus thanks again for your contribution and for helping to improve Nexmo Developer.

AverageMarcus:
Wow that was a quick response! :D Nice one!

PR #438, opened 2017-10-24
Update number-insight.md

paperbag-zz:
Anchor link fix


adambutler:
👍  Looks great!

🎉 Thank you very much for your contribution and for helping to improve Nexmo.

PR #485, opened 2017-11-21
PUT Response 200>204

sammachin:
Updated response for Modify call to reflect 200>204 change


sammachin:
Sit tight on this one for a little while, we might be going back to 200 😱

PR #493, opened 2017-11-22
Update private-voice-communication.md (Fix cut off function `var fromUserAToUserB = function(from, to, conversation) {` and multi-line comments.)

alwell-kevin:
## Description

Fix cut off function `var fromUserAToUserB = function(from, to, conversation) {` and multi-line comments.


adambutler:
@alwell-kevin Thanks for picking up on this. I pulled down these changes which all look great but noticed one further issue aligned with this that wasn't picked up by your changes so have added this also. 

PR #497, opened 2017-11-24
Create Node

sammachin:
Added building blocks for Number Insight with Node

### DIFF
--- a/_examples/number-insight/advanced/Node
+++ b/_examples/number-insight/advanced/Node
@@ -1,0 +1,14 @@
+var Nexmo = require('nexmo');
+
+var nexmo = new Nexmo({
+    apiKey: API_KEY,
+    apiSecret: API_SECRET})
+
+nexmo.numberInsight.get({level: 'advancedSync', number: NUMBER}, (error, result) => {
+  if(error) {
+    console.error(error);
+  }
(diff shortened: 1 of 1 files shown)


adambutler:
Hey Sam. Thanks for kicking this off, doing these is on my list, however:

All new building blocks or updates should be spec'd out like this [example](https://gist.github.com/adambutler/fdfeba152303388adc3fbc7c99272123) and should adhere to the [Contributing Code Examples](https://developer.nexmo.com/contribute/guides/code-examples) guide.

This has things like consistent use placeholders like `NEXMO_API_KEY` and what linters code examples need to pass.

Happy to adopt this and get it spec'd if you want or would you like to do that?

sammachin:
Yep sorry I just realised after I'd created the PR so I've updated it, wasn't sure what to do with the NUMBER param,
The guide says `TO_NUMBER` or `FROM_NUMBER` but for Number Insight (and Verify) there's only really one number, so we could go with `NUMBER` for now I've put in the placeholder number so its consistent with the other languages and the response makes sense.

leggetter:
Here's a good example of the new format too https://github.com/Nexmo/nexmo-developer/pull/457/files

PR #508, opened 2017-11-29
Improving wording for voiceStatusCallback in Numbers API

tommorris:
Fixing incorrect title as well

### DIFF
--- a/_api/developer/numbers.md
+++ b/_api/developer/numbers.md
@@ -1,10 +1,10 @@
 ---
-title: Developer - Numbers API Reference
+title: API reference
 description: Reference guide for the Account API.
 api: Developer API
 ---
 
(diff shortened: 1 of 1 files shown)


_api/developer/numbers.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: Developer - Numbers API Reference
+title: API reference
 description: Reference guide for the Account API.
 api: Developer API
 ---
 
-# Developer - Numbers API Reference
+# API reference
```
adambutler: Why the change here? These were changed a little while ago to include the context so that the page is clearer when searching.

tommorris: I thought that was the approach we were going with. I'll flip it back.

PR #518, opened 2017-12-04
Update documentation for TLS calls

marrold:
Updates documentation for TLS calls

## Description

Updates the documentation for TLS calls, as we migrate from `sips` to `;transport=tls`

## Deploy Notes

All changes to the proxies to handle this have been made. We will still continue to support the old `sips` method so there's no need for customers to update their existing URIs.


adambutler:
@sammachin I think you are best positioned to review this, mind talking a look please?

PR #521, opened 2017-12-08
Add SSML audio examples

adambutler:
## Description

- Adds audio examples for SSML page
- Adds Audio Filter
- Documents Audio Filter in styleguide

image

## Deploy Notes

None


adambutler:
@leggetter The bucket was `developer.nexmo.com` which was being picked up as nested subdomains so didn't match the certificate `*.s3.amazonaws.com`.

I've changed the bucket to be `nexmo-developer-production` which has resolved this issue.

PR #538, opened 2017-12-28
Update Docker getting started instructions

elof:
## Description

I think it's a helpful to call out that the user still needs to create a .env file and run yarn install before starting the web server. It's a small thing, but it tripped me up for a hot second.

## Deploy Notes

N/A


adambutler:
👍  Looks great!

🎉 Thank you very much for your contribution and for helping to improve Nexmo Developer.

PR #540, opened 2018-01-03
Replaces lambda server in favour of secured keys

adambutler:
## Description

Removes the requirement of the lambda server in favour of using a [secured key](https://www.algolia.com/doc/guides/security/api-keys/#secured-api-keys) to provide filtering.

🐎  Improves search speed approximately 3x - 10x depending on region
☯️  Simplifies stack

---

🙌  Thanks @dzello from @Algolia for the help.

## Deploy Notes

- [x] Test on staging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@
 /public/packs
 /node_modules
 /public/packs-test
+config/algolia.yml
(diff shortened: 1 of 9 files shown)


dzello:
Great to see this going live and the faster search speed 🐎, happy I could make the suggestion! 🙌

leggetter:
🙌

PR #549, opened 2018-01-16
"submitOnHash" should be boolean

sbeliakova:
## Description

"submitOnHash": "true" was changed to "submitOnHash": true because submitOnHash should be boolean


adambutler:
👍  Looks great!

🎉 Thank you very much for your contribution and for helping to improve Nexmo Developer.

PR #554, opened 2018-01-24
Remove lots of uses of the word "simply"

tommorris:
## Description

The use of the word 'simply' is redundant: if something is simple, people will know it already. If it isn't, it is marketing bumpf and shouldn't be in technical documentation. Either way, it doesn't help the user and risks infuriating them when it is incorrect.

There are quite a few more uses of the word "simple", but this is a good start.


leggetter:
👍

jameshfisher:
The revolution has begun ...

PR #576, opened 2018-02-14
Adds the ability to include scripts into head via environment variables

adambutler:
## Description

Adds the ability to include scripts into head via environment variables

## Deploy Notes

None


adambutler:
cc/ @leggetter

PR #581, opened 2018-02-16
Remove VAPI modals

### LINKED ISSUE #517: Expose phone, dtmfAnswer and number within visible text in the handle call status page
## Description

https://developer.nexmo.com/voice/voice-api/guides/handle-call-status

`phone`, `number` and `dtmfAnswer` values are hidden in a dialog and it makes them very difficult to discover. e.g. if you search for "dtmlAnswer" and end up on that page you don't know where that text is within the page.

screen shot 2017-12-04 at 10 11 02

Move out of dialog. All dialogs should be removed anyway.

tommorris:
Fixes #517


mheap:
Closing due to new voice docs

PR #582, opened 2018-02-16
Improving description of error code formatting

tommorris:
## Description

Fixes DOCS-263.


adambutler:
This change seems to change the formatting when the original issue describes that there are multiple possible formats.

tommorris:
This is actually fixed in #585.

PR #585, opened 2018-02-21
Clarified that invalid_parameters is optional

tommorris:
Fixes DOCS-263 until we move VAPI to OAS3.


tommorris:
@adambutler ping - can you approve this?

PR #573, opened 2018-02-14
Update Ruby examples to nexmo-ruby v5

timcraft:
PR to update the Ruby code examples to nexmo-ruby v5--there are a number of backwards incompatible changes from v4, as detailed here:

https://github.com/Nexmo/nexmo-ruby/blob/master/CHANGES.md#500
https://github.com/Nexmo/nexmo-ruby/blob/master/5.0-Upgrade.md

PR #592, opened 2018-03-05
Updated Ruby SMS send example to line up with current gem docs

adamzendesk:
## Description

Current Ruby example does not use current naming conventions & doesn't line up with the gem docs.

## Deploy Notes

N/A - Documentation update


adambutler:
Hey @adamzendesk, many thanks for contributing here, however, these changes had already been picked up by another PR #573 -> #606.

PR #599, opened 2018-03-07
[WIP] Error codes

adambutler:
## Description

🚧 Work in progress.

Some R&D about putting API error handling into Nexmo Developer.

## Deploy Notes

None

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -83,10 +83,10 @@ gem 'icalendar', require: false
 #
 # Development & staging environments may use a dependency from a repo:
 #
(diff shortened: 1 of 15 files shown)


mheap:
@adambutler I've added support for product specific error codes. I've not written any ruby in a while, so if you can review https://github.com/Nexmo/nexmo-developer/pull/599/commits/92537d3e3f09bbd9a3c028631e90337ec0666cf9 and we could find some time on Tues/Weds for a call to point me in the direction of writing tests for a rails controller that'd be awesome

PR #604, opened 2018-03-09
HTTP Basic auth support for OAS3

tommorris:
This is required for new APIs and has been separated off for separate merging into Nexmo Developer.


app/views/open_api/_endpoint.html.erb
```diff
@@ -49,6 +49,41 @@
         </ul>
       <% end %>
 
+      <% if endpoint.security_schemes[0] == {"type" => "http", "scheme" => "basic"} %>
```
adambutler: The problem with this is that it will not detect basic auth in all cases and requires the s`security_scheme` to be the first in the array with the exact hash provided in the test.

The `endpoint#jwt?` method in the `oas_definition` can be used as a reference of all the concerns that need to be considered:

image

This block could also be combined with the JWT block [here](https://github.com/Nexmo/nexmo-developer/pull/604/files#diff-32a6658c43fd0e18353bb1bccb5f25a7R17).

I'd suggest:

- Changing `endpoint.security_schemes[0] == {"type" => "http", "scheme" => "basic"}` to `endpoint.basic_auth?`
- I'll add this `endpoint#basic_auth?` method into the `oas_parser` library.
- Clean up this block sharing HTML where possible with the block immediatly above.

adambutler: Update: I've added support in the `oas_parser` dependency. See PR at https://github.com/Nexmo/oas_parser/pull/4

You can test against this branch by using changing the dependency in the Gemfile to be:

```ruby
gem 'oas_parser', github: 'Nexmo/oas_parser', branch: 'basic-auth'
```

PR #612, opened 2018-03-16
WIP - Audit Event API Reference

adambutler:
## Description

Adds Audit API Reference.

## Notes

This is for the definition that goes in `_open_api`. The definition in the `Nexmo/api-specification` should also be updated, then make a new release of the `api-specification` gem and then bump that gem's version in NDP https://github.com/Nexmo/nexmo-developer/blob/master/Gemfile#L125


tbedford:
Testing at http://localhost:3000/api/audit:

* Links in the TOC (left-side navigation) don't work.

tbedford:
API Ref docs specify Base64 encoded API key and secret joined by a colon. The Wiki notes say a JWT should be used. So far I can't get either approach to work.

tbedford:
Closed as _open_api/definitions/audit.yml is now part of PR #867

PR #613, opened 2018-03-16
SIP forwarding doc: explain timeouts and use SIP URIs

giavac:
## Description

Timeouts for forwarding to SIP URI is a new feature, and this PR documents it.
There's a minor change in the examples of SIP URI, which now include the `sip:` scheme.


sammachin:
I've also just tested the CLI and it works fine with this new format eg:
`nexmo link:sip 447700900000 sip:2000@example.com,sip:1234@example.org`

giavac:
Thank you.

PR #623, opened 2018-03-23
updating stitch js sdk docs to v0.1.20

AlexLakatos:


AlexLakatos:
@ericgiannini no advantage. the for loop reversed though so I needed it in order to populate the event history in the correct order

ericgiannini:
@AlexLakatos, Do you delete the merged branches later?

AlexLakatos:
yes. like now 😁

PR #625, opened 2018-03-26
Python code samples: switch to use requests

### LINKED ISSUE #48: Ensure all code samples are using supported libraries where possible
- [ ] https://developer.nexmo.com/voice/guides/create-conferences
- [ ] https://developer.nexmo.com/voice/guides/record-calls-and-conversations
- [ ] https://developer.nexmo.com/voice/guides/outbound-calls
- [ ] `creating-a-custom-call-or-conversation-for-each-user`

Anything containing a `generate_jwt` function call probably needs to get splatted too.

tommorris:
All code examples using urllib2 have been switched over to use requests except one.

The remaining one uses request signing for SMS and would potentially be too complex to move to requests.

Partly fixes #48.

### DIFF
--- a/_examples/api/application/create-an-application/Python
+++ b/_examples/api/application/create-an-application/Python
@@ -1,13 +1,7 @@
-import urllib
-import urllib2
-import json
+import requests
 
-base_url = 'https://api.nexmo.com'
-version = '/v1'
-action = '/applications/?'
-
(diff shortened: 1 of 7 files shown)


mheap:
Merging as it's better than what we have, but we should update all content to use a client library where possible

PR #626, opened 2018-03-26
Connect Two Users: use libraries

tommorris:
Replaces long-winded Python and Ruby code samples with library calls

(Doesn't fix PHP)


mheap:
Closing as we have new voice guides on the way

PR #646, opened 2018-04-09
Workaround: Show NCCO reference

ChrisGuzman:
The controller is interpreting the :document as "voice" not "ncco" so
we'll check the :code_language to see if the user is actually looking
for the the ncco reference.


ChrisGuzman:
Smoke tested it and it works: https://nexmo-developer-staging-pr-646.herokuapp.com/api/voice/ncco
vs
https://developer.nexmo.com/api/voice/ncco

PR #656, opened 2018-04-17
Handle code_language conflicts in the router

adambutler:
## Description

Changes the handling of the optional `code_language` parameter used by many routes to be removed when a document shares the same filename as a linkable code language.

By moving this concerns to a routing constraint the behaviour can easily be adopted by different routes that support the `code_language` parameter.

This change fixes an issue introduced in #629 and temporarily patched by c859298bfe9b9aa76c8d53e57bf2239bb5b3686e.

## Deploy Notes

- [ ] Test on staging


lornajane:
@ChrisGuzman You approved these changes, do you think they're still in a state where we should merge? (If so, go ahead :) )

ChrisGuzman:
I'll merge in latest master and make sure it works and is needed and will decide to merge if that's the case

ChrisGuzman:
I don't think we need this change anymore since it was a bugfix for a file that is no longer user.

PR #661, opened 2018-04-19
Messages API notice

adambutler:
## Description

Boiler place for Messages API notice.

## Deploy Notes

None


tbedford:
Closing as we have Notice entries in this file already for Messages and Workflows API.

PR #666, opened 2018-04-23
Olympus  update apr

hughhopkins:
## Description

An update to notify customers of Messages and Workflows API change by adding message_category.

### DIFF
--- a/_documentation/messages-and-workflows-apis/messages/overview.md
+++ b/_documentation/messages-and-workflows-apis/messages/overview.md
@@ -132,7 +132,7 @@ For Facebook Messenger, Nexmo sends the `response` type by default.
 
 For Viber Service Messages, Nexmo sends the `transaction` type by default.
 
-The defaults can be overridden by using the channel specific property. For Facebook Messenger the possible values for `message_category` are `response`, `update` and `message_tag`. If `message_tag` is used, then an additional `message_tag` for that type needs to be added. A full list of the possible tags is availble on [developer.facebook.com](https://developers.facebook.com/docs/messenger-platform/send-messages/message-tags). For Viber Service Message the possible values are `transaction` and `promotion`. The first message to a user on Viber Service Messages must be a `transaction` one.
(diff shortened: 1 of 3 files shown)


tbedford:
The note appears to display correctly at the top of the page. Notices has also been added to the Contents list. 

The link to `Notices` does _not_ look correct:

https://developer.nexmo.com/messages-and-workflows-apis/workflows/overview#notices

But testing locally the notice topic is at:

http://developer.test/messages-and-workflows-apis/messages/overview#notices

With the specific notice at:

http://developer.test/messages-and-workflows-apis/messages/overview#adding-message-category

> On the 7th May Facebook Messenger will make it mandatory

Might be worth specifying the year.

> The use of different tags enable the business to send messages for different use cases.

Change 'enable' to 'enables'.

> For example, with Viber Service Messages it enables the business to send promotional content; with Facebook Messenger it enables updates to be send after the 24+1 window.

Better as: 

For example, with Viber Service Messages, tagging enables the business to send promotional content.  With Facebook Messenger tagging enables updates to be sent after the 24+1 window.

Note also that 'to be send' has been corrected to 'to be sent'.

> To reduce the burden to the developer and a breaking change in the Messages API we will set defaults for each channel.

Change 'we' to Nexmo.

Change 'we will set defaults' to 'Nexmo will set defaults' (or 'Nexmo has set defaults' if we have already done so). 

> For Facebook Messenger we send the `response` type by default.

Change 'we send' to 'Nexmo sends'.

> For Viber Service Messages we send the `transaction` type by default.

Change 'we send' to 'Nexmo sends'.

> If message_tag is used then an additional message_tag for that type needs to be added.

Add comma after 'is used'.

> A full list of the possible tags are availble on developer.facebook.com  . 

Change 'are availble' to 'is available' and note typo corrected.

> These defaults will be implemented on the 7th May at 12:00 GMT. 

Also please specify the year to be pedantic.

> It will also be possible to override the defaults as well on this date. 

On this date? Do we mean after 12.00GMT? Please clarify.

hughhopkins:
@tbedford updated copy based on your feedback.

PR #672, opened 2018-04-25
Removed wip file as does not display and duplicate.

tbedford:
## Description

This page was broken:

https://developer.nexmo.com/voice/voice-api/building-blocks/connect-a-phone-call

I removed the page from source as it appears to be labelled "wip". Also it appears to duplicate an existing topic 'connect an inbound call'.


leggetter:
This has been sitting broken for too long. Thanks for fixing @tbedford. I'm going to merge too to final see this fixed.

PR #691, opened 2018-05-04
Add Nuance voices, including SSML support column

mheap:
Add Nuance voices, including SSML support


mheap:
@JonathanFarrow117 added a 👍 via Slack

PR #706, opened 2018-05-13
Review suggestions: Connecting Inbound Calls docs

lornajane:
## Description

Just a little bit of tidying up on this page: https://developer.nexmo.com/voice/voice-api/building-blocks/connect-an-inbound-call

I have a bunch of questions though:
 - is there a preferred format for person/tense in these docs?  My main choices are "now do this, now do that", "next you need to do a thing", "let's/we'll set up this other thing next" or "the other thing will be set up next".  If there's no preference, don't worry!
 - I hadn't seen the `nexmo` command line tool before, is there a good resource that should be linked here for others in the same situation?
 - in the second code sample, is there somewhere I can find out more about those placeholder variables?  I wasn't sure how to explain what the values of those should be, but I think we should try to improve this part. I don't feel like I could build it from what we have (but I didn't try yet!)
 - I moved subtitles out of the tabbed example because I thought it looked weird and also that most of the titles/words would work when we add in the other code language examples here.  Is there a general level of detail in the tabs that you find works well?  I have mostly noticed examples with just code so far.

Other comments:
 - let's try not to say things are easy!  It's even easier for a developer to get snagged on something that should be simple and feel like a complete idiot, and I like to not reinforce that.
 - we talked about this when I was in the office: I added a little bit on the aim of the tutorial but even more examples of when this is useful would be good I think.  I was a little bit mystified as to why this would even be useful.  Is there a way of putting a recording at the other end of the outbound phone call or something that I can link to?  It would make it both more fun and more finished!

### DIFF
--- a/_documentation/voice/voice-api/building-blocks/connect-an-inbound-call.md
+++ b/_documentation/voice/voice-api/building-blocks/connect-an-inbound-call.md
@@ -4,20 +4,23 @@ title: Connect an inbound call
 
 # Connect an inbound call
 
-Connecting two calls with Nexmo is easy. In this example we'll accept an inbound call and make an outbound call that will be connected.
+Connecting two calls with Nexmo is a common use case. In this example we'll accept an inbound call, then make an outbound call and connect the two.
 
-Replace the following variables in the example below.
+To use the following example you we will need to replace a variable with the number to call.
 
 Key |	Description
 -- | --
 `YOUR_SECOND_NUMBER` |	The number you wish the inbound caller to be connected to.
+
+## Set up the server
+
+To respond to the incoming call, we'll need to be able to receive a webhook so the first step is to prepare the server.
 
 ```tabbed_content
 source: '_examples/voice/connect-an-inbound-call'
 ```
 
 You'll need to expose your server to the open internet. During development you can use a tool like [Ngrok](https://www.nexmo.com/blog/2017/07/04/local-development-nexmo-ngrok-tunnel-dr/) to do that.
-
 
 ## Associate an application to your webhook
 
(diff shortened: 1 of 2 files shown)


mheap:
@lornajane Thanks for raising this. I've added responses to your questions below.

Unfortunately our voice documentation has been overhauled and will be released in the next week or two. In the interests of keeping the voice docs as stable as possible I'm going to close this PR (but I'd love to continue the discussion on the topics raised)

---

> is there a preferred format for person/tense in these docs? My main choices are "now do this, now do that", "next you need to do a thing", "let's/we'll set up this other thing next" or "the other thing will be set up next". If there's no preference, don't worry!

I don't think there's any _official_ documentation on this, but for reference documentation we tend to use "now do this, now do that" and for tutorials it's "we'll set up this other thing next".

> I hadn't seen the nexmo command line tool before, is there a good resource that should be linked here for others in the same situation?

We need to do a better job promoting this. I've made a note to add a section to the docs on the CLI tool

> in the second code sample, is there somewhere I can find out more about those placeholder variables? I wasn't sure how to explain what the values of those should be, but I think we should try to improve this part. I don't feel like I could build it from what we have (but I didn't try yet!)

The variables are specific to this code example. What additional information would you expect to see here?

> I moved subtitles out of the tabbed example because I thought it looked weird and also that most of the titles/words would work when we add in the other code language examples here. Is there a general level of detail in the tabs that you find works well? I have mostly noticed examples with just code so far.

I did the same myself when I worked with this building block on another branch 🙌 


> let's try not to say things are easy! It's even easier for a developer to get snagged on something that should be simple and feel like a complete idiot, and I like to not reinforce that.

Agreed! We've been removing language like this as we encounter it

>  I added a little bit on the aim of the tutorial but even more examples of when this is useful would be good I think. I was a little bit mystified as to why this would even be useful. Is there a way of putting a recording at the other end of the outbound phone call or something that I can link to? It would make it both more fun and more finished!

This is where things become a little more complicated. Building blocks are intended for more experienced users that understand the use case (see https://developer.nexmo.com/contribute/structure/overview). What we could do is write more in depth content such as a tutorial and link to it from the building block.

PR #711, opened 2018-05-18
Add rate limit error for Redact

mheap:
## Description

Add a new API error definition and update the OAS doc to contain HTTP 429


mheap:
@13webstreet I don't see a reason not to expose the rate limit. It allows customers to build systems to send at maximum throughput if needed. I can remove it if you have concerns, otherwise I'll merge as-is

PR #722, opened 2018-05-29
Create chatapp-accounts.yml

panayiotispanteli:
## Description

Chatapp-accounts OpenAPI Spec

## Deploy Notes


AFavreNexmo:
Yes that's good, just to check with @hughhopkins about Whatsapp, not sure we want to make it public right now.

mheap:
@panayiotispanteli Thank for for this, it's an awesome start. I've made a few changes to get it rendering in Nexmo Developer on the `add-chatapp-accounts` branch.

There's still work to be done, so I've re-raised the PR as a branch on NDP at #731

PR #729, opened 2018-05-31
Update chatapp error codes

panayiotispanteli:
## Description

Update chatapp error codes

### DIFF
--- a/_open_api/errors/messages-and-workflows-apis/messages.md
+++ b/_open_api/errors/messages-and-workflows-apis/messages.md
@@ -9,7 +9,8 @@ Code | Text | Meaning
 `1000` | Throttled | You have exceeded the submission capacity allowed on this account. Please wait and retry.
 `1010` | Missing params | Your request is incomplete and missing some mandatory parameters.
 `1020` | Invalid params | The value of one or more parameters is invalid.
-`1021` | Invalid tag | The tag value is invalid
+`1021` | Invalid tag | The tag value is invalid.
+`1022` | Invalid template parameters | Some of the parameters of the template were invalid.
 `1030` | Internal error | There was an error processing your request in the Platform.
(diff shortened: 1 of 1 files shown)


_open_api/errors/messages-and-workflows-apis/messages.md
```diff
@@ -44,3 +45,4 @@ Code | Text | Meaning
 `1330` | Unknown | An unknown error was received from the carrier who tried to send this this message. Depending on the carrier, that to is unknown. When you see this error, and status is rejected, always check if to in your request was valid.
 `1340` | Outside of the allowed window | This message is sent outside of allowed response window.
 `1350` | Phone matching fee not paid | Requires phone matching access fee to be paid by the the Facebook Page.
+`1360` | TTL was activated | TTL was activated, no callbacks and no charge will be issued.
```
mheap: What does this mean? What is a TTL?

panayiotispanteli: Time To Live. It is a Viber-specific param where you specify a ttl time in the request to viber, after which viber will stop attempting to deliver the message and instead will send us a "failed" callback

mheap: Could we put that information in the table too? Mark it as Viber specific

panayiotispanteli: @mheap, @hughhopkins sent you a message on slack about this. Have you seen it?
just repeating it here:
```
We're also trying to abstract the errors codes and we try where possible to not mark something as a channel specific error. If we take that logic we should be marking most of the others as SMS specific as well.

I think we should leave as is. However, we will need to review our error codes and ensure that what we say in the docs matches the API and we are providing helpful messages. Just that's not the biggest priority right now.
```

PR #731, opened 2018-05-31
Add chatapp-accounts first draft

mheap:


AFavreNexmo:
Have you made the changes? Could you open it on top on Panayiotis branch to see the differences?

mheap:
@afavre The changes have been made in this PR. I can't raise it on top of @panayiotispanteli's branch as it's a fork. A summary of the changes are:

* Remove the example URLs as they're intended to point to quick start files that are loaded in, not API references
* Added section grouping (`x-group`) for messenger, viber and accounts instead of each entry having it's own group
* Added `x-groups` right at the bottom which provides header titles etc

AFavreNexmo:
Thanks for the clarification.

AFavreNexmo:
@mheap do you know you will have time to finish on it? We have customers asking us for the documentation, to integrate chatapp-accounts on their side

mheap:
The only thing blocking this is a better resource name. I proposed a few in Slack but the conversation died. We can pick up tomorrow

lornajane:
Can we recap on the naming things problem?  Is that still the only blocker?

mheap:
It's still the only blocker. There's an internal discussion happening around it

AFavreNexmo:
We have more and more people asking us about this documentation. I don't think we should block people integration for a name. It could be the same as we have done for Workflows, and change it after the Dispatch.
Anyway, if we change the name, it will require more that a change in the documentation.

mheap:
I've spoken to Hugh and we're going to catch up next week.

I think we're all on the same page at the moment, in that we can document for early access with the caveat that we revisit and rename before GA

AFavreNexmo:
@mheap where are we with this PR? It is open since end of May, we should deploy ASAP, it is blocking customers integration.

lornajane:
Hi @AFavreNexmo , I'm afraid I don't have an update for you on this and Michael is on leave for a week or so. Please let us know if there's anything we can do in the meantime.

AFavreNexmo:
We can wait 1 more week for when he is back 😊, thanks @lornajane

panayiotispanteli:
@arisgoulia @AFavreNexmo @mheap 
Couldn't edit the existing PR so I created a new PR for external-accounts.yml to account for applications:
https://github.com/Nexmo/nexmo-developer/pull/932

mheap:
#932 replaces this. Closing

PR #743, opened 2018-06-11
Added new IPs

peter-nexmo:
## Description

## Deploy Notes


lornajane:
These IPs are already listed in our SIP docs (I think they were picked up in the work to re-release the Voice Docs), so this is no longer needed.  Thanks anyway @peter-nexmo !

PR #752, opened 2018-06-14
Make Docker build node_modules instead of doing it on the host.

cr0wst:
## Description

Update the Docker settings to make Docker do the building of `node_modules` instead of having to do it on the host system. I believe that the `ADD . /app` step in the Dockerfile wasn't working correctly as we are mounting the current working directory as `/app` which hides the `/app` directory in the container created in the `build` process.

Docker will now run `yarn install` and create a `node_modules` folder at `/app/node_modules`. When the web service is initialized it will mount both `.:/app` and then `/app/node_modules` which will expose the `node_modules` folder created on the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,13 +6,20 @@ RUN apt-get update -qq && apt-get install -y build-essential libpq-dev lsof
 
 # Install modern NodeJS
 run curl -sL https://deb.nodesource.com/setup_8.x | bash -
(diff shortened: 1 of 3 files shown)


cr0wst:
More information on how Docker handles volumes between the build and up step can be found here: http://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html

PR #753, opened 2018-06-18
Tony contribs review

tbedford:
## Description

Initial edits to Contrib guidelines plus first draft of writing style guide.


tbedford:
I wanted to change the filenames for write-the-docs.md and styleguide.md but they appear hard-coded into the Ruby code and Chris and I currently aren't sure exactly what the side effects might be of changing the file names. There's a Ruby method called write_the_docs for example. So, can we get this in as is and we'll figure out the file naming at some future point?

PR #754, opened 2018-06-18
Send SMS with Unicode Building Block

tbedford:
## Description

Send SMS with Unicode building block with Curl and Python example code. 

Need to add at least Node code. [UPDATE: will be added some time in the future by DevRels]


tbedford:
This is the old style of doing building blocks.

lornajane:
Looking at adding the PHP and Node examples here - should we also be switching over to the new building blocks approach?  I can do that while I'm at it if you think so

lornajane:
This work is a bit part-done while I wait for building blocks to be merged into the various quickstart repos - then I can close the loop and include the code examples.

lornajane:
This was superceded by #846 since I had some issues rebasing this older branch

PR #766, opened 2018-06-22
Use updated ACL

ChrisGuzman:
## Description

The ACL in the Stitch quickstarts are out of date. This updates them.

## Deploy Notes


ChrisGuzman:
Huge thanks to @mi-po for pointing this out!

PR #768, opened 2018-06-22
fix table of content anchors

BuonOmo:
## Description

Some anchors had typo errors, leading to un-clickable links.


mheap:
🙌 @BuonOmo! Thanks for the contribution

PR #787, opened 2018-07-02
Fix pricing by country response examples

timcraft:
The keys in the pricing by country [JSON/XML Response examples](https://developer.nexmo.com/api/developer/pricing#response) don't match what's returned by the API (and documented further down under [Keys and Values](https://developer.nexmo.com/api/developer/pricing#keys-and-values) and [Common elements in responses](https://developer.nexmo.com/api/developer/pricing#common-elements-in-responses)).

### DIFF
--- a/_examples/api/developer/account/pricing/per-country-pricing/JSON
+++ b/_examples/api/developer/account/pricing/per-country-pricing/JSON
@@ -1,14 +1,16 @@
 {
-  "country": "GB",
-  "name": "United Kingdom",
-  "prefix": "44",
-  "mt": "0.03330000",
+  "countryCode": "GB",
+  "countryDisplayName": "United Kingdom",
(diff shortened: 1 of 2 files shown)


lornajane:
I think this is valuable but I'm struggling to review it because the original source no longer exists, maybe a deleted branch or something similar? We should definitely make some changes as I agree that the docs don't match reality! I think I will need to create new changes since we can't review this, we can only blindly merge.

timcraft:
@lornajane Sorry I deleted the fork. I've rebased the change on master, it still applies cleanly:

https://github.com/timcraft/nexmo-developer/tree/fix-per-country-pricing-examples

Can open a fresh pull request if that helps?

timcraft:
Closing in favour of #916

PR #794, opened 2018-07-05
Updated verify overview

13webstreet:
## Description

Added links in the guides section to the new articles created by Eric and Chris.

### DIFF
--- a/_documentation/verify/overview.md
+++ b/_documentation/verify/overview.md
@@ -25,6 +25,9 @@ TTS messages are read in the locale that matches the phone number. (For example,
 ## Guides
 
 * [Verify a user](/verify/guides/verify-a-user)
+* [Verify API implementation guide](https://www.nexmo.com/blog/2018/05/10/nexmo-verify-api-implementation-guide-dr/)
+* [Implementing the Verify API on iOS](https://www.nexmo.com/blog/2018/05/10/add-two-factor-authentication-to-swift-ios-apps-dr/)
+* [Implementing the Verify API on Android](https://www.nexmo.com/blog/2018/05/10/add-two-factor-authentication-to-android-apps-with-nexmos-verify-api-dr/)
 
 ## Tutorials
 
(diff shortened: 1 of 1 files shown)


_documentation/verify/overview.md
```diff
@@ -25,6 +25,9 @@ TTS messages are read in the locale that matches the phone number. (For example,
 ## Guides
 
 * [Verify a user](/verify/guides/verify-a-user)
+* [Verify API implementation guide](https://www.nexmo.com/blog/2018/05/10/nexmo-verify-api-implementation-guide-dr/)
+* [Implementing the Verify API on iOS](https://www.nexmo.com/blog/2018/05/10/add-two-factor-authentication-to-swift-ios-apps-dr/)
```
tbedford: Could we just say "Using" rather than "Implementing". It's a minor point, but when I first looked at it I thought the Verify API itself was being implemented.

_documentation/verify/overview.md
```diff
@@ -25,6 +25,9 @@ TTS messages are read in the locale that matches the phone number. (For example,
...
+* [Implementing the Verify API on Android](https://www.nexmo.com/blog/2018/05/10/add-two-factor-authentication-to-android-apps-with-nexmos-verify-api-dr/)
```
tbedford: Can we again use "Using" rather than "Implementing"?

leggetter:
I don't believe we should link to blog posts as if they are official pieces of documentation. If we want to link to them they should be clearly flagged as other types of docs. Maybe a new section for things such as this?

lornajane:
On the voice API, I wondered if it would be appropriate to mix blog posts and tutorials together (maybe with another title like "Further reading" or "See also").  Would that work here too?

tbedford:
I have linked to blog posts, such as for example, our Ngrok tutorial. However, I do usually explicitly indicate it's a blog post. I've also sometimes created a 'Other resources' section with link(s) to blog posts.

PR #811, opened 2018-07-11
OAS Response Descriptions

mheap:
## Description

This PR adds a collapsible "View response field descriptions" link to all OAS documents that return a `2xx` response. The descriptions were either pre-existing or sourced from the pre-OAS documentation pages.


mheap:
See https://nexmo-developer-staging-pr-811.herokuapp.com/api/voice#getCall for an example

PR #812, opened 2018-07-12
Add building block status report

mheap:
This is an admin-only report that shows the current state of the building blocks. We've been iterating rapidly and have three different ways of providing blocks.

The (admittedly hacky) report shows coverage across all available blocks and distinguishes between the different implementation methods using colours


mheap:
> why is the product dropdown pink?

That's the default rendering for Nexmo Developer. Not sure why it's pink, but it'll change soon when the Volta design system is rolled out

> what do the colours mean? Is a yaml example better than a file one, it seems like these are traffic lights? Can we add another indicator (such as a letter code in each field) rather than relying purely on colours?

Yep, traffic lights. `yaml` is our preferred way to handle blocks today so they're green. Config based is legacy (so yellow. File based breaks our convention of having the code in a quickstart repo as they're embedded directly in Nexmo Developer and so they get orange.

> are there situations where some examples would make no sense? Such as maybe curl when there's some programmatic work needed? 

If I've understood correctly, building blocks should be simple enough that there is no programmatic work required.

> Would it be useful to indicate where things are intentionally omitted?

💯 it definitely would. I'm open to suggestions on how to do this

> related: what does it mean to have an xml version of something where the others are a programming language?

Those are response "building blocks". They're currently shown as they're mixed in with the building blocks in the `_examples` folder. They're not really building blocks and should be moved elsewhere / replaced with OAS3 definitions

mheap:
Added labels to the blocks as requested https://nexmo-developer-staging-pr-812.herokuapp.com/coverage

lornajane:
Looks good to me, thanks for adding those extra labels, it makes it easier to understand

PR #816, opened 2018-07-12
Bump Nexmo Api Spec gem

ChrisGuzman:
## Description

The new API Spec includes updates to the Stitch OAS doc

## Deploy Notes

Just bumping a gem


ChrisGuzman:
The bug breaking the API pages has been fixed. Will merge after CI.

PR #822, opened 2018-07-16
Add Link to Source for Code Example in Building Blocks

cr0wst:
## Description

Added link generation to the building block filter which links to the underlying quickstart example that the source came from. This allows users to click and see the full working example in the quickstart repo from the building blocks.

The URL is formatted from the `.yml` file's source parameter and is modified to support GitHub's linking. Only impacts building blocks that are generated from config.

image

### DIFF
--- a/app/filters/building_block_filter.rb
+++ b/app/filters/building_block_filter.rb
@@ -20,7 +20,8 @@ class BuildingBlockFilter < Banzai::Filter
           dependency_html = generate_dependencies(lexer.tag, config['dependencies'])
       end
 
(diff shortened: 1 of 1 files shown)


lornajane:
I think you read my mind, I was thinking about this late last week!  Is there grounds for highlighting the range we actually show in the building block or is that a step too far?

cr0wst:
I had similar thoughts about line highlighting! It _should_ follow the same line selection process as the code snippet process. By that I mean:
- `from_line` must be present
- `to_line` either exists or select the max line from the local file

PR #824, opened 2018-07-17
Building block curl support

mheap:


mheap:
This is in anticipation of the move + rename from nexmo-community to nexmo for the quickstarts

PR #825, opened 2018-07-17
Volta

mheap:
Creating a review app for Volta (the visual refresh using the Vonage design system)


leggetter:
Items in tabs are `display: none`

screen shot 2018-07-17 at 13 38 13

mheap:
That's intended, it's a collapsible section. I mentioned to @BennyZ28 today that the indicator that it's collapsible is too far away on the right hand side and they're going to take a look

leggetter:
@mheap there is no expand/collapse indicator for me

screen shot 2018-07-17 at 13 46 58

mheap:
So there isn't! There was this morning when we did a run through

It's on the list to be fixed 👍

PR #829, opened 2018-07-17
Refactor building_block filter

mheap:
* Use polymorphism for rendering dependencies and run_command
* Extract HTML in to ERB templates, rather than keeping it in the class

---

There's still more work to be done (e.g. HTML in the render classes), but this is a big step in the right direction to me

### DIFF
--- a/app/filters/building_block_filter.rb
+++ b/app/filters/building_block_filter.rb
@@ -3,13 +3,15 @@ class BuildingBlockFilter < Banzai::Filter
     input.gsub(/```single_building_block(.+?)```/m) do |_s|
       config = YAML.safe_load($1)
 
+      @renderer = get_renderer(config['language'])
+
       lexer = CodeLanguageResolver.find(config['language']).lexer
 
-      application_html = generate_application_block(config['application']);
+      application_html = generate_application_block(config['application'])
 
       # Read the client
       if config['client']
-          highlighted_client_source = generate_code_block(config['language'], config['client'], config['unindent'])
+        highlighted_client_source = generate_code_block(config['language'], config['client'], config['unindent'])
       end
 
       # Read the code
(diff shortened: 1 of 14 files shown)


app/filters/building_block_filter.rb
```diff
@@ -24,29 +26,16 @@ def call(input)
...
-        </h3>
-
-        <div id="#{id}" class="collapsible-content" style="display: none;">
-          <p>Create a file named <code>#{config['file_name']}</code> and add the following code:</p>
-          <pre class="highlight bash"><code>#{highlighted_client_source}</code></pre>
-        </div>
-      HEREDOC
+          id = SecureRandom.hex
+          erb = File.read("#{Rails.root}/app/views/building_blocks/_configure_client.html.erb")
+          client_html = ERB.new(erb).result(binding)
```
mheap: `binding` is magical! It automatically takes all scoped variables and makes them available to the view

app/filters/building_blocks_filter.rb
```diff
@@ -112,10 +112,6 @@ def content_from_source
 ```
       HEREDOC
 
-      if @config['title']
```
mheap: We didn't use this anywhere

app/filters/building_blocks_filter.rb
```diff
@@ -140,39 +136,6 @@ def resolve_language(contents)
     end
   end
 
-  def format_code(contents)
```
mheap: This was dead code, copied in from tabbed content that we don't use

app/filters/building_block_filter.rb
```diff
@@ -82,76 +71,37 @@ def generate_code_block(language, input, unindent)
...
-                'text' => 'Add the following to <code>build.gradle</code>:',
-                'code' => dependencies.map { |d| "compile '#{d}'" }.join("<br />")
-              }
-          },
-      }
-
-      deps = dep_managers[language].call(dependencies)
-      deps = {'code' => deps} unless deps['code']
-      deps['text'] = '' unless deps['text']
+  def get_renderer(language)
```
cr0wst: I like this a lot better than what we had before.

app/views/building_blocks/_configure_client.html.erb
```diff
@@ -0,0 +1,12 @@
+<h3 class="collapsible">
+    <a class="js-collapsible" data-collapsible-id="<%= id %>">
+        Initialize your dependencies
+    </a>
+</h3>
+
+<div id="<%= id %>" class="collapsible-content" style="display: none;">
+    <p>Create a file named <code><%= config['file_name'] %></code> and add the following code:</p>
```
cr0wst: Most likely outside the scope of the PR but I'm wondering how feasible it would be to have the verbiage distinguish between "Create a file named" and something like "Create a class named".

I don't want to get too granular and have the config files become unwieldy but it would be more accurate for Java and probably .NET.

app/filters/building_block_filter.rb
```diff
@@ -82,76 +71,37 @@ def generate_code_block(language, input, unindent)
...
-
-        <pre class="highlight bash"><code>$ #{command}</code></pre>
-
-      HEREDOC
-      end
-
-      command
-    end
-
     def generate_application_block(app)
```
cr0wst: You didn't touch it but I think this method is indented more than the others around it. You could probably fix that while you're here. :)

cr0wst: FWIW I ran auto formatter on RubyMine and it seems like there are quite a few indentation inconsistencies. This one just looks visually out of place.

app/views/building_blocks/_configure_client.html.erb
mheap: We can do it using the language specific renderers rather than the config files

We'd also need to change the `file_name` parameter to be just the class name rather than the full file name. I'm happy to work on that in a separate PR

mheap:
@cr0wst I ran Rubocop and fixed everything it complained about. Should be good to go now

PR #838, opened 2018-07-19
For dotnet, use ASPNET branch when linking to source

lornajane:
## Description

The "View full source" links for .NET examples weren't taking account of the branch name used on that repository.  This fixes those links.


cr0wst:
I agree with @mheap although I also like ternary so even something like
```rb
replacement = code['source'].include?('dotnet') ? '\\0blob/ASPNET/' : '\\0blob/master/' 
```

lornajane:
Thanks!  I was so focused on writing working Ruby code that I lost touch with everything else.  Is this version better?

PR #845, opened 2018-07-23
Add the GSM extended characters when we check if we need to send a message as unicode on the try-it-out page

### LINKED ISSUE #725: Try-it-Out component for SMS concatenation has issues
## Description

The try-it-out on [this page](https://developer.nexmo.com/messaging/sms/guides/concatenation-and-encoding#maximum-number-of-characters
) has an issue:

If you type a '{' character or other character from GSM extended table it will flag that as requiring a type of 'unicode' whereas a type of 'text' is fine. However, note that a character from the GSM extended table does require two characters (ESC + char). In short, characters from GSM standard and extended table are message `type` of `text`. Only characters outside the GSM range require a `type` of `unicode`.

## Related resources

* https://developer.nexmo.com/messaging/sms/guides/concatenation-and-encoding#maximum-number-of-characters
* #701

lornajane:
## Description

Fixes #725 - we weren't checking for the extended characters before deciding that everything except the safe characters must mean we need unicode.  This stops the unicode symbol from turning on if the extended characters are used.

### DIFF
--- a/app/javascript/packs/Concatenation.jsx
+++ b/app/javascript/packs/Concatenation.jsx
@@ -29,7 +29,6 @@ const safeCharacters = [
   '§', 'o', 'à', ' '
 ]
 
-// TODO: Need to realize that these characters do not require Unicode type, but do require two bytes per character.
 // These require two bytes per character: ESC followed by the character
 const extGSMChars = [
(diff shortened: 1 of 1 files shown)


mheap:
Just tested in a review app, nice work!

PR #848, opened 2018-07-26
wip cli page

AlexLakatos:


lornajane:
Can we get some words with this? What's the aim of the CLI page? What stage are we at and what else needs doing? Can others help?

AlexLakatos:
I'd ideally want something like cli.heroku.com but I'd settle for something like cli.angular.io

lornajane:
Given the absence of any content, and knowing that there are changes planned to the CLI tool, I'm closing this for now. This change isn't useful by itself but it's a great idea for the future so let's revisit at an appropriate time?

PR #859, opened 2018-08-01
Update websockets.md

sammachin:
added a bit to clarify that you can send at 8 or 16 Khz


tbedford:
So, to summarize:

- Sampling is 16 bits per sample (2 bytes per sample). 
- Sample time window is 20 ms of audio.

| Sampling rate (samples per second) | Number of samples in 20ms | Bytes |
|----|----|----|
| 8000 | 160 | 160 x 2 = 320 |
| 16000 | 320 | 320 x 2 = 640 |

sammachin:
@tbedford Yep thats correct! nice way of displaying/explaining it, if you can find a way to slot that table into the page I'd go for it

PR #862, opened 2018-08-02
Add Ruby building blocks

mheap:
The definitions need updating by @ChrisGuzman


ChrisGuzman:
This is all good! I've checked the sample app to make sure everything renders correctly and we've added all the correct building blocks;

.repos/nexmo-community/nexmo-ruby-quickstart/voice/record_a_call.rb
```diff
@@ -15,25 +22,23 @@ def parsed_body
 route :get, :post, '/webhooks/answer' do
   [
     {
-      action: 'talk',
```
martyndavies: Curious as to why the _talk_ action is removed here @ChrisGuzman . This would make it inconsistent with the other language building blocks.

ChrisGuzman: The spec doesn't include the talk action: https://github.com/Nexmo/building-block-specs/blob/master/voice/specs/record-a-call-SPEC.md

and 2/3 of the other langs don't have it:
https://nexmo-developer-staging-pr-862.herokuapp.com/voice/voice-api/building-blocks/record-a-call/java
https://nexmo-developer-staging-pr-862.herokuapp.com/voice/voice-api/building-blocks/record-a-call/php

Do you think it should still be added?

martyndavies: No, in that case I think that the action should be removed from the Python example so that everything lines up :)

ChrisGuzman: Cool. Merging this

PR #874, opened 2018-08-10
Fixed broken link and added info on answer_url payload.

tbedford:
## Description

- [ ] Fix broken link to setting webhook endpoint
- [ ] Added info on `answer_url` payload


tbedford:
Yes, I will revisit the whole webhooks topic next week.

PR #876, opened 2018-08-10
Adds Python download recording building block.

tbedford:
## Description

Adds Python building block to download a recording.


lornajane:
Where can I see where this is used?  Also I can see that this is merged into the python repo, but do we need to pull the updates in from those quickstart repos via some process rather than just making the same changes here?  I'm not sure what's the best/safest process here.

tbedford:
I can see get-recording.py in my local .repos but not in the master/.repos. @ChrisGuzman should be able to advise on what the correct process is here.

PR #883, opened 2018-08-13
WIP - Quick mockup of Stitch content.

tbedford:
## Description

A quick mockup of some Stitch content to provide an initial visual reference point for further discussions. It is not meant to be comprehensive - only to indicate the new direction.

> NOTE: This will NEVER be merged.


tbedford:
I'm going to close as this was only for an initial idea of where we were headed.

PR #888, opened 2018-08-14
Fixes + context of how calls fundamentally work

mariatechmaniac:


_partials/stitch/guides/calling-users/javascript.md
```diff
@@ -3,29 +3,29 @@ title: JavaScript
...
-In this getting started guide we'll cover adding call methods to the Conversation we created in the [simple conversation with audio](/stitch/in-app-voice/guides/enable-audio/javascript) getting started guide. We'll deal with member call events that trigger on the application and call state events that trigger on the Call object.
+In this guide we'll cover adding call methods to the application we have created in the [simple conversation with audio](/stitch/in-app-voice/guides/enable-audio/javascript) guide. We'll deal with member call events that trigger on the application and call state events that trigger on the Call object.
 
 ## Concepts
 
-This guide will introduce you to the following concepts.
+This guide will introduce you to the following concepts:
 
-- **Calls** - calling an User in your application without creating a Conversation first
+- **Calls** - a method for calling a User in your application. The method essentially wraps the creation of a conversation, adding users to it and finally the audio enabling and disabling for you. At the same time it also generates a Call object.
```
AlexLakatos: Calls is not a method.

_partials/stitch/guides/calling-users/javascript.md
```diff
@@ -3,29 +3,28 @@ title: JavaScript
...
 - **Call Events** - `member:call` event that fires on an Application
-- **Call Status Events** - `call:status:changed` event that fires on an Application when a status in the Call changes
+- **Call Status Events** - `call:status:changed` event that fires on an Application when the status of the Call object changes
```
AlexLakatos: the whole idea with this section is to be code independent. in the context Call is not an object

_partials/stitch/guides/calling-users/javascript.md
```diff
@@ -95,7 +94,7 @@ showCallMembers(member) {
 }
 ```
 
-We've added some logic in the methods in order to identify the caller as a User in the application or a person calling from a phone. When we receive a phone call into our application, the member is listed as `unknown`.
+We've added some logic in the methods in order to identify the caller as a User in the application or a person calling from a phone (see following guides for how to do that). When we receive a phone call into our application, the member is listed as `unknown`.
```
AlexLakatos: making `(see following guides for how to do that)` a link to the next guide would be useful I think

_partials/stitch/guides/calling-users/javascript.md
```diff
@@ -160,7 +159,7 @@ setupUserEvents() {
 }
 ```
 
-If we want to be notified when the call status changes, we need to add a listener for `call:status:changed` on the Application. Let's update the `app` promise after `login(userToken)` with the code::
+If we want to be notified when the [call status](https://developer.nexmo.com/stitch/in-app-voice/call-statuses) changes like for example started, busy, rejected, answered etc, we need to add a listener for `call:status:changed` on the Application. Let's update the `app` promise after `login(userToken)` with the code::
```
AlexLakatos: links should be relative, so this should be `(/stitch/in-app-voice/call-statuses)`. the call statuses should also be highlighted using "`", `started`, `busy`, `rejected`, `answered`

_partials/stitch/guides/calling-users/javascript.md
```diff
@@ -180,11 +179,9 @@ Now run `index.html` in two side-by-side browser windows, making sure to login w
...
 
-### 1.6 - Calling a Stitch user from a phone
-
-After you've set up you're app to handle incoming calls, you can follow the [PSTN to IP tutorial](https://www.nexmo.com/blog/2018/05/13/connect-phone-call-to-stitch-in-app-voice-dr/) published on our blog to find out how you can connect a phone call to a Stitch user. Now you can make PSTN Phone Calls via the Nexmo Voice API and receive those calls via the Stitch SDK.
+### 1.6 - Bonus: Phone to App Calls
 
-Because we've added logic in this quick start guide in order to differentiate between Users and PSTN phone calls, we don't need to change the quick start code, just set up an NCCO, as shown in the [PSTN to IP tutorial](https://www.nexmo.com/blog/2018/05/13/connect-phone-call-to-stitch-in-app-voice-dr/).
+After you've set up you're app to handle incoming calls, you can follow the [Phone to App Calls Guide](https://developer.nexmo.com/stitch/in-app-voice/inbound-pstn) to find out how you can connect calls from phones, in addition to calls from other app users, to app user via the Voice API.
```
AlexLakatos: links should be relative, so the link should be `(/stitch/in-app-voice/inbound-pstn)`

_partials/stitch/guides/calling-users/javascript.md
```diff
@@ -180,11 +179,9 @@ Now run `index.html` in two side-by-side browser windows, making sure to login w
 
 Thats's it! Your page should now look something like [this](https://github.com/Nexmo/stitch-js-quickstart/blob/master/calling-users/index.html).
 
-### 1.6 - Calling a Stitch user from a phone
-
-After you've set up you're app to handle incoming calls, you can follow the [PSTN to IP tutorial](https://www.nexmo.com/blog/2018/05/13/connect-phone-call-to-stitch-in-app-voice-dr/) published on our blog to find out how you can connect a phone call to a Stitch user. Now you can make PSTN Phone Calls via the Nexmo Voice API and receive those calls via the Stitch SDK.
+### 1.6 - Bonus: Phone to App Calls
```
AlexLakatos: I don't really like calling this a Bonus. I'd suggest we change it to "Calling an App from a phone" if you want to go with the App terminology. But I really think we should use the User terminology here, cause we're not calling a whole application, just a user in one

_partials/stitch/guides/calling-users/javascript.md
tbedford: "Essentially" can safely be omitted.

_partials/stitch/guides/calling-users/javascript.md
tbedford: "code::" Remove one colon here?

_partials/stitch/guides/calling-users/javascript.md
tbedford: "After you've set up you're app". Typo. Should be "After you've set up your app".

"to find out how you can connect calls from phones, in addition to calls from other app users, to app user via the Voice API." I found this sentence fragment a little confusing, but I'm not sure what it should be.

_partials/stitch/guides/calling-users/javascript.md
mariatechmaniac: ok  reverted to the old

_partials/stitch/guides/calling-users/javascript.md
ChrisGuzman: I think the meat of this addition could be added to line 125 since that's where we actually talk about the `call()` method

_partials/stitch/guides/calling-users/javascript.md
```diff
@@ -3,29 +3,29 @@ title: JavaScript
...
+We will use the application we have already created for [the first audio getting started guide](/stitch/in-app-voice/guides/enable-audio/javascript). All the basic setup has been done in the previous guides and should be in place. We can now focus on updating the client-side application.
 
 ### 1.1 - Add call control UI
 
-First, we'll add the UI for a user to call another user, and then see when it's being called. We'll also add UI in order to hang up the call. We'll hide the `call-incoming` and `call-members` using CSS until the user is interacting with a call. Let's add the UI at the top of the conversations area.
+First, we'll add the UI for a user to create and end a call to another user. We'll also add UI in order to be notified for an incoming call. We'll hide the `call-incoming` and `call-members` using CSS until the user is interacting with a call. Let's add the UI at the top of the conversations area.
```
ChrisGuzman: to start and end a call with another user

_partials/stitch/guides/calling-users/javascript.md
ChrisGuzman: what guide should this link to?

_partials/stitch/guides/calling-users/javascript.md
mariatechmaniac: reverted the concept explanation of the 'call'
removed essentially
explained later on during implementation how call() works

_partials/stitch/guides/calling-users/javascript.md
mariatechmaniac: ok renamed to 'Calling an App from a phone'

PR #890, opened 2018-08-14
Clarifications on call states and event object status

tbedford:
## Description

DEVX-503

Clarifications on call states and event object status.


tbedford:
What happened here? Is now showing as no commits and no files changed!

PR #893, opened 2018-08-14
Phone to App Guide - remove Vapi, edit ncco& title

mariatechmaniac:
removed the VAPI section and added it in the voice overview. makes more sense since it will be handling ALL stitch voice stuff pretty much. IP IP calls will soon be done through VAPI as well.
edidet ecplanation of NCCO
changed title for consistency with the rest

### DIFF
--- a/_documentation/stitch/in-app-voice/inbound-pstn.md
+++ b/_documentation/stitch/in-app-voice/inbound-pstn.md
@@ -2,12 +2,9 @@
 title: Calling In-App users from a phone
 ---
 
-# Calling In-App users from a phone
+# Phone to App Calls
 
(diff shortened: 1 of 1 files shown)


_documentation/stitch/in-app-voice/inbound-pstn.md
```diff
@@ -2,12 +2,9 @@
 title: Calling In-App users from a phone
```
ChrisGuzman: @mariatechmaniac if you want to change the title in the sidebar we need to change this line here too

AlexLakatos: I'll take this on as part of the Calling re-write

PR #896, opened 2018-08-16
Disclaimer for unsupported APAC LVNs | PSTN-IP

mariatechmaniac:


AlexLakatos:
shouldn't have merged this in the re-write branch, but I got trigger happy, sorry.

PR #902, opened 2018-08-20
Code Style Fixes (Rubocop)

mheap:
## Description

Apply code style auto fixes, plus some changes that it couldn't make itself (e.g. guard clauses rather than wrapping a block in an `if` statement. Things look ok at first glance, but I'd like to test on a review app and run an automated checker to see if any pages now `500`

## Deploy Notes

There are a *lot* of changes, so please be extra-careful when reviewing this. You'll probably want to test on a review app rather than reading the diff.

### DIFF
--- a/app/filters/code_filter.rb
+++ b/app/filters/code_filter.rb
@@ -23,12 +23,6 @@ class CodeFilter < Banzai::Filter
 
       highlighted_source = highlight(code, lexer)
 
-      line_numbers = (1..total_lines).map do |line_number|
-        <<~HEREDOC
-          <span class="focus__lines__line">#{line_number}</span>
-        HEREDOC
-      end
-
       <<~HEREDOC
         <pre class="highlight #{lexer.tag}"><code>#{highlighted_source}</code></pre>
       HEREDOC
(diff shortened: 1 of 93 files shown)


mheap:
I've played with the review app and tested the following:

* Feedback positive/negative still works
* Slack invitation still works
* OAS renders
* Markdown renders
* Building blocks render
* Legacy API docs render

I'm going to remove the `don't merge` label, but would appreciate someone else trying this out too

app/filters/code_filter.rb
```diff
@@ -23,12 +23,6 @@ def call(input)
 
       highlighted_source = highlight(code, lexer)
 
-      line_numbers = (1..total_lines).map do |line_number|
```
ChrisGuzman: Was this deleted by rubocop or was done manually? Is it something we need?

app/views/layouts/partials/_footer.html.erb
```diff
@@ -53,13 +53,6 @@
         <span class="api-status-status">Loading</span>
       </a>
 
-      <div id="footer-social-links">
```
ChrisGuzman: Why get rid of this?

app/filters/code_filter.rb
mheap: Manually, it's unused

app/views/layouts/partials/_footer.html.erb
mheap: This was a mistake, I'll put it back

PR #903, opened 2018-08-20
Update Java User Input Building Block

### LINKED ISSUE #895: Handle User Input with DTMF - Java Example Broken
## Description

The building block for "Handle User Input with DTMF" for Java does not work correctly. The response that gets sent to `webhooks/dtmf` has a json body. Java users will need to create an additional object in order to deserialize the json and get the `dtmf` property.

Raising an issue because I'm not entirely sure where this extra class should live in the documentation. Maybe I update the Java library to include it as an added functionality?

I expected the correct digit to be read back.

## What actually happened?

`null` was read back.

## Steps to reproduce:

Follow the building block example and call your number.

cr0wst:
## Description

The building block had a reliance on another class that was not documented. This class was added to the example as a static class to help make it more clear.

I'm still not completely happy with this implementation, but it's better than not being correct entirely. Somewhat resolves #895 but see comments there for possible improvements to add extra readability.


mheap:
This all renders fine locally

PR #897, opened 2018-08-16
[WIP] partial for calling JS guides re-write

AlexLakatos:

PR #909, opened 2018-08-22
updating menu items for stitch voice

AlexLakatos:


_documentation/stitch/in-app-voice/guides/ncco-guide.md
```diff
@@ -36,7 +44,7 @@ $ nexmo setup api_key api_secret
 
 ## Setup instructions
 
-For this guide, you'll need to connect to an answer and event URL from either a preexisting Nexmo Application or a newly created one:
+For this guide, you'll need to link it an answer and event URL to either a preexisting Nexmo Application or create a new one:
```
mariatechmaniac: For this guide, you'll need to link an answer and event URL to either a preexisting Nexmo Application or a newly created one

_documentation/stitch/in-app-voice/guides/ncco-guide.md
```diff
@@ -59,21 +67,20 @@ nexmo link:app 16625461410 aaaaaaaa-bbbb-cccc-dddd-0123456789ab
...
 ```
-?from_user=demo
+?from=demo
 &to=14155550100\
 &conversation_uuid=CON-4e977dab-2abc-42b5-bf64-d468d4763e54\
 &uuid=NONE
 ```
 
-The `to` parameter will contain the string or phone number you passed into the SDK's `callPhone` method. This parameter can be used to dynamically tell Nexmo which phone number to call. If you choose to pass in a string other than a phone number, such as a user ID or name, you can use that reference to look up that user's phone number instead. The `from_user` parameter is the name of the Stitch user that initiated the call. The `conversation_uuid` is the ID of the conversation that is created for the IP and PSTN user for this call. The `UUID` parameter does not apply to this guide.
+We can use the `to` parameter to dynamically tell Nexmo which phone number to call. The `from` parameter is the name of the Stitch user that initiated the call. The `conversation_uuid` is the ID of the conversation that is created for the IP and PSTN user for this call. The `UUID` parameter does not apply to this guide.
```
mariatechmaniac: from 62 to 79,
why dont we want to communicate that any string can be passed?
unless the call is a result of the user using a dialer to type the phone, the typical usecase will be 'i want to call user:maria.p', and then the customer backend fetches the phone number from their CRM/database) and puts it in the NCCO

mariatechmaniac: same with line 75-83

_documentation/stitch/in-app-voice/guides/ncco-guide.md
```diff
@@ -1,21 +1,29 @@
 ---
-title: Calling NCCO Guide
+title:  NCCO Guide for Calling
+navigation_weight: 5
 ---
 
-# Calling NCCO Guide 
+# Outbound PSTN Calling NCCO guide
```
mariatechmaniac: why not: NCCO Guide for Calling?
NCCOs are relevant for both IP PSTN & PSTN to IP 
so not just 'outbound' :)

AlexLakatos:
it got messed from the re-write branch. should be fine now.

_documentation/stitch/in-app-voice/guides/inbound-pstn.md
```diff
@@ -1,13 +1,11 @@
 ---
-title: Calling In-App users from a phone
+title: Phone to App Calls
+navigation_weight: 3
 ---
 
-# Calling In-App users from a phone
+# Phone to App Calls
 
-The [Voice API](/voice/voice-api/overview) allows you to build high-quality programmable voice applications in the cloud. With the Voice API, you can manage outbound and inbound calls in JSON, record and store calls, create a conference call, send text-to-speech messages in 23 languages with varieties of voices and accents, and so on.
```
ChrisGuzman: Why are we getting rid of this line?

AlexLakatos: @mariatechmaniac made this change in the branch I merged for #897

ChrisGuzman: ah ok. approved!

PR #913, opened 2018-08-23
Update from curl building blocks repo

lornajane:
## Description

Pulling in newest examples from curl repo. Haven't done this before, so can someone check it looks sane? It includes more than just the examples I have been working on ...


mheap:
That's how the tool currently works, but I agree that we should only be adding the `.repos/nexmo/curl-building-blocks/voice/voice-api` path in this commit. Can you unstage the un-needed files?

lornajane:
Ah OK, I thought we'd want all the changes.  I'll start again and include the building blocks additions along with the changes that use them.

PR #915, opened 2018-08-24
Curl additions for Voice Building Blocks

lornajane:
## Description

This PR brings in the new curl building blocks from their external repo, and adds the yml configuration to make them appear with the other blocks for Voice.

### DIFF
--- a/.repos/nexmo/curl-building-blocks/voice/voice-api/earmuff-a-call.sh
+++ b/.repos/nexmo/curl-building-blocks/voice/voice-api/earmuff-a-call.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+source "../../config.sh"
+source "../../jwt.sh"
+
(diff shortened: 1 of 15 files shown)


mheap:
These all render successfully. It looks as though there are two examples missing:

* http://localhost:3000/voice/voice-api/building-blocks/download-a-recording (GET request using a JWT)
* http://localhost:3000/voice/voice-api/building-blocks/transfer-a-call (PUT to a leg UUID with a new NCCO url)

lornajane:
Ah, I thought I was just "filling some gaps" - I took at stab at the things I understood well and thought I could test.  I'm sure there are others we could add, do you want to hold these examples until we have them all?

mheap:
As far as I can tell, those are the only two missing blocks. Once they're added that's 100% coverage for outbound requests in VAPI using `curl`

mheap:
Merging as the remaining blocks will take a little while to put together

PR #917, opened 2018-08-24
Add additional redirects based on 404s in Bugsnag

mheap:
## Description

Added redirects for some 404 links that I found in the Bugsnag 404 list

## Deploy Notes

N/A


config/redirects.yml
```diff
@@ -57,10 +59,15 @@
 "/voice/api-reference": "/voice/voice-api/api-reference"
 "/voice/sip": "/voice/sip/overview"
 "/voice/overview": "/voice/voice-api/overview"
+"/voice/voice-api/overview": "/voice/voice-api/overview"
```
lornajane: I don't understand what this one is doing!

mheap: Good catch! That's a copy/paste error

lornajane:
Looks good, all the links work, thanks for picking this up!

PR #918, opened 2018-08-24
Upgrade ruby version to 2.5.1

timcraft:
Ruby 2.5.1 is the current stable version and [includes some security fixes](https://www.ruby-lang.org/en/news/2018/03/28/ruby-2-5-1-released/).


mheap:
I can't seem to break anything in the review app 👍

PR #919, opened 2018-08-27
Block copy stats

mheap:
## Description

Adds a copy to clipboard button to building blocks. It works, but needs some design love.

screen shot 2018-08-27 at 15 09 31

Clicking copy will copy dependency configuration *and* the code itself as people seem to miss the dependency config section.

This only works on blocks using the `building_block` filter (currently Voice)

When the user clicks on the copy button for the first time per page load, we store which language they have selected, and which building block they're using.

## Deploy Notes

This adds a new DB migration to create the usage_copy_to_clipboard table. I believe they're run automatically but we should take care when deploying this

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,10 @@ gem 'coffee-rails', '~> 4.2'
 
 # Use jquery as the JavaScript library
 gem 'jquery-rails'
+# Add clipboard for copying content
+gem 'clipboard-rails'
+# Icons!
+gem 'octicons_helper'
 # Turbolinks makes navigating your web application faster. Read more: https://github.com/turbolinks/turbolinks
 gem 'turbolinks', '~> 5'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -95,6 +95,7 @@ GEM
     builder (3.2.3)
     byebug (9.0.6)
     chartkick (2.2.5)
+    clipboard-rails (1.7.1)
     coderay (1.1.2)
     coffee-rails (4.2.2)
       coffee-script (>= 2.2.0)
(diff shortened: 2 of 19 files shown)


mheap:
I've added more visual clues to the icon as I don't think it was obvious what the button did

screen shot 2018-08-27 at 15 18 02

Once you click:

screen shot 2018-08-27 at 15 18 08

db/migrate/20180827133110_usage_copy_to_clipboard.rb
```diff
@@ -0,0 +1,13 @@
+class UsageCopyToClipboard < ActiveRecord::Migration[5.1]
```
cr0wst: In `devrel-private` we talked about the possibility of making this table less granular so we can use it for other event tracking.

cr0wst:
Just a few discussion items from me. I like it, though. 👍

mheap:
@cr0wst I refactored to a more generic event store and now track links on dependency and main code source links in addition to copy to clipboard

lornajane:
I'm not sure about copying the dependencies. It sort of makes sense in PHP but makes no sense in Java where the two blocks don't belong next to one another and once they're in your clipboard, that's much less obvious. As a user, I would expect the "copy" link to only copy the block it was attached to.

mheap:
@lornajane I think it makes sense in Java too. `Initialize your dependencies` contains things like the following that are incidental, but required:

```
AuthMethod auth = new JWTAuthMethod(NEXMO_APPLICATION_ID, FileSystems.getDefault().getPath(NEXMO_PRIVATE_KEY));
NexmoClient nexmo = new NexmoClient(auth);
```

https://developer.nexmo.com/voice/voice-api/building-blocks/make-an-outbound-call/java

For me, I'd expect to have that copied

lornajane:
Right but these are two non-continuous blocks of code that might go in different locations or might need other stuff in between (my Java is rusty and my .NET non-existent, I'm judging this on what I see when I click the "view full source code" link).  We're offering to copy something that a) isn't what is in the box with the copy button attached and b) isn't a complete example.  I think it's less confusing to have the copy button copy what's in the box that the button is attached to since we can't really solve the problem of providing all the context that might be required to go with it.

mheap:
I don't feel strongly about copying both, so let's scope the copy to just the block selected and tackle the dependency configuration issue in another PR. I'll make the relevant changes

mheap:
@lornajane I've updated the branch with what we discussed. Could you review and make sure it does what you expect?

PR #929, opened 2018-09-03
Updated reports.yml

13webstreet:
## Description

Changed the request status field to request_status to avoid ambiguity with the status filter parameter.

## Deploy Notes

Can be deployed once reviewed.


ChrisGuzman:
Is `request_status` the parameter that's currently in production?

13webstreet:
Yes request_status is currently in production.

PR #930, opened 2018-09-04
WIP: Messages applications

panayiotispanteli:
## Description

## Deploy Notes


arisgoulia:
So this PR is blocked because a decision has not been taken where exactly the Applications concept should be in NDP. Shouldn't that be another concern rather than blocking an unrelated PR of messages product? @mheap.

PR #932, opened 2018-09-04
External accounts applications

panayiotispanteli:
## Description
Update external-accounts.yml to document applications

## Deploy Notes


mheap:
Rendered OAS doc at https://nexmo-developer-staging-pr-932.herokuapp.com/api/external-accounts

_open_api/definitions/external-accounts.yml
```diff
@@ -0,0 +1,694 @@
...
+  /messenger:
+    post:
+      security:
+        - bearerAuth: []
+        - basicAuth: []
+      summary: Create a Messenger account
+      operationId: CreateMessengerAccount
+      x-group: messenger
+      requestBody:
+        description: Parameters marked with * are optional
```
mheap: We have the `required` flag on parameters. Why do we need to mark optional with a `*`?

_open_api/definitions/external-accounts.yml
```diff
@@ -0,0 +1,694 @@
...
+                  type:
+                    type: string
+                    example: "https://www.nexmo.com/messages/Errors#InvalidParams"
+                  title:
+                    type: string
+                    example: "Your request parameters didn't validate."
+                  detail:
+                    type: string
+                    example: "Found errors validating 1 of your submitted parameters."
+                  instace:
```
mheap: Typo

_open_api/definitions/external-accounts.yml
```diff
@@ -0,0 +1,694 @@
...
+      summary: Link application to an account
+      operationId: LinkApplication
+      x-group: applications
+      parameters:
+        - in: path
+          name: provider
+          schema:
+            type: string
+          required: true
+          description: One of&#58;  messenger, viber_service_msg, whatsapp
```
mheap: I think we can use `enum` here

_open_api/definitions/external-accounts.yml
```diff
@@ -0,0 +1,694 @@
...
+          required: true
+          description: External id of the account you want to retrieve
+        - in: path
+          name: application_id
+          schema:
+            type: string
+          required: true
+          description: Id of the application you want to unlink
+      responses:
+        '200':
```
mheap: A delete would return a `204` not a `200` wouldn't it?

_open_api/definitions/external-accounts.yml
```diff
@@ -0,0 +1,694 @@
...
+      type: object
+      properties:
+        title:
+          type: string
+          example: "Wrong authentication - You are not authorised to access this resource"
+x-groups:
+  applications:
+    name: "Application"
+    order: 1
+    description: Foo Bar Baz
```
mheap: All of these descriptions need updating

_open_api/definitions/external-accounts.yml
panayiotispanteli: We don't I will change it

_open_api/definitions/external-accounts.yml
panayiotispanteli: It should but it doesn't for now. I will update the doc to 204 and create a ticket to make this change

mheap: You're deleting the link between the application and the account, not acting on the account resource. In this case I'd return a 204 with no body

mheap:
I've resolved the `applications` issue. I *think* it's all rendering correctly now, but I've looked at it too many times to be confident.

Review app at https://nexmo-developer-staging-pr-932.herokuapp.com/api/external-accounts

@arisgoulia Can you double check that the API is documented accurately?
@lornajane Could you take a look from a rendering point of view and highlight any issues you spot?

Long URL paths breaking out of the box are a known issue and will be resolved outside of this PR

mheap:
When this is merged can we `Squash and merge` please?

lornajane:
Do we need to add this into the navigation or home/api pages somewhere?

lornajane:
This all seems to render OK - I spied one broken link but it's for all OASes so I've fixed that separately.  The copy needs some work, we have some long and confusing sentences, and some abbreviations not defined. I would also like to see the descriptions for the response fields completed and some more example data throughout but nothing so major that I think we should block release.

_open_api/definitions/external-accounts.yml
```diff
@@ -0,0 +1,724 @@
...
+      type: object
+      properties:
+        title:
+          type: string
+          example: "Wrong authentication - You are not authorised to access this resource"
+x-groups:
+  applications:
+    name: "Application"
+    order: 1
+    description: Callbacks related an external-account which is linked to an application, will be delivered to the application's callback endpoints.
```
arisgoulia: Inbound messages to an external account which is linked to an application will  be delivered to the applications inbound url

PR #579, opened 2018-02-16
Add api-secret-management.yml

adambutler:
## Description

Adds API Secret Management API

## Deploy Notes

None

PR #935, opened 2018-09-05
Add redirects for moved number pages

lornajane:
## Description

We moved some pages (the number docs moved from under account to being their own level) but there were some URLs that were broken by this (#579 ) - this adds redirects to keep that content accessible.


lornajane:
This is actually more complicated than I thought. I've drawn a diagram and will start again with a new PR.

PR #938, opened 2018-09-06
Fixing links from the #579 secret mangement patch

lornajane:
## Description

In #579 we added the secret management API as a sublevel under the account area but we also moved the Numbers content from being the only thing under account to being its own top-level thing. Quite a few links were broken with the move, this PR aims to fix the broken ones and I'll open an issue to talk about some of the structure issues.


tbedford:
In the Management Account product box on the `/documentation` page the Overview, Guide, and Building Blocks links refer to Secret Management content. However, the API Reference link jumps to `/api/developer/account`. Should it in fact jump to the Secret Management API at `/api/account/secret-management`?

lornajane:
Great question and I'm really not sure. I aimed to fix the broken links only with this PR, and I'm capturing outstanding queries in https://github.com/Nexmo/nexmo-developer/issues/939 since I felt the broken links were urgent.

tbedford:
OK let's capture that as a question to be answered in #939. We can come back to that.

PR #955, opened 2018-09-14
Moving events above videos in Community page

leggetter:
## Description

I feel that the event listings are more important than the videos so should be above them.

We need to give the community page a rethink. But this is a small improvement for now.


mheap:
I've checked locally and this looks good. Should we also move the "Join community Slack" link above events to emphasise it? It's a little hidden at the moment

PR #959, opened 2018-09-18
Make concept filter look in concepts folder in addition to guides

mheap:
## Description

`concepts_filter` previously only searched the `guides` folder. This PR makes it search `concepts` too and merges the results together

## Deploy Notes

N/A

### DIFF
--- a/app/models/concept.rb
+++ b/app/models/concept.rb
@@ -63,7 +63,7 @@ class Concept
   end
 
   def self.files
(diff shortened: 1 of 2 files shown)


ChrisGuzman:
Looks like rubocop is complaining a bit. We don't have to worry about those pending tests do we? Don't see any other reason why the CI would fail.

PR #961, opened 2018-09-19
Call-Conversation-Leg Clarifications

mariatechmaniac:
Added some minor clarifications that will make this more understandable to your average user since just read about this for the first time.
We will like to this from the SDK Docs. It is good

## Description

## Deploy Notes

### DIFF
--- a/_documentation/voice/voice-api/guides/legs-conversations.md
+++ b/_documentation/voice/voice-api/guides/legs-conversations.md
@@ -12,9 +12,9 @@ The Nexmo Voice API has two concepts associated with a call - a leg and a conver
 
 A leg refers to a single connection either inbound to, or outbound from, the Nexmo platform.
 
-A conversation contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation is its own leg.
+A conversation has a 1:1 mapping with a call and it contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation has its own leg.
 
-When a new inbound or outbound phone call is received, a new conversation is created and the leg for the phone call is created. Once this has happened, an NCCO is retrieved from your `answer_url` and executed within the context of the conversation.
+When a new inbound or outbound phone call request is received, a new conversation is created and the respective legs for the phone call are created. Once this has happened, an NCCO is retrieved from your `answer_url` and executed within the context of the conversation.
 
 Additional legs may be added to the conversation using the `connect` action. If an NCCO contains multiple `connect` actions, then each new participant will be joined into the same conversation and all parties will be able to hear each other.
 
@@ -22,6 +22,6 @@ Legs and conversations are each identified by their own UUID. Conversation UUID'
 
 ## Conferences
 
-You can create a special type of conversation by using the `conversation` action within an NCCO. This resulting named conversation then acts like a conference bridge. New legs can then be added to this conversation by executing a `conversation` action with the same name.
+You can create a special type of conversation by using the `conversation` action within an NCCO. This resulting named conversation then acts like a conference bridge; a permanent conversation. New legs can then be added to this conversation via an NCCO by executing a `conversation` action with the same name.
 
 Conversation names are scoped to an application level. This means that any legs that are connected to a conversation with the same name will be able to hear each other, provided they are using the same Nexmo Application.
--- a/_documentation/voice/voice-api/guides/legs-conversations.md
+++ b/_documentation/voice/voice-api/guides/legs-conversations.md
@@ -12,9 +12,9 @@ The Nexmo Voice API has two concepts associated with a call - a leg and a conver
 
 A leg refers to a single connection either inbound to, or outbound from, the Nexmo platform.
 
-A conversation has a 1:1 mapping with a call and it contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation has its own leg.
+A conversation has a 1:1 mapping with a call and it contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation can have one or more legs.
 
-When a new inbound or outbound phone call request is received, a new conversation is created and the respective legs for the phone call are created. Once this has happened, an NCCO is retrieved from your `answer_url` and executed within the context of the conversation.
+When a new inbound or outbound phone call request is received, a new conversation is created and a leg is initiated linking the caller to the conversation. Once this has happened, an NCCO is retrieved from your `answer_url` and executed within the context of the conversation. In the case of a `connect` action, a leg is created and linked to the conversation as well; connecting the two ends.
 
 Additional legs may be added to the conversation using the `connect` action. If an NCCO contains multiple `connect` actions, then each new participant will be joined into the same conversation and all parties will be able to hear each other.
 
@@ -22,6 +22,6 @@ Legs and conversations are each identified by their own UUID. Conversation UUID'
 
 ## Conferences
 
-You can create a special type of conversation by using the `conversation` action within an NCCO. This resulting named conversation then acts like a conference bridge; a permanent conversation. New legs can then be added to this conversation via an NCCO by executing a `conversation` action with the same name.
+You can create a special type of conversation by using the `conversation` action within an NCCO. This resulting named conversation then acts like a conference bridge; a persisted conversation. New legs can then be added to this conversation via an NCCO by executing a `conversation` action with the same name.
 
 Conversation names are scoped to an application level. This means that any legs that are connected to a conversation with the same name will be able to hear each other, provided they are using the same Nexmo Application.


ChrisGuzman:
@mheap since this is a vapi thing would you like to take a look?

_documentation/voice/voice-api/guides/legs-conversations.md
```diff
@@ -12,16 +12,16 @@ The Nexmo Voice API has two concepts associated with a call - a leg and a conver
 
 A leg refers to a single connection either inbound to, or outbound from, the Nexmo platform.
 
-A conversation contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation is its own leg.
+A conversation has a 1:1 mapping with a call and it contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation has its own leg.
```
mheap: >  has a 1:1 mapping with a call and it

This makes me think that a conversation and a cal are different concepts and that there's a relationship between them. What are we trying to communicate? That `call` is another word for `conversation`?

> has its own leg

An endpoint connected to a conversation *is* a leg I thought? There's no concept of an endpoint to have a leg attached to it

_documentation/voice/voice-api/guides/legs-conversations.md
```diff
@@ -12,16 +12,16 @@ The Nexmo Voice API has two concepts associated with a call - a leg and a conver
...
 
 Additional legs may be added to the conversation using the `connect` action. If an NCCO contains multiple `connect` actions, then each new participant will be joined into the same conversation and all parties will be able to hear each other.
 
 Legs and conversations are each identified by their own UUID. Conversation UUID's are prefixed with `CON-`.
 
 ## Conferences
 
-You can create a special type of conversation by using the `conversation` action within an NCCO. This resulting named conversation then acts like a conference bridge. New legs can then be added to this conversation by executing a `conversation` action with the same name.
+You can create a special type of conversation by using the `conversation` action within an NCCO. This resulting named conversation then acts like a conference bridge; a permanent conversation. New legs can then be added to this conversation via an NCCO by executing a `conversation` action with the same name.
```
mheap: > ; a permanent conversation

I'm guessing that you're trying to communicate something similar to "Named conversations are persisted, and their conversation ID will be consistent." but I'm not sure on how to phrase it. Is this what you meant?

_documentation/voice/voice-api/guides/legs-conversations.md
```diff
@@ -12,16 +12,16 @@ The Nexmo Voice API has two concepts associated with a call - a leg and a conver
 
 A leg refers to a single connection either inbound to, or outbound from, the Nexmo platform.
 
-A conversation contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation is its own leg.
+A conversation has a 1:1 mapping with a call and it contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation has its own leg.
 
-When a new inbound or outbound phone call is received, a new conversation is created and the leg for the phone call is created. Once this has happened, an NCCO is retrieved from your `answer_url` and executed within the context of the conversation.
+When a new inbound or outbound phone call request is received, a new conversation is created and the respective legs for the phone call are created. Once this has happened, an NCCO is retrieved from your `answer_url` and executed within the context of the conversation.
```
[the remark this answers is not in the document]
mariatechmaniac: i see. ok let me fix with a middle soltion that will explain that first one joins then the ncco connects the other :)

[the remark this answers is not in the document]
mariatechmaniac: my other point is that: one cannot really 'receive' an outbound call no? that is why i used the word 'request'

_documentation/voice/voice-api/guides/legs-conversations.md
mariatechmaniac: a call is a conversation. there can be no call without a conversation. there can be a conversation without a call. that is what i was trying to convey. they are difference concepts but fundamentally call uses(is) a conversation.

_documentation/voice/voice-api/guides/legs-conversations.md
mariatechmaniac: yep exactly.
if you have any ideas on how to do it better; go for it!
ill change it to 'persisted'

_documentation/voice/voice-api/guides/legs-conversations.md
```diff
@@ -12,16 +12,16 @@ The Nexmo Voice API has two concepts associated with a call - a leg and a conver
...
 
-When a new inbound or outbound phone call is received, a new conversation is created and the leg for the phone call is created. Once this has happened, an NCCO is retrieved from your `answer_url` and executed within the context of the conversation.
+When a new inbound or outbound phone call request is received, a new conversation is created and a leg is initiated linking the caller to the conversation. Once this has happened, an NCCO is retrieved from your `answer_url` and executed within the context of the conversation. In the case of a `connect` action, a leg is created and linked to the conversation as well; connecting the two ends.
 
 Additional legs may be added to the conversation using the `connect` action. If an NCCO contains multiple `connect` actions, then each new participant will be joined into the same conversation and all parties will be able to hear each other.
 
 Legs and conversations are each identified by their own UUID. Conversation UUID's are prefixed with `CON-`.
 
 ## Conferences
```
mariatechmaniac: as i said in the call , i would change this title to 'named conversation` or somthing
but no need to do so now

_documentation/voice/voice-api/guides/legs-conversations.md
```diff
@@ -12,16 +12,16 @@ The Nexmo Voice API has two concepts associated with a call - a leg and a conver
 
 A leg refers to a single connection either inbound to, or outbound from, the Nexmo platform.
 
-A conversation contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation is its own leg.
+A conversation has a 1:1 mapping with a call and it contains one or more call legs. Each [endpoint](/voice/voice-api/guides/endpoints) connected to a conversation can have one or more legs.
```
mheap: How can an endpoint have more than one leg? I thought that each connection in to a call is it's own leg

_documentation/voice/voice-api/guides/legs-conversations.md
sammachin: I think the original statement is clearer and better than the proposed change.

_documentation/voice/voice-api/guides/legs-conversations.md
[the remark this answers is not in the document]
sammachin: `answered` or `connected` would be a better term that covers both inbound and outbound, I think technically for inbound the leg is created before the conversation,
The key point to get across here is that the NCCO is retrieved at the point the call is answered.

_documentation/voice/voice-api/guides/legs-conversations.md
sammachin: but if you jus have an NCCO with multiple `connect` actions that doesn't result in a named converstaion only a `conversation` action does that.

mariatechmaniac: agreed. 
but that is not described here. i dont think the point of this section is 'how to create a conference'(thats a separate tutarial id assume)
it is to introduce the concept of a long-lived conversation
(?)

_documentation/voice/voice-api/guides/legs-conversations.md
mariatechmaniac: Video - there are many legs in a single endpoint because there is one for each other member
and this is an example where the VAPI(voice) world clashes with the general conversational concepts again ;P

mheap: Perhaps we should close this PR against the VAPI docs and talk about it again when the CS concepts section is being developed. This guide could be migrated with specific explanations of how endpoints/legs work for messaging, audio and video as separate concepts?

mariatechmaniac: i understand they are correct for voice but my point is that docs will become more and more intertwined so we should work on a smooth transition between SDK, Conversation API & VAPI docs (especially since the Conversation API will be used by VAPI customers that want to do stuff in a lower level - aka Serenova; ask jonny)

tempporarily they can stay but i am just challenging us to start ironing out all the striking differences
hence move closer and closer to this single platform vision

mheap: I think we're saying the same thing. Between comms SDK, conversation, stitch and VAPI we have common concepts. We should have an explanation of how it works in detail in the conversation docs and reference those from the VAPI docs which only mention the parts that are relevant to audio.

With that in mind, I think our focus should be with spinning up a conversation concepts section rather than changing the VAPI docs to be technically accurate, but confusing for the majority of non-stitch consumers

lornajane:
I can't tell if we need to make changes to the existing docs at this stage, or if this PR was useful discussion that should not be merged. @mheap this looks like your department, we are moving forward with this?

mheap:
I think it was a valuable conversation, but it shouldn't be merged. We'll revisit as part of the conversation docs

PR #966, opened 2018-09-21
Quick fix to add Node example to Verify guide

leggetter:
This uses an old style Building Block but is a quick fix to ensure we have a Node example in the guide.

Note: this has been added via the GitHub UI so we should test on a staging app.


AlexLakatos:
LGTM 🚦 . I'll start a staging app

AlexLakatos:
Seems to be showing fine. https://nexmo-developer-staging-pr-966.herokuapp.com/verify/guides/verify-a-user

mheap:
We'll update this to a new style block when we revisit the Verify section. Merging

PR #967, opened 2018-09-21
Add node code for building blocks plus fix bad block for workflow guide.

tbedford:
## Description

- [x] **BUG FIX:** Add Node code to building blocks where we have them.
- [x] **BUG FIX:** Fix broken Workflow guide
- [x] Add Viber message with failover block (Node)
- [x] Add Viber message with failover block (Curl)
- [x] Add Basic auth blocks to repos.
- [x] Replace hard-coded code samples with tabbed_examples from repos. 
- [x] Add WhatsApp Node building block code as per Michael's request 

## Deploy Notes

Create review app and test all the pages.


mheap:
Should there be redirects added for `send-a-message` => `send-a-facebook-message`?

Within the `_examples` folder can we make the path match what's in the URL when viewing an example e.g. `_examples/olympus/messaging/send-sms/` rather than `_examples/olympus/send-sms/`

tbedford:
> Should there be redirects added for `send-a-message` => `send-a-facebook-message`?

send-a-message-with-failover should now redirect to send-a facebook-message-with-failover 

> Within the `_examples` folder can we make the path match what's in the URL when viewing an example e.g. `_examples/olympus/messaging/send-sms/` rather than `_examples/olympus/send-sms/`

Currently the layout matches that used by the curl building block repo. 

The Node quickstart repo uses subdirs of `channels` and `workflows`. 

We could create:

* `_examples/olympus/messages`
* `_examples/olympus/workflows` (the latter which will change shortly anyway)

If I create the sub-folders I'd have to reorganize the curl building block and the node building block repos to be consistent with that change, so that should probably be done first in a separate PR.

mheap:
Ok, we can do that as a separate PR.

Can we update the `tabbed_examples` sections to use `building_blocks` instead?

tbedford:
> Ok, we can do that as a separate PR.

Yep.

> Can we update the `tabbed_examples` sections to use `building_blocks` instead?

The reason I use tabbed examples rather than building blocks in the two Overviews only is simplicity. They are both Basic Auth examples - you don't need to create an application to use them as they don't need a JWT. If you do those examples as building blocks then users may then think they need to create an application as per the block prereqs - which is not required as they don't need a JWT.

mheap:
If you don't specify an `application` section in the YAML the application creation section won't be rendered in the building block. That should allow for our basic auth examples whilst still using `building_blocks`

tbedford:
> If you don't specify an `application` section in the YAML the application creation section won't be rendered in the building block. That should allow for our basic auth examples whilst still using `building_blocks`

My building block YAML looks like:

```
---
title: cURL
language: curl
code:
    source: .repos/nexmo/curl-building-blocks/olympus/send-facebook-message.sh
    from_line: 6
    to_line: 19
file_name: send-facebook-message.sh
run_command: bash send-facebook-message.sh
``` 
Which bit would I leave out to avoid the 'create an application' pre-req?

mheap:
Looking good to me, but we're missing a node building block for sending whatsapp. Once that's resolved I think we're good to merge

mheap:
All looks good to me. When we revisit the messaging/workflows docs it would be good to split the building blocks in to config/code, but in the interests of moving forward let's merge

Thanks for putting in the time to do this migration 🙌

PR #971, opened 2018-09-21
Update number insight overview page

marklewin:
## Description

Updated the Number Insight Overview page to reflect format used in audit and voice apis.

## Deploy Notes

Create test app and verify that `/number-insight/overview` renders correctly and reads well.

### DIFF
--- a/_documentation/number-insight/overview.md
+++ b/_documentation/number-insight/overview.md
@@ -2,60 +2,150 @@
 title: Overview
 ---
 
-# Number Insight Overview
+# Number Insight API Overview
 
-Nexmo's Number Insight API provides details about the validity, reachability and roaming status of a phone number, as well as give you details on how to format the number properly in your application.
+Nexmo's Number Insight API delivers real-time intelligence about the validity, reachability and roaming status of a phone number and tells you how to format the number correctly in your application.
 
-There are three levels of Number Insight API available: Basic, Standard and Advanced. The advanced API can be used asynchronously as well as synchronously.
+There are three levels of Number Insight API available: *Basic*, *Standard* and *Advanced*.
 
-Feature | Basic | Standard | Advanced
---|--|--|--
-Retrieve international and local format | ✅ | ✅ | ✅
-Country recognition | ✅ | ✅ | ✅
-Line type detection (mobile/landline/virtual number/premium/toll-free) | ❎ | ✅ | ✅
-Detect mobile country code (MCC) and mobile network code (MNC) | ❎ | ✅ | ✅
-Detect if number is ported | ❎ | ✅ | ✅
-Identify caller name (USA only) | ❎ | ✅ | ✅
-Check if phone number is reachable | ❎ | ❎ | ✅
-Identify network when roaming | ❎ | ❎ | ✅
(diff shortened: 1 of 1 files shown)


_documentation/number-insight/overview.md
```diff
@@ -2,60 +2,150 @@
...
 ---
 
-# Number Insight Overview
+# Number Insight API Overview
 
-Nexmo's Number Insight API provides details about the validity, reachability and roaming status of a phone number, as well as give you details on how to format the number properly in your application.
+Nexmo's Number Insight API delivers real-time intelligence about the validity, reachability and roaming status of a phone number and tells you how to format the number correctly in your application.
 
-There are three levels of Number Insight API available: Basic, Standard and Advanced. The advanced API can be used asynchronously as well as synchronously.
+There are three levels of Number Insight API available: *Basic*, *Standard* and *Advanced*.
```
cr0wst: I don't see the immediate value of having this sentence here. I would suggest maybe switching it to a bulleted list and provide a _very brief_ summary of each.

_documentation/number-insight/overview.md
```diff
@@ -2,60 +2,150 @@
...
+United Kingdom
+
+[country_prefix]
+44
+````
+
+
+
+
+## Guides
```
cr0wst: I know this isn't really NI related, but some thoughts I had:

I don't think it's clear what a Guide is compared to a Tutorial or what a Building Block is. I suggest that we add a global definition of each category and include it above the lists.

Something like:

```
## Building Blocks
Small snippets of code to help you get started with the following concepts:
```

This could be a global change and have the same text on each of the overview pages.

_documentation/number-insight/overview.md
lornajane: I think this is covered later on pretty well

_documentation/number-insight/overview.md
lornajane: This is outlined in the "in this guide" section at the top of the overview page. I do agree that the naming is hard but I think in this instance it's covered.

_documentation/number-insight/overview.md
cr0wst: I do too, but what value does this sentence add at this point? Why should I, as the reader, care that there are three levels of offerings if they aren't defined until further down?

_documentation/number-insight/overview.md
cr0wst: If I was able to miss that, how commonly do you think somebody _not_ familiar with Nexmo will?

lornajane:
I've approved but not merged in case we want to discuss more on @cr0wst 's suggestions first - @marklewin this is your call if you want it merged or you want to make any more changes.

_documentation/number-insight/overview.md
marklewin: I included them right up-front because an impatient (i.e. developer) reader, might see them listed in the building blocks in the left-hand navigation menu. I'll remove it for now.

_documentation/number-insight/overview.md
marklewin: I agree: it's not instantly obvious what the differences are between these items. This is something that we need to address about the platform as a whole though and is not immediately relevant to this PR.

PR #988, opened 2018-09-24
Add stitch-redirects file

mheap:
## Description

Adds a Stitch specific redirect file as we're going to be adding a *lot* more redirects with the rename/split

## Deploy Notes

N/A


ChrisGuzman:
Tested locally

PR #1003, opened 2018-09-27
Partials for use in new tutorials

tbedford:
## Description

Partial snippets that will be used for new layout Olympus tutorials. These will help to make the tutorials easier to maintain as a single maintained snippet can be reused in multiple tutorials.

## Deploy Notes

None. When the tutorials are rewritten to use the partials, the partials can be tested and modified at that time.


_partials/olympus/configure-webhook-urls.md
```diff
@@ -0,0 +1,38 @@
+## Configure your webhook URLs
+
+There are at least two webhooks you may need to configure:
+
+* Inbound message webhook
+* Delivery receipt webhook
+
+If you want to receive inbound messages you will need to configure an Inbound Message Webhook URL. When an inbound message is received this webhook URL will be invoked with the message payload.
+
+If you wish to get a status update on a sent message, such as received or read, then you will need to configure the delivery receipt webhook.
```
marklewin: Are "received" and "read" actual status messages? If so, might be nice to render them as `received` and `read`.

_partials/olympus/generate-a-jwt.md
```diff
@@ -0,0 +1,17 @@
+## Generate a JWT
+
+Once you have created a Nexmo Application you can use the Nexmo Application ID and   downloaded or locally created the private key file, `private.key`, to generate a JWT.
```
marklewin: "...Nexmo Application ID and    downloaded..." - think there's something missing/wrong here.

_partials/olympus/configure-webhook-urls.md
tbedford: Yes, that was me being lazy. Now fixed.

_partials/olympus/generate-a-jwt.md
tbedford: Err, copy and paste error? Now simplified.

PR #1012, opened 2018-09-28
Bibi dotnet building blocks

RabebOthmani:
PR contains Number Insight building blocks and Verify. 
@lornajane , @marklewin I created the PR to keep things moving but will continue adding more Building Blocks to this branch.


marklewin:
Bibi, the `CSharp.yml'` file should be renamed to `csharp.yml`. I'm not sure if the renderer requires it (it might!) but it keeps things consistent across all the building blocks. :)

tbedford:
Looking at the review app these blocks aren't loading correctly. I will try to find time to look into it. I believe that YAML files need a new line at EOF in order to load correctly.

lornajane:
@marklewin where did we get to with this? Do the files still need renaming? Have we hit a bad combination of old and new building blocks?

marklewin:
@lornajane: I'd actually already incorporated the Number Insight blocks that @RabebOthmani provided, but my changes were held up in an earlier PR. The other building blocks are for the Verify API and those for the other languages haven't been YAMLized yet and are still in `tabbed-content`.

I also note that the committed files are named `CSharp.yml` whereas convention (and perhaps the building blocks filter? I don't know) requires that they are all lower case (`csharp.yml`). Perhaps best to close the PR until we revisit the Verify docs?

lornajane:
OK, we're not ready to accept building blocks in the new style for verify - I'd love to have the voice ones though as I know @leggetter is keen to get a complete set for that. I'll close this for now.

PR #1018, opened 2018-10-03
Webhook documentation for Voice API answer and event URLs

lornajane:
## Description

This adds some documentation of the fields and possible values for each of the webhooks that we have for both answer and event on the Voice API.

This covers the main use cases but misses a few things including the `failed`, `rejected` and `error` events that I'm not sure how to trigger. If anyone can fill the gaps as well as review if the documentation is a useful format etc, then that would be grand. I think we need to look at where we link to it from and how it is rendered in the menu as well.

We probably need some discussion before it merges, but I'd appreciate reviews.

### DIFF
--- a/_documentation/voice/voice-api/webhook-reference.md
+++ b/_documentation/voice/voice-api/webhook-reference.md
@@ -64,6 +64,8 @@ The format of the data included depends on which event has occurred:
 * [`answered`](#answered)
 * [`busy`](#busy)
 * [`cancelled`](#cancelled)
+* [`unanswered`](#unanswered)
+* [`failed`](#failed)
 * [`human/machine`](#human-machine)
 * [`timeout`](#timeout)
 * [`completed`](#completed)
@@ -140,6 +142,34 @@ Field | Example | Description
 `uuid` | `aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this call
 `conversation_uuid` | `CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this conversation
 `status` | `cancelled` | Call status
+`direction` | `outbound` | Call direction, this will be `outbound` in this context
+`timestamp` | `2020-01-01T12:00:00.000Z` | Timestamp (ISO 8601 format)
+
+### Unanswered
+
+The outgoing call is ringing, but is never answered.
+
+Field | Example | Description
+ -- | -- | --
+`from` | `442079460000` | The number the call came from
+`to` | `447700900000` | The number the call was made to 
+`uuid` | `aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this call
+`conversation_uuid` | `CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this conversation
+`status` | `unanswered` | Call status
+`direction` | `outbound` | Call direction, this will be `outbound` in this context
+`timestamp` | `2020-01-01T12:00:00.000Z` | Timestamp (ISO 8601 format)
+
+### Failed
+
+The outgoing call could not be connected.
+
+Field | Example | Description
+ -- | -- | --
+`from` | `442079460000` | The number the call came from
+`to` | `447700900000` | The number the call was made to 
+`uuid` | `aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this call
+`conversation_uuid` | `CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this conversation
+`status` | `failed` | Call status
 `direction` | `outbound` | Call direction, this will be `outbound` in this context
 `timestamp` | `2020-01-01T12:00:00.000Z` | Timestamp (ISO 8601 format)
 
--- a/_documentation/voice/voice-api/webhook-reference.md
+++ b/_documentation/voice/voice-api/webhook-reference.md
@@ -66,6 +66,7 @@ The format of the data included depends on which event has occurred:
 * [`busy`](#busy)
 * [`cancelled`](#cancelled)
 * [`unanswered`](#unanswered)
+* [`rejected`](#rejected)
 * [`failed`](#failed)
 * [`human/machine`](#human-machine)
 * [`timeout`](#timeout)
@@ -75,7 +76,7 @@ The format of the data included depends on which event has occurred:
 
 ### Started
 
-Indicates that the call has begun.
+Indicates that the call has been created.
 
 Field | Example | Description
  -- | -- | --
@@ -126,7 +127,7 @@ Field | Example | Description
 
 ### Busy
 
-The destination number was busy.
+The destination is on the line with another caller.
 
 Field | Example | Description
  -- | -- | --
@@ -142,7 +143,7 @@ Field | Example | Description
 
 ### Cancelled
 
-An outgoing call was cancelled before it was answered.
+An outgoing call is cancelled by the originator before being answered.
 
 Field | Example | Description
  -- | -- | --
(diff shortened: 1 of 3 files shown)


sammachin:
My main thought is that its quite a lot of scrolling on the page to find each piece of information, I wonder it would be better to add the initial "this is what a webhook is" type stuff to the guides and then have a pure reference of each event in the same style as the API ref with example JSON on the right hand side. I think the OAS parser has support for webhooks? In fact I thought I added the call events to at least one version of the OAS spec for vapi

lornajane:
Perhaps we should combine the ones that are identical apart from the `status` field! I make that these:
* started
* ringing
* busy
* cancelled
* unanswered
* failed
* timeout

The human and machine statuses are very, very similar as well

sammachin:
The explanations for `cancelled` and `unanswered` are a little confusing to me that says the same thing? 
```
Cancelled
An outgoing call was cancelled before it was answered.
```
```
Unanswered
The outgoing call is ringing, but is never answered.
```
I suspect in some cases it may depend on which party cancelled/rejected the call?

lornajane:
I was hoping someone would just correct me on those! I'm really not at all sure what the difference is...

sammachin:
I believe the status are derived  from SIP responses, it might be worth asking engineering to confirm the mapping.
The other complicating factor is that those responses can vary by carrier, so when a user presses the Busy key on their phone to reject a call some carriers might return a Busy event and others might return a failed or rejected event.

_documentation/voice/voice-api/webhook-reference.md
```diff
@@ -0,0 +1,247 @@
...
+description: Details of the webhooks that Nexmo sends relating to voice calls.
+api: "Voice API: Webhooks"
+---
+
+# Voice API Webhooks Reference
+
+Nexmo uses webhooks alongside its Voice API to enable your application to interact with the call. There are two webhook endpoints:
+
+* [Answer webhook](#answer-webhook) is sent when a call is answered. This is for both incoming and outgoing calls.
+* [Event webhook](#event-webhook) is sent for all the events during a call. Your application can log, react to or ignore each event type.
```
marklewin: I would change "all the events during a call" to "all the events _that occur_ during a call". I'd find it difficult to defend that (so please feel free to ignore) but it just sounds clearer to me.

_documentation/voice/voice-api/webhook-reference.md
```diff
@@ -0,0 +1,247 @@
...
+
+For more general information, check out our [webhooks guide](/concepts/guides/webhooks).
+
+## Answer Webhook
+
+When an incoming call is answered, an HTTP request will be sent to the `answer_url` you specified when setting up the application. For outgoing calls, the `answer_url` is specified when making the call.
```
marklewin: "will be sent" -> "is sent"

"For outgoing calls, specify the `answer_url` when you make the call."

_documentation/voice/voice-api/webhook-reference.md
```diff
@@ -0,0 +1,247 @@
...
+
+By default, the answer webhook will be a `GET` request but this can be overridden to `POST` by setting the `answer_method` when making an outbound call with the Nexmo Voice API.
+
```
marklewin: I'm a bit unclear (to me at least) if this section (changing the `answer_method`to `POST`) applies only to outgoing calls? Is it possible to change the `answer_method` for an incoming call too?

_documentation/voice/voice-api/webhook-reference.md
```diff
@@ -0,0 +1,247 @@
...
+`to` | `447700900000` | The number the call was made to 
+`uuid` | `aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this call
+`conversation_uuid` | `CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this conversation
+`status` | `cancelled` | Call status
+`direction` | `outbound` | Call direction, this will be `outbound` in this context
+`timestamp` | `2020-01-01T12:00:00.000Z` | Timestamp (ISO 8601 format)
+
+### Unanswered
+
+The outgoing call is ringing, but is never answered.
```
marklewin: Do you mean that it rang, but was not answered? "Is ringing" implies that you can tell that it won't be answered in advance?

_documentation/voice/voice-api/webhook-reference.md
```diff
@@ -0,0 +1,247 @@
...
+`to` | `447700900000` | The number the call was made to 
+`conversation_uuid` | `CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this conversation
+`status` | `completed` | Call status
+`direction` | inbound | Call direction, can be either `inbound` or `outbound`
+`timestamp` | `2020-01-01T12:00:00.000Z` | Timestamp (ISO 8601 format)
+
+### Record
+
+This webhook arrives when an NCCO with a "record" action has finished. When creating a record action, you can set a different `eventURL` for this event to be sent to if you wish.
+
```
marklewin: So that you can keep recording events separate from other events?

_documentation/voice/voice-api/webhook-reference.md
```diff
@@ -0,0 +1,247 @@
...
+  "from": "442079460000",
+  "to": "447700900000",
+  "uuid": "aaaaaaaa-bbbb-cccc-dddd-0123456789ab",
+  "conversation_uuid": "CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab"
+}
+```
+
+### Responding to the answer webhook
+
+Nexmo expect you to return an array of [NCCO](/voice/voice-api/ncco-reference) in JSON.
```
mheap: The collection is an `NCCO`. Each entry is an `action`

_documentation/voice/voice-api/webhook-reference.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Answer webhook data field examples
+
+For a `GET` request, the variables will be in the URL, like this:
+
+```
+/answer.php?to=442079460000&from=447700900000&conversation_uuid=CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab&uuid=aaaaaaaa-bbbb-cccc-dddd-0123456789ab
+```
+
+If you set the `answer_method` to `POST` then you will receive the request with JSON-format data in the body:
```
mheap: I know that if we offer functionality we should document it, but should we be encouraging people to use `GET` for their `answer_url`?

_documentation/voice/voice-api/webhook-reference.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Responding to the answer webhook
+
+Nexmo expect you to return an array of [NCCO](/voice/voice-api/ncco-reference) in JSON.
+
+## Event webhook
+
+HTTP requests will arrive at the event webhook endpoint when there is any status change for a call. The URL will be the `event_url` you specified when creating your application unless you override it by setting a specific `event_url` when starting a call.
+
+The incoming requests are `POST` requests with a JSON body.
```
mheap: I believe this is the default, but we can make it a `GET`. If we document both options for `answer_url`, we should document both for `event_url`

_documentation/voice/voice-api/webhook-reference.md
lornajane: I'm trying to find out which thing means what. I haven't found anyone who is really clear about all these statuses yet and some of them can overlap ...

_documentation/voice/voice-api/webhook-reference.md
lornajane: It's the default ....

lornajane:
I've applied a bunch of the suggestions (thanks, everyone!), added more links back to the list of statuses to help make that long page more manageable, linked to these docs from more places and added an Error example with some help from Sam. I'm waiting to hear back from engineering but I notice we also have statuses listed in the Call Flow Guide https://nexmo-developer-stagin-pr-1018.herokuapp.com//voice/voice-api/guides/call-flow#call-states so I've updated from those for now. I think we're pretty close to merge, would anyone like to approve or request any further changes at this stage? I'm happy to keep on iterating!

mheap:
This is a huge improvement. I'm merging, and if anyone has additional requests we can apply them in a follow up PR

PR #1019, opened 2018-10-03
Update validate a number tutorial

marklewin:
## Description

The "Validate a number" Ruby tutorial was out of date and used an old version of the ruby client library. Updated the tutorial source code and the tutorial itself.

## Deploy Notes

Check tutorial against the new version of the sample app: git@github.com:Nexmo/ruby-ni-customer-number-validation.git


marklewin:
Great feedback, thanks. Have added a "Resources and Further Reading" section as suggested.

PR #1027, opened 2018-10-06
Add redirect tracking

mheap:
## Description

This increments a row in the Redirects table each time a redirect is used. Using
this data we can find out which redirects are no longer used and remove
them from our config file

## Deploy Notes

N/A


mheap:
@AlexLakatos Early christmas gift for you here

AlexLakatos:
Is this visible in the Admin dashboard?

mheap:
@AlexLakatos The redirect data is in the admin dashboard, yes. The part that isn't in the dashboard (but would be useful) is any redirects that exist that *don't* have an entry in the redirects table

AlexLakatos:
Now that sounds like something that should be put in an issue and worked on for actual Christmas? 😁

PR #1033, opened 2018-10-09
OTT-2071 Add field descriptions for Accounts API

ctickner:
## Description

Adding field descriptions to external accounts reference documentation


mheap:
Where it says things like `Viber Service Message account` in the description, could we have links to how to find out more?

PR #1042, opened 2018-10-10
Making external accounts public

arisgoulia:
## Description

## Deploy Notes


tbedford:
Should this really be a subsection of Messages API? It also applies to Dispatch API. It's almost like it should be another API related to, but not a subset of Messages API. @mheap

tbedford:
I see what you are trying to do now. We will try to find a better place to locate this API. We can then link to it from the Reference section of both Messages API and Dispatch API. Leave it with us!

tbedford:
See #1055

PR #1044, opened 2018-10-10
Add client lib to prereqs for tutorials and improve page.

tbedford:
## Description

- [x] Add installation of Node library to prereqs for tutorials
- [x] Convert client library to partial
- [x] Add info on upgrading library if you already have it installed

### DIFF
--- a/_partials/reusable/node-client-library.md
+++ b/_partials/reusable/node-client-library.md
@@ -1,0 +1,35 @@
+
+In addition to being able to use Messages and Dispatch via the REST APIs, Nexmo also provides client library support. 
+
+> **NOTE:** Currently client library support is only available for Node.
+
+## Installation
+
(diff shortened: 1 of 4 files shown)


_partials/reusable/node-client-library.md
```diff
@@ -0,0 +1,35 @@
+
+In addition to being able to use Messages and Dispatch via the REST APIs, Nexmo also provides client library support.
```
mheap: Should this mention messages + dispatch? If so should it be used for all beta client lib requirements as a partial and mention other available products in the beta?

Alternatively, should this partial be renamed to `beta-node-client-library-messages-dispatch.md`?

tbedford: Yes, I think renaming is better - it was meant to be M+D specific.

PR #1046, opened 2018-10-10
Attempt to describe signature generation.

judy2k:
This is not ready to merge!


lornajane:
@judy2k Can you add some context here about what you hoped this PR would achieve and what isn't ready about it, please? I'm not sure what I'm looking at.

judy2k:
@lornajane Well, I only posted this so I could send you  a link to the new content, which I thought might be useful 😉

lornajane:
I'm really confused :) I thought Mark opened the branch to discuss because there was more he wanted to do. But then I didn't find out what sort of more. I definitely need to revisit this topic but agree that this would be an improvement on what we have.

lornajane:
Are we signing requests here? Or messages? We seem to use the terminology interchangeably and I'm not sure which one is accurate.

mheap:
It does both. You can use the same process to sign an a request to the Nexmo platform, and to validate that the request is from the Nexmo platform.

I think they should be separate guides

mheap:
Adding more context here - the process for both hash and hmac seems to be the same until right at the end.

## For both hash and hmac:

* Add the current timestamp to the parameters list with the key timestamp. This should be an integer containing the number of seconds since the epoch (this is sometimes also known as UNIX time)
* Loop through each of the parameters, sorted by key
* For every value in the parameter list, replace all instances of & and = with an underscore _.
Generate a string consisting of `&akey=value&bkey=value`. *Note that there is an ampersand & at the start of the string!*

Once you've followed those:

## For hash

* Add an ampersand & and the signature secret to the end of the string. It should now look something like this: `&akey=value&bkey=value&your_signature_secret`
* Now run the string through an md5 hash function and convert the resulting bytes to a string of hexadecimal digits. This is your MD5 hash signature, and should be added to the HTTP parameters of your request as the sig parameter. (In php this looks like `md5($data)`)

## For hmac

* Create an HMAC generator with your desired algorithm and your signature secret as the key.
* Now run the string through an hmac generator and convert the resulting bytes to a string of hexadecimal digits. This is your HMAC signature, and should be added to the HTTP parameters of your request as the sig parameter. 

For PHP this looks like `hash_hmac($algorithm, $data, $secret);`

judy2k:
OK, so just coming in to see a bunch of comments I missed.

To clarify: I published this PR because @lornajane was attempting to either sign a request or validate a signature. The content was provided as-was simply because it's currently the best doc we have. It's marked as `don't-merge` because I'm not that happy with the wording as-is. I'm increasingly being persuaded we should just put this live though.

I think @mheap's context is quite useful, but in this case I think some duplication makes the process clearer. Once you've determined the kind of hash you're creating it's then just a series of linear instructions. Not dogmatic about this though.

I think it makes sense to have signing a request and validating a signature as separate guides, but we need to make clear that the process for generating the signature is _the same_, otherwise someone might implement this twice?

Also, OMG, our signature escaping is horrible. Why didn't we just use URL encoding?

judy2k:
Actually, I've just read @mheap's description again and I think I prefer it 🐑

lornajane:
I've pushed some commits and I think we're close so I've removed the "don't merge".

_documentation/concepts/guides/signing-messages.md
```diff
@@ -60,16 +39,55 @@ source: '_examples/messaging/signing-messages/create-request'
...
 
 ```tabbed_examples
 source: '_examples/messaging/signing-messages/validate-signature'
 ```
 
+## Manually generate a signature
+
+It is *highly recommended* that you use your Nexmo library's existing functionality for generating and validating signatures. If you aren't using a library with this functionality, you'll need to generate the signature yourself. The technique is slightly different if are generating an 'MD5 hash' signature or one of the HMAC signatures.
+
+Validating a Nexmo Signature requires generating a signature from an incoming request's parameters and comparing it to the signature that was provided in the request. **Note**: Remember to remove the `sig` parameter before generating your signature, and instead of adding a `timestamp`, use the `timestamp` provided in the request parameters.
```
mheap: I'd add the note here under step 1:

* **If you're generating a signature**: Add the current timestamp to the parameters list with the key `timestamp`. This should be an integer containing the number of seconds since the epoch (this is sometimes also known as UNIX time)
* **If you're validating a signature from Nexmo**:  Remove the `sig` parameter before generating your signature, and instead of adding a `timestamp`, use the `timestamp` provided in the request parameters.

_documentation/concepts/guides/signing-messages.md
```diff
@@ -60,16 +39,55 @@ source: '_examples/messaging/signing-messages/create-request'
...
+At this point, the process for hash and HMAC will differ, so use the "Step 2" section that fits your needs:
+
+### Step 2: for hash
+
+* Add an ampersand `&` and the signature secret to the end of the string. It should now look something like this: `&akey=value&bkey=value&your_signature_secret`
+* Now run the string through an md5 hash function and convert the resulting bytes to a string of hexadecimal digits. This is your MD5 hash signature, and should be added to the HTTP parameters of your request as the `sig` parameter.
+
+### Step 2: for HMAC
+
+For hmac
```
mheap: Looks like a copy/paste error

_documentation/concepts/guides/signing-messages.md
```diff
@@ -60,16 +39,57 @@ source: '_examples/messaging/signing-messages/create-request'
...
 source: '_examples/messaging/signing-messages/validate-signature'
 ```
 
+## Manually generate a signature
+
+It is *highly recommended* that you use your Nexmo library's existing functionality for generating and validating signatures. If you aren't using a library with this functionality, you'll need to generate the signature yourself. The technique is slightly different if are generating an 'MD5 hash' signature or one of the HMAC signatures.
+
+Validating a Nexmo Signature requires generating a signature from an incoming request's parameters and comparing it to the signature that was provided in the request. **Note**: Remember to remove the `sig` parameter before generating your signature, and instead of adding a `timestamp`, use the `timestamp` provided in the request parameters.
+
+### Step 1: for both hash and HMAC signatures
```
mheap: Headings should be title case

_documentation/concepts/guides/signing-messages.md
```diff
@@ -60,16 +39,57 @@ source: '_examples/messaging/signing-messages/create-request'
...
 
 ```tabbed_examples
 source: '_examples/messaging/signing-messages/validate-signature'
 ```
 
+## Manually generate a signature
+
+It is *highly recommended* that you use your Nexmo library's existing functionality for generating and validating signatures. If you aren't using a library with this functionality, you'll need to generate the signature yourself. The technique is slightly different if are generating an 'MD5 hash' signature or one of the HMAC signatures.
+
+Validating a Nexmo Signature requires generating a signature from an incoming request's parameters and comparing it to the signature that was provided in the request. **Note**: Remember to remove the `sig` parameter before generating your signature, and instead of adding a `timestamp`, use the `timestamp` provided in the request parameters.
```
mheap: Do we still need this note if we have the notice about validating a signature below?

_documentation/concepts/guides/signing-messages.md
lornajane: Title case for top-level headings only, says https://developer.nexmo.com/contribute/guides/writing-style-guide#headings but happy to break the "rules" if you think it would make things clearer here

lornajane:
I had somehow not really looked at the code samples and they are essentially nonsense. They should use the functionality in the libraries. I'm not sure if it might be better to remove the code completely and publish, and ask each language initiative owner to add code samples as appropriate?

_documentation/concepts/guides/signing-messages.md
lornajane: Oops, too much surgery and those two sections ended up right next to one another. I've removed it since it is in the next section almost verbatim.

_documentation/concepts/guides/signing-messages.md
mheap: Ah, I'd missed that it was for top level only

I think what's throwing me is that the `for` is lower case after a colon, which is technically correct but looks wrong to me. @tbedford - any thoughts?

_documentation/concepts/guides/signing-messages.md
```diff
@@ -1,52 +1,31 @@
...
+## Use signatures when sending messages
 
-* The parameters - all the parameters in a request sorted in alphabetic order
-* A timestamp - a UNIX timestamp at UTC + 0 to protect against replay attacks
-* Your `SIGNATURE_SECRET` - the key supplied by Nexmo that you use to sign or validate requests
+To send a message with a signature, you will need to use the `SIGNATURE_SECRET` instead of your `API_SECRET` when sending the message. You can find the signature secret and choose which signing algorithm to use by visiting the [dashboard](https://dashboard.nexmo.com). The default algorithm is'[MD5 hash](https://en.wikipedia.org/wiki/MD5)' and we also support `MD5 HMAC`, `SHA1 HMAC`, `SHA-256 HMAC` and `SHA-512 HMAC`.
 
-The signature has a leading `&`. All parameters in the hash input, apart from your `SIGNATURE_SECRET` are separated by `&`.
+We always recommend you use your Nexmo library's ability to generate or validate signatures. If you *can't* do this for some reason, you *can* generate and validate signatures yourself, but this can be complicated and potentially error-prone. Refer to the [section on manually generating signatures](#manually-generate-a-signature).
```
tbedford: Nexmo recommends you use the Nexmo library's ability...

_documentation/concepts/guides/signing-messages.md
```diff
@@ -1,52 +1,31 @@
...
-The following example shows a signed request to the SMS API:
-
-```
-https://rest.nexmo.com/sms/xml?api_key=API_KEY&from=Nexmo&to=447700900000&type=text&text=Hello+from+Nexmo&status-report-req=false&timestamp=1461605396&sig=SIGNATURE
-```
-
-The workflow for using signed messages is:
-
-![Signing requests workflow](/assets/images/workflow_call_api_outbound.svg)
+The overall process for sending a signed message looks like this:
```
tbedford: The process for sending a signed message is as follows:

_documentation/concepts/guides/signing-messages.md
tbedford: Yes title case for top-level only. Blog posts have a different convention - they use title-case for all headings for historical reasons. 

On this specific case. What you could do is have **Step1. For generating both...** That gets around the slight contention.

judy2k:
Thanks everybody for getting involved with this -- the collaboration is really helping improve it :heart:

_documentation/concepts/guides/signing-messages.md
lornajane: Thanks Tony, I went with that

PR #1049, opened 2018-10-10
Delegate Class or File Verbiage to Language Renderers

cr0wst:
## Description

Update the verbiage under "initialize your dependencies" and "write the code" to pull verbiage from the language renderers. This allows for languages like Java to refer to it as class creation and languages like PHP or JS to refer to it as file creation.


app/filters/building_block_filter.rb
```diff
@@ -31,10 +31,12 @@ def call(input)
       if highlighted_client_source
         client_url = generate_source_url(config['client'])
         id = SecureRandom.hex
+        create_verbiage = @renderer.create_verbiage(config['file_name'])
```
mheap: It may just be me, but `verbiage` makes me stop and think every time I read it. Is there another phrase we could use?

app/services/building_block_renderer/curl.rb
```diff
@@ -12,5 +12,13 @@ def self.run_command(command, _filename, _file_path)
 
       HEREDOC
     end
+
+    def self.create_verbiage(filename)
+      "<p>Create a file named <code>#{filename}</code> and add the following code:</p>"
```
mheap: I'm not too sure on HTML in these renderers. The `run_command` method uses markdown. Could we do the same in these methods?

app/views/building_blocks/_configure_client.html.erb
```diff
@@ -5,7 +5,7 @@
 </h3>
 
 <div id="<%= id %>" class="collapsible-content" style="display: none;">
-    <p>Create a file named <code><%= config['file_name'] %></code> and add the following code:</p>
+    <%= create_verbiage %>
```
mheap: If we switch to markdown, this will become `create_verbiage.render_markdown`

app/filters/building_block_filter.rb
cr0wst: I struggled with that too. I'd appreciate any suggestions.

PR #1061, opened 2018-10-11
Number insight oas

marklewin:
Initial PR for @mheap to look at issue where response field descriptions are repeating within each level of the API.

For example, open the response fields for Basic, and you'll see that the fields start repeating after `country_prefix`.


marklewin:
@mheap: I have done what I can to make this complete as possible, but it's still struggling to render the XML. Also, the callback response seems willing to render the nested objects (without descriptions) but the other end points do not so I have provided descriptions for those. I think it's an improvement on what we have right now (in that the three endpoints are clearly defined) but way off perfect and I'm not sure whether you want to merge or not at this point?

marklewin:
In particular, the responses don't seem to be displaying just JSON/XML when you select that option from the dropdown on the right hand side. I'm unsure at the moment whether these are NDP problems or the way the spec has been crafted. (And I suspect a mixture of the two). So no good for merging, but would appreciate guidance!

mheap:
@marklewin The spec was good, it was NDP's rendering that was broken. I've pushed up a fix for the following issues:

* Response fields now shows only JSON or XML depending on dropdown
* XML response example is now rendered correctly

Take a look and let me know if there are any other issues you'd like me to look at

marklewin:
@mheap This is a great improvement, thanks!

A few other things I noticed:
* The nested objects are shown for the Advanced API (Async callback), but not the others
* When the nested objects are shown, it would be nice if we could show the top level description for that object too
* The enumerated types are not being shown in the responses for the Basic and Standard API levels, but they are in the Advanced API? (Async callback)

If any/all of these are to do with the spec, I can't see it 😕 ?

mheap:
> The nested objects are shown for the Advanced API (Async callback), but not the others

This is an issue with the rendering. It doesn't use recursion, so it shows the top level only. It's something to address, but not in this PR

> When the nested objects are shown, it would be nice if we could show the top level description for that object too

There's a `x-nexmo-developer-collection-description-shown: true` flag that you can add to an object to have it render. I've added it to all of the properties I could see. This isn't ideal, and we'll look to remove it in the future

> The enumerated types are not being shown in the responses for the Basic and Standard API levels, but they are in the Advanced API? (Async callback)

Can you provide specific field names / screenshots so that I know what to look for?

marklewin:
> > The enumerated types are not being shown in the responses for the Basic and Standard API levels, but they are in the Advanced API? (Async callback)
> 
> Can you provide specific field names / screenshots so that I know what to look for?

@mheap This is looking great. Thanks for all your efforts :)

If you look at the `status` field across the three API levels, you'll see that only the Advanced API renders the `enum` info from the spec:
screen shot 2018-10-19 at 11 20 49

PR #1071, opened 2018-10-13
Update oauth.md

SoftCreatR:
The link to oauth.net leads to a list of OAuth 2.0 libraries, but Nexmo only supports OAuth 1.0a. The list of OAuth 1 libraries has been moved to `https://oauth.net/1/`.


lornajane:
Good spot, thanks very much :)

PR #1081, opened 2018-10-15
Update documentation

spencerng:
## Description

I fixed the grammar in several documentation files and changed the passive voice to the active voice, which is easier for users to understand. I also added links to the client libraries mentioned in the article.

## Deploy Notes

N/A


spencerng:
Would it be possible to complete a more comprehensive code review so that I can make the necessary changes? Thanks!

PR #1075, opened 2018-10-14
Updated README.md & Added nexmo_dev.PNG

Cooops:
## Description

**Updated README.md to:**

1. Include the Nexmo Developer brand image.
2. Add anchor tags to each section.
3. Add <a> href's for each anchor.
4. Added urls for Ruby, Bundler, and PostgreSQL.
5. Merged redundant Contributing section into top section Contributions.

This is one of my very first pull requests as part of Hacktoberfest, so I hope I did everything correctly. Please let me know if I did not. I hope you find the (minor) changes suitable.

Thanks for checking out my PR 😄

PR #1089, opened 2018-10-16
Remove Nexmo_dev.PNG

vipulgupta2048:
This PR might be wrong, but still wanted to open one. 
#1075 Added a png in the repository. Later in the commit message, the same image was written to be removed. But was never did. Hence removing it now, since it's of no use. (I think)


vipulgupta2048:
@mheap @lornajane  Please review

vipulgupta2048:
Thank you @mheap  :hatching_chick:

PR #1095, opened 2018-10-17
Render all documentation pages at build time in case exceptions are thrown

mheap:
## Description

In a recent commit a change was made that added a new `type` field to
building blocks without a default. The decision to fail fast was made as
it looked like all building blocks had been updated.

It wasn't until we tried to update the search index (which renders every
page) that we noticed an exception being thrown by a building block that
hadn't been updated.

This commit adds a build step that renders every markdown file at build
time to prevent this happening again

## Deploy Notes

N/A

### DIFF
--- a/lib/tasks/ci.rake
+++ b/lib/tasks/ci.rake
@@ -1,6 +1,6 @@
 namespace :ci do
-  desc 'Render all pages to make sure that no exceptions are thrown'
-  task 'render_pages': :environment do
+  desc 'Verify all pages to make sure that no exceptions are thrown'
+  task 'verify_pages': :environment do
     document_paths =
       [
         "#{Rails.root}/_documentation/**/*.md",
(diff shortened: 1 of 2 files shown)


lornajane:
I have questions. The build isn't particularly quick, how would this addition change the build time? How is the time spent in the various build stages? I mean, with a more comprehensive set of build steps hopefully we won't be pushing in a rush to fix things anyway, but I'm still curious.

mheap:
This takes 5-10 seconds on my local machine. Most of the build time at the moment is cloning the project / installing dependencies

lib/tasks/ci.rake
```diff
@@ -0,0 +1,24 @@
+namespace :ci do
+  desc 'Render all pages to make sure that no exceptions are thrown'
+  task 'render_pages': :environment do
```
cr0wst: How would you feel about naming this task something like `verify_pages` instead of `render_pages`? I think having the task name reflect the goal of the task makes it a bit more clear.

Non-blocking, open for discussion.

mheap:
I made the naming changes - thanks @cr0wst

PR #1105, opened 2018-10-18
Add support for oneOf for a single parameter

mheap:
## Description

This is required to render the Voice `from` column accurately. Previously it was rendered as an array of `oneOf` phone, sip or websocket. This PR updates it to the latest `oas_parser` which has support for rendering `oneOf` directly without an array

## Deploy Notes

N/A


mheap:
I'm going to be making changes to the rendering and will get a ton of conflicts if this isn't merged first. I've tested locally and it all seems to be fine

PR #1112, opened 2018-10-18
Change CTA for how to get in touch

leggetter:
Adding "contact support" with link to help.nexmo.com


leggetter:
Need to be checked by @PhilHolcombe and there may be other places we should update similar text.

lornajane:
Pushed the preview app again and sent the link to @PhilHolcombe on slack to get his input

lornajane:
Looks like the podcast and migration landing pages will also need this update once we're happy with it. There's a bunch of other places where we recommend emailing support but they are all "email support with X in the subject line" rather than just emailing for all sorts of help.

PR #1117, opened 2018-10-18
Added client_ref and locale for whatsapp

arisgoulia:
## Description

## Deploy Notes

### DIFF
--- a/_open_api/definitions/messages-olympus.yml
+++ b/_open_api/definitions/messages-olympus.yml
@@ -98,7 +98,7 @@ components:
           $ref: '#/components/schemas/MessageProperty'
         client_ref:
           type: string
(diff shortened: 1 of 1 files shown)


_open_api/definitions/messages-olympus.yml
```diff
@@ -182,6 +186,10 @@ components:
               type: string
               example: '0.0333'
               description: The charge amount as a stringified number.
+        client_ref:
+          type: string
+          description: The client's reference.
```
peterjurkovic: Maybe we should mention that it will appear in all callbacks

arisgoulia: Fair point.

PR #1118, opened 2018-10-18
Reinstate the Messages and Dispatch APIs on the APIs page

### LINKED ISSUE #1113: Dispatch and Messages API and External accounts missing from Api reference
## Description

Dispatch and Messages API and External accounts missing from Api reference
https://developer.nexmo.com/api

lornajane:
## Description

Fixes #1113 - we had lost the new APIs that were added to `/api` on NDP. I also removed the extra buttons that arrived with the redesign because they didn't make much sense, they linked to one of the APIs but the whole point was that there were lists of APIs to choose from. These two things are in different commits in case anyone wants to keep the buttons ...

### DIFF
--- a/app/views/api/index.html.erb
+++ b/app/views/api/index.html.erb
@@ -32,11 +32,6 @@
             <p>For verifying users with US Short Codes</p>
           </a>
         </nav>
-        <div class="Vlt-card__cta">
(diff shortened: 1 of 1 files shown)


mheap:
I checked locally to see what it looks like with the buttons and agree that it's better without

PR #1122, opened 2018-10-18
Unify descriptions in table at get-events-with-filtering page

StrikerRUS:
## Description

Unify descriptions in table at ``audit/building-blocks/get-events-with-filtering`` page.

## Deploy Notes


mheap:
Thanks for the PR! 🙌

PR #1127, opened 2018-10-19
Usage for the consistency while referring to API

### LINKED ISSUE #1109: Consistency in using 'the' when referring to APIs
## Description

Need to check docs for consistency in use of 'the' when referring to APIs. 

For example, this page:

https://developer.nexmo.com/tutorials/sending-whatsapp-messages-with-messages-api

Does not use 'the' in the title.

arshadkazmi42:
Fixes #1109 
----------

## Description

Consistency in using `the` when referring to API

## Deploy Notes

N/A


_open_api/definitions/external-accounts.yml
```diff
@@ -769,7 +769,7 @@ x-groups:
   account:
     name: "Account"
     order: 2
-    description: An external-account used as the `from` field in Messages API and Dispatch API
+    description: An external-account used as the `from` field in  Messages API and Dispatch API
```
mheap: Should `the` be here?

arshadkazmi42: I added it before, then removed, as after adding `the` it becomes `the Messages API and the Dispatch API`

arshadkazmi42: Should I add it back?

mheap: How about `in the Messages API and Dispatch API`?

arshadkazmi42: This sounds better. Let me make that change right away

PR #1133, opened 2018-10-20
Add SVGs for JavaScript, Android, iOS and modified tabs to display them

Axle7XStriker:
## Description
SVGs for JavaScript, Android and iOS were not present even though they were being used to represent tabbed content in stiches API especially for JavaScript which is also a language as well. So to make them consistent with other tabbed content, I added SVGs for them and also modified tabs in the `tab_filter.rb` file to display SVGs for platform as well.

screenshot from 2018-10-20 17-24-42


mheap:
Thanks for this @Axle7XStriker

PR #1138, opened 2018-10-21
Add support for subfolders in building_block_list

### LINKED ISSUE #1108: Building block list does not handle subfolders
## Description

If in the `building-blocks` directory you have subdirectories that contain blocks, these are not handled by the `building_block_list` macro. The `building_block_list` macro is used to display an autogenerated list of building blocks in the Overview.

We should be able to handle a directory structure such as:

<ul>
<li>building-blocks</li>
       <ul>
          <li>conversation</li>
          <ul>
                 <li>block-1.md</li>
                 <li>block-2.md</li>
          </ul>
       </ul>
</li>

mheap:
## Description

This adds support for subfolders when using `building_block_list`. Fixes #1108

screen shot 2018-10-21 at 12 35 15

Any top level building blocks will now be rendered before subfolders. There's currently no way to choose which order subfolders appear in.

I'm not sold on how they're rendered, but this is a step in the right direction for the Conversation API docs

## Deploy Notes


tbedford:
I currently get a non-zero postdeploy error on Heroku so I am not able to confirm this works OK yet.

mheap:
@tbedford I've triggered another deploy. There are no nested building blocks in this PR, so the screenshot is all you'll have to go on other than making sure that nothing is broken.

To test more thoroughly, create some subfolders in a `building-blocks` folder on your local NDP and test that way

PR #1146, opened 2018-10-22
Add allOf support for OAS documents

mheap:
## Description

This should probably be in OasParser itself and we should update our
rendering to check explicitly for allOf. As the semantics are the same
no matter if we loop over each allOf, or merge all of the schemas this
is the quickest way to keep us moving with valid OAS documents

## Deploy Notes

Check API reference pages carefully please!

### DIFF
--- a/app/views/open_api/_parameter_groups.html.erb
+++ b/app/views/open_api/_parameter_groups.html.erb
@@ -15,13 +15,28 @@
 <% if endpoint.request_body %>
   <% format = endpoint.request_body.formats[0] %>
   <% unless endpoint.request_body.exhibits_one_of_multiple_schemas?(format) %>
-    <% if endpoint.request_body.properties_for_format(format).any? %>
+
+    <%
+      # @mheap: This is hacky handling for allOf, but it works for now
+      # We probably want real allOf support in OasParser, but that's more time than I have right now
+      allOf = endpoint.request_body.content[format]['schema']['allOf']
+      if allOf
+        params = allOf.reduce { |a,b| a.deep_merge(b) }
+        params = params['properties'].map do |name, definition|
+          OasParser::Property.new(nil, params, name, definition)
+        end
+      else
+        params = endpoint.request_body.properties_for_format(format)
+      end
+      if params
+    %>
+
       <h4>
         Request body
         <span class="Vlt-badge Vlt-badge--blue"><%= format %></span>
       </h4>
 
-      <%= render 'parameters', parameters: endpoint.request_body.properties_for_format(format), callback: callback %>
+      <%= render 'parameters', parameters: params, callback: callback %>
     <% end %>
   <% else %>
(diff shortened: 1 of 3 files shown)


app/views/open_api/_parameter_groups.html.erb
```diff
@@ -15,13 +15,28 @@
 <% if endpoint.request_body %>
   <% format = endpoint.request_body.formats[0] %>
   <% unless endpoint.request_body.exhibits_one_of_multiple_schemas?(format) %>
-    <% if endpoint.request_body.properties_for_format(format).any? %>
+
+    <%
+      # @mheap: This is hacky handling for allOf, but it works for now
+      # We probably want real allOf support in OasParser, but that's more time than I have right now
+      allOf = endpoint.request_body.content[format]['schema']['allOf']
+      if allOf
```
cr0wst: I'm not 100% sure what this is doing. However, I can't seem to find a page where this statement is true. I tried with secret management as I noticed it had `allOf` in its `schema` but this block was never reached.

cr0wst: Ah it looks like it's because that's not part of the `request_body.content` in the secret management definition.

mheap: We don't have any yet. Here's a minimal example that renders in NDP (it'll complain about missing responses in Swagger editor)

```
---
openapi: "3.0.0"
info:
  version: "v2"
  title: "Number Pools"
servers:
  - url: https://api.nexmo.com
paths:
  /v2/pools:
    post:
      operationId: createPool
      summary: "Create a new pool"
      description: ""
      requestBody:
        description: Update Pool Description
        required: true
        content:
          application/json:
            schema:
              allOf:
                - $ref: "#/components/schemas/Pool"
                - required:
                    - pool_id
      responses: {}

  /v2/pools/{pool_id}:
    patch:
      operationId: patchPool
      summary: "Update a pool"
      description: ""
      requestBody:
        description: Update Pool Description
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Pool"
      responses: {}

components:
  schemas:
    Pool:
      properties:
        pool_id:
          type: string
          description: An identifier for this pool to be created — it may be alpha-numeric but MUST be unique to the account
          example: pool_id
```

mheap:
@marklewin Thanks for the catch! I'd added support for request parameters, but not response descriptions. The PR has now been updated

cr0wst:
I'm going to defer to @marklewin with how this is supposed to work. I imported the `number-insight.yml` file he linked and the ~Response Fields` still seem a bit odd.

There's the initial Response Fields:
image

Then it looks like the same fields repeat further down but in a different order?
image

marklewin:
Yes, agree with @cr0wst that this isn't working as I'd expect. Steve is seeing the same behaviour as I noted in an earlier comment.

marklewin:
When I check this in the review app everything looks fine. So, like @cr0wst, I must have something weird going on locally. Merging.

PR #1170, opened 2018-10-27
add a note about port numbers

### LINKED ISSUE #1145: Add note on port numbers
## Description

Svetlana Beliakova [5 days ago]
Hey team, I have noticed that in JS Stitch Quickstart guide we have this `Load localhost:8080 in two side-by-side browser windows.`. But I would suggest adding a tiny note there that port can be different to be more consistent, what do you think? We do not that path can be different above though. :slightly_smiling_face: Thank you!

Svetlana Beliakova [5 days ago]
https://developer.nexmo.com/stitch/in-app-messaging/guides/simple-conversation/javascript

Follow up to clarify the issue and apply the fix.

dagolinuxoid:
Fix #1145 
I think it is ok to add some clarification about the fact that 8080 isn't only magical combination of digits used here as a port number, but rather just an example.

## Description

Take a look at the text above the description :)

## Deploy Notes

### DIFF
--- a/_partials/stitch/guides/simple-conversation/javascript.md
+++ b/_partials/stitch/guides/simple-conversation/javascript.md
@@ -374,7 +374,7 @@ Your `index.html` file should now look something like [this](https://github.com/
 To test your application:
 
 1. Run a local webserver using `static`, `http-server` or any that is convenient.
-2. Load `localhost:8080` in two side-by-side browser windows. This will automatically load `index.html`.
+2. Load `localhost:8080` (where `8080` is a port number and it can be different like `3000` etc) in two side-by-side browser windows. This will automatically load `index.html`.
 3. Type any user name and click the Login button.
 4. Type a message and send it. You can create a two-way conversation by switching between the browser tabs or windows.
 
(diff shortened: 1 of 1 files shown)


_partials/stitch/guides/simple-conversation/javascript.md
```diff
@@ -373,8 +373,8 @@ Your `index.html` file should now look something like [this](https://github.com/
 
 To test your application:
 
-1. Run a local webserver using `static`, `http-server` or any that is convenient.
-2. Load `localhost:8080` in two side-by-side browser windows. This will automatically load `index.html`.
+1. Run a local webserver using `static`, `http-server` on any free port. The default port is `8080`.
+2. Prompt `localhost:port` in two side-by-side browser windows. This will automatically load `index.html`. Note: `port` is the port you're running your local web server on.
```
tbedford: Not sure why you use "prompt". Normally we would say "Load a web browser". Perhaps the following is better:

2. Load two browser windows side-by-side using address `localhost`:_port_. This will automatically load `index.html`. Note: _port_ is the port you're running your local web server on.

Note, I deliberately put _port_ in italics as it is a replaceable value.

dagolinuxoid: Yes, it makes sense.

PR #1192, opened 2018-10-31
Simplification of tutorials plus external account linking.

tbedford:
## Description

Various improvements:

- [x] Simplified tutorials.
- [x] For Facebook put linking Facebook to Nexmo first (so app can link to external account when created).
- [x] Create application now contains info on creating webhooks and a step in linking the application to displayed external accounts (linked Facebook pages). 
- [x] Added creating JWT to Before you Begin via partial.
- [x] Removed JWT creation as a main step as you only need it if using Curl. Added note below the code to the effect if you are using Curl go "here" to create JWT. 
- [x] Tried to make it more obvious that for WhatsApp you need a verified account.
- [x] Added note on linking your Facebook Page to your Nexmo account.
- [x] Added note on 200 response requirement for callbacks and callback queue management.


tbedford:
@hughhopkins: Not tested yet - just to make you aware of this PR.

PR #1193, opened 2018-11-01
Add stats by product summary page

mheap:
## Description

Adds a page that lists stats per product for a given date range. This is admin-only functionality

## Deploy Notes

N/A


mheap:
Merging as it's non-public facing and required for today

PR #1199, opened 2018-11-02
Update SMPP concept doc

marklewin:
## Description

Draft SMPP concept doc after discussion with @thepeachbeetle and @parthawasthi. 

## Deploy Notes

Check `Documentation > Messaging > SMS > Guides > SMPP access` in the [review app](https://nexmo-developer-pr-1199.herokuapp.com/): 
Marking as `don't merge` so that Paul and Parth can review.

### DIFF
--- a/_documentation/messaging/sms/guides/SMPP-access.md
+++ b/_documentation/messaging/sms/guides/SMPP-access.md
@@ -1,5 +1,5 @@
 ---
-title: SMPP Access
+title: SMPP access
 ---
 
 # SMPP Access
--- a/_documentation/messaging/sms/guides/SMPP-access.md
+++ b/_documentation/messaging/sms/guides/SMPP-access.md
@@ -1,37 +1,67 @@
 ---
-title: SMPP access
+title: SMPP Access
 ---
 
-# SMPP access
+# SMPP Access
 
-By default, your Nexmo account is configured for REST. Compared to REST, SMPP uses more resources. However, most of the SMS volume processed at Nexmo uses SMPP and if you would prefer to use SMPP, we are happy to assist you to switch your traffic over to us.
+> This concept document describes using SMPP instead of REST to access Nexmo's APIs. Implementing SMPP access requires a deep understanding of this complex telecommunications protocol and a lot of low-level development work. Most developers can safely ignore this information.
 
-If you already use SMPP and want to use Nexmo, you can switch to us in minutes without changing your app. Just update your configuration to point to Nexmo's SMPP server.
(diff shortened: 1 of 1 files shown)


_documentation/messaging/sms/guides/SMPP-access.md
```diff
@@ -41,14 +71,32 @@ To use our configuration:
...
 
-> Note: Check out our [SMPP FAQ] for information about supported character encodings, delivery receipt formats, throttle management, concatenated messages and more.
+## Our SMPP instances
+
+Nexmo hosts three instances of SMPP:
+
+- `SMPP1/2` - clustered
+- `SMPP0` - not clustered
+
+We recommend that you bind to `SMPP1/2`.
```
parthawasthi: We recommend that you bind to both these clusters for ensuring the best possible redundancy. If not feasible, please bind to either one of the two clusters.

marklewin: Just to clarify: my understanding is that `SMPP1/2` is a cluster and `SMPP0` is not? If so, we can't really say "both of these clusters". The [KB article](https://nexmoinc.atlassian.net/wiki/spaces/MSG/pages/590086408/SMPP+Availability+Internal+Communication) suggests that they should not bind to `SMPP0` at all unless they cannot bind to `SMPP1/2`. Is that incorrect?

marklewin:
Thanks very much @parthawasthi. I am adding @lornajane for a general review after Paul has had a chance to look at it. If you could please check my query about your earlier comment I'd appreciate it.

_documentation/messaging/sms/guides/SMPP-access.md
parthawasthi: That is correct, Mark and your choice of words is accurate and more meaningful. What I mean is, we advise customers to bind to (the cluster) SMPP1 and in the event their bind to it goes down, initiate bind with (the cluster) SMPP2; not necessarily in that order. That way even if we lost a whole cluster (SMPP1 or 2) they will not be impacted. If they don't want to (or cannnot) do this, they must bind to either SMPP1 or 2, both of which are clustered, and thereby offer redundancy.

thepeachbeetle: so to clarify, the 'cluster' means  both smpp1 and smpp2 (i.e., they are each 1 half of the cluster),  to be bound to the cluster you should be bound to both smpp1 and smpp2 at all times (and if you are not able to do that, then smtp0 is more appropriate for you (which isn't a cluster at all but is a single standalone endpoint

marklewin:
Thanks @parthawasthi and @thepeachbeetle. I have updated the doc to recommend binding to both `SMPP1` **and** `SMPP2` and only binding to `SMPP0` if you cannot connect to a clustered setup.

PR #1204, opened 2018-11-02
Update to SMS error detail

### LINKED ISSUE #940: SMS error responses missing from docs
## Description

From Carroll: - the SMS errors on the response to the request seem to be missing the docs.  (not the DR errors, the errors in the response) - I can't find them in the SMS API anymore.  e.g.:
```
{
"message-count": "1",
"messages": [
{
"status": "4",
"error-text": "Bad Credentials"
}
]
}
```

And there seems to be a mix-up on https://developer.nexmo.com/api-errors/sms - because this is showing the DR errors, but at the top of the page it says
```When a Nexmo API returns an error, for instance, if your account has no credit, the API will return an error in RFC 7807 format.```
- but I think that's the type of error you get on the response to the request, not the DR.

The possible response codes would be available in the documentation, either in or linked from the SMS API documentation

## What actually happened?

The possible response codes don't seem to be documented.

## Steps to reproduce:

Read the SMS API documentation

## Related resources

https://developer.nexmo.com/api/sms

## Also

The API returns a 200 status code when this happens! Not really a docs problem but not great either and we should probably verify/document this.

lornajane:
## Description

This fixes #940 and follows on from the error docs we added to the site last week (https://developer.nexmo.com/messaging/sms/guides/troubleshooting-sms). The `x-errors` field in the OAS is updated and this can be seen on the URL `/api-errors/sms.

**NOTE** This is entirely unrelated to the error information that appears on in the API spec in its rendered version.

### DIFF
--- a/_open_api/definitions/sms.yml
+++ b/_open_api/definitions/sms.yml
@@ -415,68 +415,78 @@ x-groups:
 
 x-errors:
   1:
-    description: An unknown error was received from the carrier who tried to send this this message. Depending on the carrier, that `to` is unknown.
-    resolution: When you see this error, and status is rejected, always check if `to` in your request was valid.
+    description: Throttled. You are sending SMS faster than the account limit.
+    resolution: Refer to [What is the Throughput Limit for Outbound SMS?](https://help.nexmo.com/hc/en-us/articles/203993598) for more information.
 
   2:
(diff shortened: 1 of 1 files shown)


lornajane:
preview the changes here: https://nexmo-developer-pr-1204.herokuapp.com/api-errors/sms

PR #1205, opened 2018-11-02
Fix deep links when navigating via Turbolinks

mheap:
## Description

When you have a link to “voice/voice-api/guides/call-flow#call-states” in the NCCO reference it won’t jump down the page to ‘call-states’ - you have to press enter on the URL again to make the jump.

This PR fixes the issue

## Deploy Notes

N/A


lornajane:
Weirdly, I can't replicate the problem on `master` right now. Leave that with me ...

mheap:
@lornajane I can replicate on master in both Chrome and Firefox

PR #1211, opened 2018-11-05
Force a full page reload for tutorial links

mheap:
## Description

Turbolinks can't find the sidenav (as it doesn't exist) and throws an
error, meaning that the page is not navigated to.

Initial bug report:

> - broken link to tutorial ‘Sending Facebook Messenger messages with the Messages API’ on https://developer.nexmo.com/messages/building-blocks/send-with-facebook-messenger

## Deploy Notes

N/A

### DIFF
--- a/app/filters/tutorial_link_filter.rb
+++ b/app/filters/tutorial_link_filter.rb
@@ -1,0 +1,19 @@
+class TutorialLinkFilter < Banzai::Filter
+  def call(input)
+    @input = input
+
+    document.css('a').each_with_index do |link, _index|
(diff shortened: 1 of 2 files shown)


mheap:
@cr0wst I’d we do that, we should remove Turbolinks entirely. At the moment this issue only surfaces with tutorial links as they don’t have the sidebar

I’d rather put a specific fix in for this and defer the larger decision until we have a tech lead

cr0wst:
It would still work for the side navigation, it would just prevent any links contained within the contents from being turbolinked. Fair enough for now.

PR #1229, opened 2018-11-13
Add new SMS API concatenation tutorial

marklewin:
## Description

Added a new tutorial. The source code is [here](https://github.com/Nexmo/sms-node-concat-tutorial).

## Deploy Notes

Check in sample app.

### DIFF
--- a/_tutorials/receiving-concat-sms.md
+++ b/_tutorials/receiving-concat-sms.md
@@ -48,7 +48,7 @@ cd myapp
 npm init
 ```
 
-Press [Enter] to accept each of the defaults.
+Press [Enter] to accept each of the defaults except `entry point` for which you should enter `server.js`.
 
 Then, install the [express](https://expressjs.com) web application framework and the [body-parser](https://www.npmjs.com/package/body-parser) packages:
 
@@ -74,9 +74,10 @@ Use the `ngrok` web interface at <http://localhost:4040> and make a note of the 
 
 ## Create the basic application
 
-Create an `index.js` file in your application directory with the following code, which will be our starting point:
+Create a `server.js` file in your application directory with the following code, which will be our starting point:
 
 ```javascript
+require('dotenv').config();
 const app = require('express')();
 const bodyParser = require('body-parser');
 
@@ -88,11 +89,9 @@ app
     .get(handleInboundSms)
     .post(handleInboundSms);
 
-function handleInboundSms(request, response) {
+const handleInboundSms = (request, response) => {
     const params = Object.assign(request.query, request.body);
 
-    console.log('Inbound SMS received');
-    
     // Send OK status
     response.status(204).send();
 }
@@ -124,37 +123,37 @@ Now, if any of your virtual numbers receive an SMS, Nexmo will call that webhook
 
 ## Send a test SMS
 
-1. Open a new terminal window and run the `index.js` file so that it listens for incoming SMS:
+1. Open a new terminal window and run the `server.js` file so that it listens for incoming SMS:
 
     ```sh
-    node index.js
+    node server.js
     ```
 
 2. Send a test SMS to your Nexmo number from your mobile device, with a short text message. For example, "This is a short text message".
 
-If everything is configured correctly you should receive a `Inbound SMS received` message in the terminal window running `index.js`.
+If everything is configured correctly you should receive a `Inbound SMS received` message in the terminal window running `server.js`.
 
 Now, let's write some code to parse the incoming SMS to see what the message contains.
 
-1. Press [Ctrl+C] to terminate the running `index.js` application.
+1. Press [Ctrl+C] to terminate the running `server.js` application.
 
-2. Create a new function in `index.js` called `displaySms()`:
+2. Create a new function in `server.js` called `displaySms()`:
 
     ```javascript
-    function displaySms(msisdn, text) {
+    const displaySms = (msisdn, text) => {
         console.log('FROM: ' + msisdn);
         console.log('MESSAGE: ' + text);
         console.log('---');
     }
     ```
 
-3. Also in `index.js` and just before your code sends the `204` response, add a call to `displaySms()` using the following parameters:
+3. Also in `server.js` and just before your code sends the `204` response, add a call to `displaySms()` using the following parameters:
 
     ```javascript
     displaySms(params.msisdn, params.text);
     ```
 
-4. Restart `index.js` and then send another short message from your mobile device. This time, you should see the following in the terminal window running `index.js`:
+4. Restart `server.js` and then send another short message from your mobile device. This time, you should see the following in the terminal window running `server.js`:
 
     ```sh
     Inbound SMS received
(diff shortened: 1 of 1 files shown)


_tutorials/receiving-concat-sms.md
```diff
@@ -0,0 +1,398 @@
...
+
+SMS messages that [exceed a certain length](/messaging/sms/guides/concatenation-and-encoding) are split into multiple SMS and transmitted seperately. The receiver must gather all the constituent parts and reassemble them to display the full message.
+
+When you use the SMS API to receive [inbound SMS](/messaging/sms/guides/inbound-sms) that might be longer than the byte-length allowed for a single SMS, you must check to see if the messages delivered to your [webhook](/concepts/guides/webhooks) are standalone or just one part of a multi-part SMS. If there are multiple parts to the message, you must reassemble them to display the full message text.
+
+This tutorial shows you how.
+
+## In this tutorial
+
+In this tutorial, you create a simple Node.js application using the Express framework that receives inbound SMS via a webhook and determines whether the message is a single-part or multi-part SMS.
```
mheap: you *will* create?

_tutorials/receiving-concat-sms.md
```diff
@@ -0,0 +1,398 @@
...
+products: messaging/sms
+description: If an inbound SMS exceeds the maximum length allowed for a single SMS, it is split into parts. It is then up to you to reassemble those parts to show the full message. This tutorial shows you how.
+languages:
+    - Node
+---
+
+
+# Receiving Concatenated SMS
+
+SMS messages that [exceed a certain length](/messaging/sms/guides/concatenation-and-encoding) are split into multiple SMS and transmitted seperately. The receiver must gather all the constituent parts and reassemble them to display the full message.
```
mheap: Is it multiple SMS, or is it a single SMS comprised of multiple parts?

_tutorials/receiving-concat-sms.md
```diff
@@ -0,0 +1,398 @@
...
+## Expose your application to the Internet
+
+When the SMS API receives an SMS destined for one of your virtual numbers, it alerts your application via a [webhook](/concepts/guides/webhooks). The webhook provides a mechanism for Nexmo's servers to communicate with yours.
+
+For your application to be accessible to Nexmo's servers, it must be publicly available on the Internet. A simple way to achieve this during development and testing is to use [ngrok](https://ngrok.com), a service that exposes local servers to the public Internet over secure tunnels. See [this blog post](https://www.nexmo.com/blog/2017/07/04/local-development-nexmo-ngrok-tunnel-dr/) for more details.
+
+Download and install [ngrok](https://ngrok.com), then start it with the following command:
+
+```sh
+$ ./ngrok http 5000
```
mheap: I think most people install `ngrok` globally, in which case `./ngrok` won't work (but `ngrok` will)

_tutorials/receiving-concat-sms.md
```diff
@@ -0,0 +1,398 @@
...
+Enter your webhook endpoint in the field labeled **Webhook URL for Inbound Message** and click the [Save changes] button.
+
+```screenshot
+script: app/screenshots/webhook-url-for-inbound-message.js
+image: public/assets/screenshots/smsInboundWebhook.png
+```
+
+Now, if any of your virtual numbers receive an SMS, Nexmo will call that webhook endpoint with the message details.
+
+## Send a test SMS
```
mheap: Does this work? I didn't think we had Nexmo->Nexmo routing set up.

In the event that it does, I still think that we could cut this section and have the user send an SMS from their personal device to their Nexmo number. This removes the requirement for the Nexmo client lib and any authentication credentials

_tutorials/receiving-concat-sms.md
```diff
@@ -0,0 +1,398 @@
...
+    FROM: TEST-NEXMO
+    MESSAGE: n only.
+    ---
+    ```
+
+What happened? The message exceeded the single SMS byte limit and so was sent as multiple SMS messages.
+
+To enable us to present such messages to our users in the format they were intended, we need to detect if an incoming message has been split in this way and then reassemble it from the parts.
+
+> Notice in the above output that the parts did not arrive in the correct order. This is not uncommon so we need to code our webhook to handle this eventuality.
```
mheap: This is a great example 🙌

_tutorials/receiving-concat-sms.md
```diff
@@ -0,0 +1,398 @@
...
+
+### Detect if a message is concatenated
+
+First, you need to detect if a message is concatenated. Modify the `handleInboundSms()` function so that it displays a single-part SMS to the user in the usual way, but performs extra processing for multi-part SMS which you will implement in a later step:
+
+```javascript
+function handleInboundSms(request, response) {
+    const params = Object.assign(request.query, request.body);
+
+    if (params['concat'] === 'true') {
```
mheap: The strict comparison here could cause issues if the value becomes a boolean rather than a string (similar things have happened in the past). `if (params['concat'])` should be enough

_tutorials/receiving-concat-sms.md
```diff
@@ -0,0 +1,398 @@
...
+    // Send OK status
+    response.status(204).send();
+}
+```
+
+### Gather all the message parts
+
+Before we even attempt to reassemble the message from its parts, we need to ensure that we have all the parts for a given message reference. Remember that there is no guarantee that all the parts will arrive in the correct order, so it is not simply a matter of checking if `concart-part` equals `concat-total`.
+
+We can do this by filtering the `concat_sms` array to include only those SMS objects that share the same `concat-ref` as the SMS that we have just received. If the length of that filtered array is the same as `concat-total`, then we have all the parts for that message and can then reassemble them:
```
mheap: Good catch! I'd never have thought about the situation where multiple concatenated messages arrive in the same time frame

_tutorials/receiving-concat-sms.md
marklewin: Agreed!

_tutorials/receiving-concat-sms.md
marklewin: Reworded

_tutorials/receiving-concat-sms.md
marklewin: Changed to `ngrok`

_tutorials/receiving-concat-sms.md
marklewin: This _does_ work. I agree though that it will make everything simpler if we just get readers to send SMS from their mobile devices.

_tutorials/receiving-concat-sms.md
marklewin: Noted, and changed.

_tutorials/receiving-concat-sms.md
marklewin: :)

_tutorials/receiving-concat-sms.md
mheap: We had a chat about this on Slack and the official line is that it's "not supported", so I think we should cut it and get them to send an SMS from their personal device

marklewin:
Incorporated changes requested by @mheap and updated code samples (and accompanying tutorial code [here](https://github.com/Nexmo/sms-node-concat-tutorial)) to ES6.

_tutorials/receiving-concat-sms.md
marklewin: Done!

_tutorials/receiving-concat-sms.md
mheap: `if (params['concat'])` works for truthy values like strings too

```
> x = "true"
> if (x) { console.log("Foo"); }
Foo

> x = false
false
> if (x) { console.log("Foo"); }
undefined
```

PR #1240, opened 2018-11-14
Java Client Version 4 Building Blocks

cr0wst:
## Description

This PR updates the building block examples to be in line with Version 4 of the Java Client. I also added some missing NI examples.


mheap:
It looks like:

* SMS blocks haven't changed (but client auth setup has)
* Call creation/modification hasn't changed, but anything NCCO related now uses a builder. Call filtering does too
* Verify (legacy style) updated to use new auth builder
* NI updated to use the new auth builder

Missing Java entries for:

* /number-insight/building-blocks/number-insight-advanced-async
* /number-insight/building-blocks/number-insight-advanced-async-callback

_examples/number-insight/advanced/java.yml
```diff
@@ -0,0 +1,16 @@
+---
+title: Java
+language: java
+dependencies:
+- 'com.nexmo:client:4.0.0'
+code:
+  source: .repos/nexmo-community/nexmo-java-quickstart/src/main/java/com/nexmo/quickstart/insight/AdvancedInsight.java
+  from_line: 33
```
mheap: This contains the client setup as well as the code. I think it should be `41` to `43`

PR #1249, opened 2018-11-19
Printable version: remove left hand nav when printing tutorials

lornajane:
## Description

I think this was a typo, not sure! But I don't need the menu when I print (the font is too big as well but I've got no idea how to fix that)

### DIFF
--- a/app/assets/stylesheets/layout/_print.scss
+++ b/app/assets/stylesheets/layout/_print.scss
@@ -4,7 +4,7 @@
     overflow: hidden;
   }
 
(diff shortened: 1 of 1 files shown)


mheap:
@lornajane Which tutorial was this an issue on? On the main docs and on tutorials both the ID and the class are present, so this should work as intended without the PR

lornajane:
It was this tutorial https://developer.nexmo.com/tutorials/private-sms-communication and I was printing from Chrome if that matters.

PR #1252, opened 2018-11-19
update cli command for app creation for messages

AlexLakatos:
## Description

## Deploy Notes

### DIFF
--- a/_partials/reusable/link-facebook-to-nexmo.md
+++ b/_partials/reusable/link-facebook-to-nexmo.md
@@ -9,7 +9,7 @@ You will need to paste in a valid JWT. If you don't have one you can create one 
 **1.** Create a temporary application:
 
 ``` shell
-nexmo app:create "Delete Me Later" https://example.com/answer https://example.com/event --keyfile=temp.key
+nexmo app:create "Delete Me Later" https://example.com/answer https://example.com/event --keyfile=temp.key --type=messages
 ```
 
 **2.** Copy the generated Application ID to the clipboard.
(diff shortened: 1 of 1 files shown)


_partials/reusable/link-facebook-to-nexmo.md
```diff
@@ -9,7 +9,7 @@ You will need to paste in a valid JWT. If you don't have one you can create one
 **1.** Create a temporary application:
 
 ``` shell
-nexmo app:create "Delete Me Later" https://example.com/answer https://example.com/event --keyfile=temp.key
+nexmo app:create "Delete Me Later" https://example.com/answer https://example.com/event --keyfile=temp.key --type=messages
```
mheap: We don't have answer and event URLs. We use `inbound_url` and `status_url`

AlexLakatos: We convert those under the hood, the command is valid

mheap: I mean the names of the URLs. They should probably be `/inbound` and `/status` so that people know which order to place them in

tbedford:
My only concern now is which version of CLI do they need to have installed?

AlexLakatos:
The beta version. And it matches the API status. I think we need to add instructions for installing the beta CLI. And maybe re-write the App creation page to use the CLI instead. Not sure where to add all those bits, so I'll leave it up to you @tbedford

tbedford:
That's fine. I'll add instructions on installing the Beta version of the CLI and also update the info on creating applications.

_partials/reusable/link-facebook-to-nexmo.md
AlexLakatos: for some reason this isn't updated with the latest commit, but the URLs are example.com/inbound and example.com/status

The type of app is relevant I think because creating a voice app here doesn't list it in the dashboard under Messages, so if we're telling people to create an app, it should be consistent.

PR #1257, opened 2018-11-21
Better explanation of user creation

mariatechmaniac:
Explain that no params are required
Help them understand that those users are mapped to their users
Give examples of other fields they can use in the obj

### DIFF
--- a/_documentation/conversation/building-blocks/user/create-user.md
+++ b/_documentation/conversation/building-blocks/user/create-user.md
@@ -8,12 +8,18 @@ In this building block you will see how to create a User.
 
 ## Example
 
-Ensure the following variables are set to your required values using any convenient method:
+In order to create a user, no parameters are required in the request's payload.
+In the response you will receive a user UUID & href.
+If you wish you can define the following variables in order to further customise the user object:
 
 Key | Description
 -- | --
 `USER_NAME` | The unique name of the User.
 `USER_DISPLAY_NAME` | The display name of the User.
+
+For example you can use the USER_NAME as a means to map your Nexmo User to your own system's users.
+
+You can enrich the user object by populating even more parameters like channel details(ex. phone number) and image_url to point to a thumbnail. You can find those in the API Ref[https://developer.nexmo.com/api/conversation#createUser]
 
 ```building_blocks
 source: '_examples/conversation/user/create-user'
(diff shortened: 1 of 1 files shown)


_documentation/conversation/building-blocks/user/create-user.md
```diff
@@ -8,13 +8,19 @@ In this building block you will see how to create a User.
 
 ## Example
 
-Ensure the following variables are set to your required values using any convenient method:
+In order to create a user, no parameters are required in the request's payload.
```
mheap: What's our recommended way to create users? Building blocks should promote best practices.

Some libraries will have username and display name as required parameters, so we should decide what experience we want to promote

mariatechmaniac: @mheap 
i have rewritten it

tbedford:
Maria thanks for the input. It feels like this additional material would be better placed in the API Ref as that is the definitive documentation for the call. Building Blocks are never meant to fully document a call and all parameters, they are only to show a single use case. I will also add some inline comments.

_documentation/conversation/building-blocks/user/create-user.md
```diff
@@ -8,13 +8,21 @@ In this building block you will see how to create a User.
...
+You can create a user by defining the following variables as part of the request:
 
 Key | Description
 -- | --
 `USER_NAME` | The unique name of the User.
 `USER_DISPLAY_NAME` | The display name of the User.
 
+Typically, you can utilise the `USER_NAME` as a means to map your Nexmo User to your own system's users.
+If that is not passed in the request, we will generate and populate the field with a unique ID.
+The `USER_DISPLAY_NAME` can be used for your UI.
```
tbedford: We should be describing the individual parameters in the API Reference as that is the definitive guide to all calls/parameters. This is true for all calls, not just create user.

_documentation/conversation/building-blocks/user/create-user.md
```diff
@@ -8,13 +8,21 @@ In this building block you will see how to create a User.
...
+
+You can enrich the user object by populating even more parameters like channel details(ex. phone number) and image_url to point to a thumbnail. You can find those in the API Ref[https://developer.nexmo.com/api/conversation#createUser] though in general, no parameters are required in the request's payload in order to create your users.
```
tbedford: More text required for the parameters and their application in the API Reference. The API Reference is also clear on which parameters are "Required" and which are optional.

mheap:
Closing in favour of a CS API review

PR #1258, opened 2018-11-22
Create app using CLI

tbedford:
## Description

Initial go at adding support for command line app creation.

- [x] Add create app using CLI
- [x] Move some concepts into building blocks (as they were more task-based)
- [x] Added concept on external linking


_partials/reusable/create-a-nexmo-application.md
```diff
@@ -1,19 +1,36 @@
...
-4. Enter the URL for your inbound message webhook, for example, http://example.com/webhooks/inbound-message.
+4. Enter the URL for your inbound message webhook, for example, `https://example.com/webhooks/inbound-message`.
 
 5. Click the link *Generate public/private key pair* just below the public key text area. This will create a public/private key pair and the private key will be downloaded by your browser.
 
 6. Click the *Create Application* button. You will be taken to phase 2 of the Create Application procedure where you can link external accounts such as Facebook to this application. If you do not have any external accounts listed here, or do not want to link one of those accounts to this application you can click *Done*.
 
 7. If there is an external account you want to link this application to, click the corresponding *Link* button and then click *Done*.
+
+### How to create a Messages and Dispatch application using the Nexmo CLI
```
mheap: This should be above dashboard imo. It's our preferred way of creating/configuring applications and allows for some use cases that the dashboard doesn't e.g. multiple API keys

_documentation/messages/building-blocks/create-an-application.md
```diff
@@ -0,0 +1,11 @@
+---
+title: How to create a Nexmo Messages and Dispatch Application
```
mheap: This feels more like a concept than a building block

_partials/reusable/create-a-nexmo-application.md
tbedford: Fixed.

_documentation/messages/building-blocks/create-an-application.md
tbedford: It was a concept but I moved it. The reason for doing this is there were a collection of items like creating an app, installing lib, installing CLI etc. that were carrying out a task, rather than . purely conceptual. I decided to move them all into the task-based section (building blocks), because they are not long enough to warrant a tutorial, and I wanted similar types of topics in the same place. I thought they looked "right" in that section!

tbedford: Let's give it a go there, if a customer complains I can move them back.

PR #1266, opened 2018-11-26
Add warning about US Short Codes for alerts

mheap:
## Description

AT&T are deprecating shared short codes for alerting purposes. This PR adds a large banner on any alert related page

https://nexmo-developer-pr-1266.herokuapp.com/api/sms/us-short-codes/alerts/subscription

https://nexmo-developer-pr-1266.herokuapp.com/api/sms/us-short-codes/alerts/sending

https://nexmo-developer-pr-1266.herokuapp.com/messaging/us-short-codes/guides/alerts

## Deploy Notes

N/A

### DIFF
--- a/_api/sms/us-short-codes/alerts/sending.md
+++ b/_api/sms/us-short-codes/alerts/sending.md
@@ -4,6 +4,10 @@ description: Reference guide for Event Based Alerts.
 api: Alerts API - Sending
 ---
 
+```partial
+source: _partials/discontinued-alert-shortcode.md
+```
+
 # Alerts - Sending API Reference
 
 This defines Event Based Alerts API:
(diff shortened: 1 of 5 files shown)


mheap:
I'm not sure we want to update the glossary. The glossary is the definition of the shared short code, but not how we can use them.

This PR is currently blocked on a dashboard release

lornajane:
The glossary says "A Short Code that is instantly available to Customers" <-- my understanding is that we no longer offer these, or that there are limitations now in place. It seemed worth calling this out or linking to more docs or something. It's your call though.

mheap:
They're still available for 2FA based use cases, but we need much more clarification around what US SC are used for I think. I'd leave this for now and we can revisit when we're looking at the US Short Code docs as a whole

PR #1274, opened 2018-11-30
Add additional information + example to multichannel recording guide

mheap:
## Description

From feedback:

> Um...you mention we can record multi-channel and yet we can't find how you do that with attributes or anything anywhere.

This adds more information about the NCCO to use, and links to the `record` action in the NCCO reference

## Deploy Notes

N/A

### DIFF
--- a/_documentation/voice/voice-api/guides/recording.md
+++ b/_documentation/voice/voice-api/guides/recording.md
@@ -41,6 +41,46 @@ When recording a call, you can enable split recording which will result in the r
 ## Multi channel recording
 
 When recording a call, you can enable multichannel recording which allows up to 32 call legs to be recorded separately. One file with the number of channels set will be returned.
+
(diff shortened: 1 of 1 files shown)


_documentation/voice/voice-api/guides/recording.md
```diff
@@ -41,6 +41,46 @@ When recording a call, you can enable split recording which will result in the r
...
+      {
+        "type": "phone",
+        "number": "447700900002"
+      }
+    ]
+  }
+]
+```
+
+If you added another `connect` action to this NCCO, the first two participants would appear in their own channels, whilst participants three and four would appear in channel 3 together. To make each participant have their own channel, update the `channels` parameter to be `channels: 4` (or however many participants you have, with a maximum of 32).
```
marklewin: A very minor criticism, but I would amend the second sentence slightly to make it obvious that it is a continuation of the example in the first sentence. Otherwise someone skimming the page might read this as "setting `channels: 4` gives all participants in a call their own channel."

Perhaps something like "In this scenario, you would need to increase the number of `channels` to 4 to give each participant their own channel."

PR #1278, opened 2018-12-03
change exp datetime to add 86400 secs

lukeoliff:
## Description

Changed the example `new Date().getTime()` which returns ms not seconds to `Math.round(new Date().getTime()/1000)+86400`.

As `time()` in PHP returns seconds and `new Date().getTime()` in JavaScript return miliseconds for these to be equivalent it should devide `getTime()` by `1000` or `+86400000`.

I opted for this, so the modification was the same.

## Deploy Notes

Just a code example MD change.


lornajane:
Indeed, the expiry should be seconds since the epoch, not Javascript's milliseconds.

PR #1281, opened 2018-12-04
Adds Send WhatsApp Template

tbedford:
## Description

Adds a 'Send WhatsApp Template' building block plus some minor fixes.


mheap:
@tbedford That's merged now, so this PR can be updated

PR #1284, opened 2018-12-07
[WIP] Add more words to describe our APIs

### LINKED ISSUE #980: API Reference Page Layout (discussion)
## Description

The grid layout on the APIs page makes (IMO and I know this is subjective!) it harder to find things since you need to scan all over the page rather than read a simple list.

Compare the existing version https://developer.nexmo.com/api with the proposed version https://nexmo-developer-staging-pr-825.herokuapp.com/api

A few people have mentioned this and I think we need to discuss this more. The same could apply to the tutorials page https://nexmo-developer-staging-pr-825.herokuapp.com/tutorials which is "pinterest style" on both versions of the site

lornajane:
## Description

To make it easier for users to find the API they need to achieve a particular outcome, I've started on more descriptive wording with each link to API reference pages. I've also changed the layout to accomodate the additional wordiness and to try to make better use of the space. This needs feedback on the general idea and particularly on the API descriptions/wording for each API. It will also need some frontend work to do something with the icons/badges/etc!!

Note that Secret Management is missing as I'm also incorporating it into the account API.

Fixes #980

### DIFF
--- a/app/views/api/index.html.erb
+++ b/app/views/api/index.html.erb
@@ -2,116 +2,170 @@
 <h1>API Reference</h1>
 
 <div class="Vlt-grid Vlt-grid-center">
-    <div class="Vlt-col Vlt-col--1of2 Vlt-col--M-1of2">
-        <h2><svg class="Vlt-icon Vlt-icon--small Vlt-purple"><use xlink:href="/symbol/volta-icons.svg#Vlt-icon-message"/></svg>Messaging<h2>
+  <div class="Vlt-col Vlt-col--1of2 Vlt-col--M-1of2">
+    <div class="Vlt-card Vlt-card--image">
+      <div class="Vlt-card__image Vlt-card__image--app Vlt-card__image--purple">
+        <div class="Vlt-card__image__icon">
+          <svg><use xlink:href="/symbol/volta-icons.svg#Vlt-icon-message"/></svg>
+        </div>
+      </div>
+      <div class="Vlt-card__content">
+        <h2 id="messaging-apis">
+          <a href="/api/sms">
+            Messaging
+          </a>
+        </h2>
+        <nav>
+          <a href="/api/messages-olympus" class="Nxd-card__link">
+            <h5>Messages API <span class="Vlt-badge Vlt-badge--green">Beta</span></h5>
+            <p>For SMS, MMS, Facebook, WhatsApp, Viber and much more: Messages API allows easy integration with multiple messaging platforms through a single integration. This API is currently in Beta, so expect some changes before it is production-ready.</p>
+          </a>
 
-        <a href="/api/messages-olympus" class="Nxd-card__link">
-        <h3>Messages API <span class="Vlt-badge Vlt-badge--green">Beta</span></h3>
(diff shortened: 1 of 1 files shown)


app/views/api/index.html.erb
```diff
@@ -1,244 +1,152 @@
...
+        <h2><svg class="Vlt-icon Vlt-icon--small Vlt-purple"><use xlink:href="/symbol/volta-icons.svg#Vlt-icon-message"/></svg>Messaging<h2>
+
+        <a href="/api/messages-olympus" class="Nxd-card__link">
+        <h3>Messages API <span class="Vlt-badge Vlt-badge--green">Beta</span></h3>
+        <p>For SMS, Facebook, WhatsApp, Viber and much more: Messages API allows easy integration with multiple messaging platforms through a single integration. This API is currently in Beta, so expect some changes before it is production-ready.</p>
+        </a>
+
+        <a href="/api/dispatch" class="Nxd-card__link">
+        <h3>Dispatch API <span class="Vlt-badge Vlt-badge--green">Beta</span></h3>
+        <p>Pair the Dispatch API with the Messages API for a powerful fallback-retry workflow engine to go with your multichannel messaging solution</p>
```
tbedford: You can use Dispatch stand-alone, so you don't need to be aware of Messages API to use it. However, it is true Messages API is used underneath. So you could just say:

"Use the Dispatch API for a flexible multichannel messaging solution with failover."

app/views/api/index.html.erb
```diff
@@ -1,244 +1,152 @@
...
+        </a>
+
+        <a href="/api/redact" class="Nxd-card__link">
+        <h3>Redact API</h3>
+        <p>Ensure your organisation is compliant with data regulations such as GDPR by redacting individual's data from the Nexmo platform.</p>
+        </a>
+
+        <a href="/api/audit" class="Nxd-card__link">
+        <h3>Audit API</h3>
+        <p>Get a full audit trail of all changes to your account.</p>
```
tbedford: We don't really track everything so probably better as:

"Get an audit trail of changes to your account for security and compliance purposes."

app/views/api/index.html.erb
```diff
@@ -1,244 +1,152 @@
...
+
+        <h2><svg class="Vlt-icon Vlt-icon--small Vlt-blue"><use xlink:href="/symbol/volta-icons.svg#Vlt-icon-code"/></svg>Core Services<h2>
+
+        <a href="/api/application" class="Nxd-card__link">
+        <h3>Application API</h3>
+        <p>Create and manage the applications you use across Nexmo APIs. Create, manage and configure your applications and authentication.</p>
+        </a>
+
+        <a href="/api/conversation" class="Nxd-card__link">
+        <h5>Conversation API</h5>
```
tbedford: Uses h5 but previous heading from h3.

mheap:
If we can get some frontend help to style the icons/titles that could make
the difference. The words/categorisation are a huge improvement so I'd be
tempted to split the work in to two parts. Merge with the cards and new
layout then look at alternative styling options in the future

PR #1290, opened 2018-12-10
Fix a few broken links and titles spotted by the link checker

lornajane:
## Description

I've been running a link checker and here's a new installment of what it spotted, nothing major to note but I did sneak in a link to the webhooks reference for VAPI onto the overview with the other reference documents (for continuity, not just because I couldn't scroll the left hand menu!)

### DIFF
--- a/app/views/building_blocks/_application_messages_dispatch.html.erb
+++ b/app/views/building_blocks/_application_messages_dispatch.html.erb
@@ -4,6 +4,6 @@
   </h5>
 
   <div id="acc<%=id %>"  class="Vlt-js-accordion__content Vlt-accordion__content Vlt-accordion__content--noborder">
-    If you do not have an application you can <a href="http://developer.nexmo.com/messages/concepts/create-an-application">create one</a>. Make sure you also <a href="https://developer.nexmo.com/messages/building-blocks/configure-webhooks">configure your webhooks</a>.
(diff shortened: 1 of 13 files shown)


app/views/building_blocks/_application_messages_dispatch.html.erb
```diff
@@ -4,6 +4,6 @@
   </h5>
 
   <div id="acc<%=id %>"  class="Vlt-js-accordion__content Vlt-accordion__content Vlt-accordion__content--noborder">
-    If you do not have an application you can <a href="http://developer.nexmo.com/messages/concepts/create-an-application">create one</a>. Make sure you also <a href="https://developer.nexmo.com/messages/building-blocks/configure-webhooks">configure your webhooks</a>.
+    If you do not have an application you can <a href="/messages/building-blocks/create-an-application">create one</a>. Make sure you also <a href="https://developer.nexmo.com/messages/building-blocks/configure-webhooks">configure your webhooks</a>.
```
tbedford: I'll need to check this in the review app.

tbedford: Looks good.

mheap: There's another absolute link on this line. Otherwise a solid PR

lornajane: Ah good spot! The aboslute links don't get caught by the checker (this was a full-on broken link) - and since I run it locally, it doesn't catch the localhost:3000 ones either! I try to grep for both periodically and will fix this one quickly

PR #1293, opened 2018-12-11
Make past events render as a grid

### LINKED ISSUE #1283: Past events are in a single column
## Description

Past events are shown in a single column resulting in a lot of wasted real estate on the page.

screen shot 2018-12-07 at 15 56 33

mheap:
## Description

Past events are shown in a single column resulting in a lot of wasted real estate on the page. This renders them as a grid

Resolves #1283

## Deploy Notes

N/A

### DIFF
--- a/app/views/static/past_events.html.erb
+++ b/app/views/static/past_events.html.erb
@@ -1,6 +1,6 @@
 <br>
-<div class="row">
-  <div class="columns small-12">
+<div class="Vlt-grid">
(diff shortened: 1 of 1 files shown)


tbedford:
I don't see any past events here:

https://nexmo-developer-pr-1293.herokuapp.com/community

Where should I be looking?

mheap:
@tbedford I've added some seed data. It's the same event over and over, but the layout should be there now

PR #1294, opened 2018-12-11
Update/extend

kellyjandrews:
## Description

Updates to a few extend items:

- Corrected a bad URL and added redirect
- Updated the CTA on pages to be customizable


app/controllers/extend_controller.rb
```diff
@@ -28,8 +30,9 @@ def show
     description = frontmatter['description']
     tags = frontmatter['tags'] || []
     image = frontmatter['image'] || ''
+    cta = frontmatter['cta'] || ''
```
mheap: We could set the default to be `Use This` in the controller and simplify the view

app/controllers/extend_controller.rb
```diff
@@ -7,18 +7,20 @@ def index
     @extensions = document_paths.map do |document_path|
       document = File.read(document_path)
       frontmatter = YAML.safe_load(document)
+      next if frontmatter['published'] == false
```
mheap: This could be `next unless frontmatter['published']`

kellyjandrews: My lack of ruby skills are showing :)

Issue #1300, opened 2018-12-13
OpenAPI spec rendering should fall back to OperationID if Summary is not provided


lornajane:
## Description

Summary is an optional field in OpenAPI, but our rendering relies on it and simply displays blank titles and navigation is not present.

## What did you expect to happen?

Some other operation-specific field would be used

## What actually happened?

Blank spaces

## Steps to reproduce:

Render an OpenAPI spec on NDP where the spec has an operation that does not have a summary field.

mheap:
@lornajane Should we have a standard naming for operation IDs? `camelCase`? `kebab-case`? Bear in mind that these are also used as URL anchors

mheap:
👍 Sounds good to me.

So for an `operationId` of `createSecret` it would show a title of `Create Secret`.

* `createSecret` => `Create Secret`
* `CreateSecret` => `Create Secret`
* `create-secret` => `Create-secret`
* `create_secret` => `Create_secret`

mheap:
One for @benhayehudi

benhayehudi:
I've opened a PR #1313 that works on addressing this issue. Looking forward to reviewing.

benhayehudi:
I think this issue has been resolved with #1313. Is it okay to go ahead and close out this issue?

PR #1313, opened 2018-12-19
New helper method to address summary title issue

benhayehudi:
I have created a new helper method in `application_helper.rb` to address Issue #1300. The method takes in the `summary` and `operationID`. It either returns the `summary`, if it exists, or runs the `operationID` through some string manipulation to produce a new summary title to be used in the rendering of the navigation, if a `summary` does not already exist.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -207,4 +207,30 @@ module ApplicationHelper
     base_url = Rails.env.production? ? 'https://developer.nexmo.com' : request.base_url
     canonical_path.prepend(base_url)
   end
+
+  def normalize_summary_title(summary, operationID)
+    # define the regex for the matches
+    camelMatch = /^[a-zA-Z]\w+(?:[A-Z]\w+){1,}/x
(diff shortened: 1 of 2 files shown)


benhayehudi:
Yes, of course. I'll work on adding those.

app/helpers/application_helper.rb
```diff
@@ -207,4 +207,37 @@ def canonical_url
...
+    # return summary and exit if it is provided
+    if !summary.nil?
+      return summary 
+    end
+
+    if summary.nil?
+      # run through the various possibilities
+      if operationID.match?(camelMatch)
+        revisedSummary = operationID.underscore.split('_').collect{|c| c.capitalize}.join(' ')
+        revisedSummary = revisedSummary.split(' ').collect{|c| uppercase_array.include?(c.upcase) ? c.upcase : c}.join(' ')
```
mheap: This looks the same in every `if` block. We could add it before the `return` once.

In addition, we don't need to `.join(' ')` `revisedSummary` just to split it again on the next line.

PR #1315, opened 2018-12-21
Added another test for AnchorFilter

benhayehudi:
This test ensures that if the input is already in a link format that it returns it, as is, without applying any modifications.


mheap:
Which other link formats should we test are not modified? `camelCase`? `snake_case`?

mheap:
I caught up with @benhayehudi. The first pass of tests will add minimal coverage to increase confidence in deploys. We'll do another round later with additional test cases

PR #1325, opened 2018-12-24
began testing for IndentFilter

benhayehudi:
## Description

Began the testing for #IndentFilter, but ran into the problem that the test input cannot match the expected output because of a disparity in new line breaks.

## Deploy Notes


mheap:
@benhayehudi This test is now passing. There was a missing space in the input (only 3 spaces rather than the required 4)

PR #1329, opened 2018-12-25
Added testing for LabelFilter

benhayehudi:
## Description

Added test coverage for `LabelFilter`, specifically:

* It does not transform a random string
* It returns an HTML `<span>` tag when provided with random text inside brackets
* It doesn't transform a matching string of text when not inside brackets
* It returns color appropriate `<span>` tags when the input is matching exactly

One item to discuss:
* Possible undesired behavior was discovered that when the matching input is provided by it is not all CAPS then it is left unaltered. Is this the desired outcome? A possible solution is to `upcase` the input in the private `class_name` method to eliminate this from happening.

## Deploy Notes

### DIFF
--- a/spec/filters/label_filter_spec.rb
+++ b/spec/filters/label_filter_spec.rb
@@ -25,7 +25,7 @@ RSpec.describe LabelFilter do
     expect(described_class.call(input)).to eq(expected_output)
   end
 
-  it 'makes the span green' do
+  it 'converts [POST] to a green label' do
     input = '[POST]'
 
     expected_output = "<span class='Vlt-badge Vlt-badge--green'>POST</span> "
@@ -33,7 +33,7 @@ RSpec.describe LabelFilter do
     expect(described_class.call(input)).to eq(expected_output)
   end
 
-  it 'makes the span blue' do
+  it 'converts [GET] to a blue label' do
     input = '[GET]'
 
     expected_output = "<span class='Vlt-badge Vlt-badge--blue'>GET</span> "
@@ -41,7 +41,7 @@ RSpec.describe LabelFilter do
     expect(described_class.call(input)).to eq(expected_output)
   end
 
-  it 'makes the span red' do
+  it 'converts [DELETE] to a red label' do
     input = '[DELETE]'
 
     expected_output = "<span class='Vlt-badge Vlt-badge--red'>DELETE</span> "
@@ -49,7 +49,7 @@ RSpec.describe LabelFilter do
     expect(described_class.call(input)).to eq(expected_output)
   end
 
-  it 'makes the span yellow' do
+  it 'converts [PUT] to a yellow label' do
     input = '[PUT]'
 
     expected_output = "<span class='Vlt-badge Vlt-badge--yellow'>PUT</span> "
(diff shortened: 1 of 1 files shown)


spec/filters/label_filter_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  it 'returns a yellow HTML class when string "[PUT]" is provided and puts "PUT" in between <span> tags stripped of brackets' do
+    input = '[PUT]'
+
+    expected_output = "<span class='Vlt-badge Vlt-badge--yellow'>PUT</span> "
+
+    expect(described_class.call(input)).to eq(expected_output)
+  end
+
+  # possible undesired behavior to be discussed
```
mheap: This is intended. We use this label functionality elsewhere on the site that is unrelated to HTTP verbs e.g. `[Beta]` so it should persist the capitalization that is provided

spec/filters/label_filter_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  it 'does not transform a matching string if it is not inside brackets' do
+    input = 'POST'
+
+    expected_output = 'POST'
+
+    expect(described_class.call(input)).to eq(expected_output)
+  end
+
+  it 'returns a green HTML class when string "[POST]" is provided and puts "POST" in between <span> tags stripped of brackets' do
```
mheap: This is fairly verbose. Could we shorten to something like:

> Makes the [POST] label green

That way we're capturing the intent of what we're trying to prove rather than going in to the specifics around a `<span>`. We already know that this filter strips brackets thanks to earlier tests.

spec/filters/label_filter_spec.rb
benhayehudi: Understood. Is it also intended that the end result is the following HTML: `<span class='Vlt-badge '>some text</span>`? If I am understanding correctly, that then is the default result if the input does not match one of the items in the case statement?

spec/filters/label_filter_spec.rb
benhayehudi: Yes, of course. I think you are discovering I'm a bit long winded. 😉

spec/filters/label_filter_spec.rb
mheap: Yep, that's correct

spec/filters/label_filter_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  it 'makes the span green' do
+    input = '[POST]'
+
+    expected_output = "<span class='Vlt-badge Vlt-badge--green'>POST</span> "
+
+    expect(described_class.call(input)).to eq(expected_output)
+  end
+
+  it 'makes the span blue' do
```
mheap: I think we went too far in the other direction with this test name as we don't know why it makes the span blue. Reading the test, I can see that:

> it `converts [GET] to a blue label`

How do you feel about this as a test name?

benhayehudi: Sure, I think that's the right place in the middle between being too verbose and not descriptive enough.

PR #1330, opened 2018-12-25
Language filter testing

benhayehudi:
## Description

Added testing for `LanguageFilter`, specifically:

* It does nothing with non-matching text
* It raises an error if the input is only whitespace
* It correctly transforms matching input
* It does nothing if the input is correct except for the formatting

## Deploy Notes

### DIFF
--- a/spec/filters/language_filter_spec.rb
+++ b/spec/filters/language_filter_spec.rb
@@ -9,12 +9,12 @@ RSpec.describe LanguageFilter do
     expect(described_class.call(input)).to eql(expected_output)
   end
 
-  it 'raises an error if input is whitespace' do
+  it 'returns whitespace unaltered' do
     input = ' '
 
-    expected_output = ' ' # rubocop:disable UselessAssignment
+    expected_output = ' '
 
-    expect { described_class(input) }.to raise_error(ArgumentError)
+    expect(described_class.call(input)).to eql(expected_output)
   end
 
   it 'wraps matching input in a span tag and assigns "lang" to the tag' do
--- a/spec/filters/language_filter_spec.rb
+++ b/spec/filters/language_filter_spec.rb
@@ -1,35 +1,35 @@
 require 'rails_helper'
 
 RSpec.describe LanguageFilter do
-    it 'does nothing to text that does not match the regex' do 
-        input = 'some text'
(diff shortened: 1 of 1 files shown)


spec/filters/language_filter_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    expected_output = 'some text'
+
+    expect(described_class.call(input)).to eql(expected_output)
+  end
+
+  it 'raises an error if input is whitespace' do
+    input = ' '
+
+    expected_output = ' ' # rubocop:disable UselessAssignment
```
mheap: Rubocop was correct, this is a useless assignment. We can delete this line

spec/filters/language_filter_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    expect { described_class(input) }.to raise_error(ArgumentError)
```
mheap: Should this be `described_class.call()` ?

spec/filters/language_filter_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  it 'wraps matching input in a span tag and assigns "lang" to the tag' do
+    input = "[נצמו](lang: 'il')"
+
+    expected_output = "<span lang='il'>נצמו</span>"
+
+    expect(described_class.call(input)).to eql(expected_output)
+  end
+
+  it 'does nothing if the text is in correct format but it is missing the correct formatting' do
```
mheap: How can the text be in the correct format but missing the correct formatting?

benhayehudi: That made sense when I wrote it! 😉 I've updated it now to be more accurate.

spec/filters/language_filter_spec.rb
benhayehudi: Yes, that is correct, and when doing so it no longer raises an error, but returns the whitespace input. I've modified the test to more accurately describe the current behavior.

spec/filters/language_filter_spec.rb
benhayehudi: This has been resolved as a result of the test rewrite.

PR #1332, opened 2018-12-26
Added tests for PartialFilter

benhayehudi:
## Description

Added tests for `PartialFilter`:

* Renders content as HTML when `config['platform']` is set to `true`
* Renders content as markdown when `config['platform']` is not present
* Returns whitespace when whitespace is inputted
* Returns any other content when it does not match regex

## Deploy Notes


benhayehudi:
I've gone ahead and stubbed out a small partial instead of relying on the filesystem and changed the respective tests accordingly.

PR #1335, opened 2018-12-27
Added testing for TechioFilter

benhayehudi:
## Description

Added testing for `TechioFilter`:
* Returns whitespace with whitespace input
* Returns unaltered non-matching input
* Formats matching input with HTML and JS script tags
* Returns `NoMethodError` if input is formatted correctly but empty

## Deploy Notes

### DIFF
--- a/spec/filters/techio_filter_spec.rb
+++ b/spec/filters/techio_filter_spec.rb
@@ -1,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe TechioFilter do
+  it 'returns whitespace with whitespace input' do
+    input = ' '
+
+    expected_output = ' '
+
+    expect(described_class.call(input)).to eql(expected_output)
+  end
+
+  it 'returns unaltered non-matching input' do
+    input = 'some text'
+
+    expected_output = 'some text'
+
+    expect(described_class.call(input)).to eql(expected_output)
+  end
+
+  it 'returns matching input formatted with HTML and JS script tags' do
+    input = <<~HEREDOC
+      ```techio
+      path: /here/is/a/path
+      title: this is a title
+      ```
+    HEREDOC
+
+    expected_output = <<~HEREDOC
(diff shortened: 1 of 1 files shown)


spec/filters/techio_filter_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        <iframe width="100%" frameborder="0" scrolling="no" allowtransparency="true" style="visibility: hidden" src="https://tech.io/playground-widget/here/is/a/path/this is a title"></iframe>
+        <script>if(typeof window.techioScriptInjected==="undefined"){window.techioScriptInjected=true;var d=document,s=d.createElement("script");s.src="https://files.codingame.com/codingame/iframe-v-1-4.js";(d.head||d.body).appendChild(s);}</script>
+      </div>
+    HEREDOC
+
+    # .chop off trailing \n from input
+    expect(described_class.call(input.chop)).to eql(expected_output)
+  end
+
+  it 'returns NoMethodError if input is formatted correctly but empty' do
```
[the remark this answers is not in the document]
benhayehudi: That makes sense and I've added those tests.

spec/filters/techio_filter_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    HEREDOC
+
+    expect { described_class.call(input) }.to raise_error(NoMethodError)
+  end
+
+  it 'returns HTML and JS script tags with no title in the iframe src path if title is missing' do
+    input = <<~HEREDOC
+      ```techio
+      path: /here/is/a/path
+      title:
```
mheap: The `title` is still provided here, but with an empty value. What happens if we don't provide the `title` key at all?

benhayehudi: I've gone ahead and added two new tests to check for this behavior for both title and path.

PR #1339, opened 2019-01-02
Added tests for TutorialsFilter

benhayehudi:
## Description

Adding tests for `TutorialsFilter`:

* Returns encoded string with proper input
* Returns a `Tutorial` with matching input
* Raises error if `layout` specified but cannot be found

More testing to be added

## Deploy Notes


mheap:
If I had to guess, I'd say it's a disk access race condition. We can delete that test as the next one covers it using a mock already

spec/filters/tutorials_filter_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  it 'raises a NoMethodError if no content is provided' do
+    input = <<~HEREDOC
+      ```tutorials
+      ```
+    HEREDOC
+
+    expect { described_class.call(input) }.to raise_error(NoMethodError)
+  end
+
+  it 'returns encoded string even if product cannot be found' do
```
mheap: We should have a mock here even though we're matching against an invalid product

benhayehudi: Gone ahead and added it there.

PR #1343, opened 2019-01-04
[WIP] begin work on default yaml config for landing pages

benhayehudi:
## Description

Begun work on the creation of a default config file for landing pages that defines the blocks and types for a landing page. This file can then be used as the basis for more specific configs such as a `/php` landing page or a certain hackathon landing page.

## Deploy Notes

### DIFF
--- a/config/landing_pages/default.yml
+++ b/config/landing_pages/default.yml
@@ -1,0 +1,64 @@
+# Default landing config page
+
+# Header section
+headers:
+  leaders:
+  subheading:
+  action:
+    header:
+    icon:
+    text:
+    link:
+
+# Column Types
+
+- row:
+  - column: building_block
+    building_block:
+      name: messaging/sms
+      language: ruby
+  - column: building_block
+    building_block:
+      name: messaging/send-an-sms
+      language: python
+
+- row:
+  - column: call_to_action
+    call_to_action:
+      header: "Sign Up"
+      icon: "sign-up"
+      subheader: "Try it for free"
+  - column: call_to_action
+    call_to_action:
+      header: "Join Slack"
+      icon: "slack-icon"
+      subheader:
+  - column: call_to_action
+    call_to_action:
+      header: "Find us on GitHub"
+      icon: "github-icon"
+      subheader:
+
+- row: 
+  - column: developer_resources
+    developer_resources:
+      type: tutorial
+      name: call-tracking
+  - column: developer_resources
+    developer_resources:
+      type: tutorial
+      name: receive-sms-with-ruby
+
+- row: 
+  - column: developer_resources
+    developer_resources:
+      type: sdk
+      name: ruby
+
+- row:
+  - column: text
+    text: Your awesome content
+  - column: text
+    text: More awesome content
+  - column: html
+    html: <a href="/find/me/here">Amazing things</a>


config/landing_pages/default.yml
```diff
@@ -0,0 +1,104 @@
...
+  - columns:
+    - width: 4
+      entries: 
+      - type: header
+        header:
+          lead: "Thanks for joining us!"
+          sublead: "So glad you made it to OneHack Jupiter!"
+          icon: "onehack-icon"
+
+rows:
```
mheap: `rows` is defined above, so we don't need this line. `- columns` is enough to indicate that it's a new row

config/landing_pages/default.yml
```diff
@@ -0,0 +1,104 @@
...
+  - columns:
+    - width: 2
+      entries:
+      - type: building_block
+      building_block:
```
mheap: This indentation is incorrect. It should be at the same level as `type`

config/landing_pages/default.yml
```diff
@@ -0,0 +1,104 @@
...
+# developer_resources: type (i.e. tutorial, sdk), name
+# text: text
+# html: html
+
+title: "Welcome to Nexmo, Residents of Jupiter"
+meta_title: "A longer title goes here."
+
+rows:
+  - columns:
+    - width: 4
```
mheap: We should support a default if `width` isn't provided. (Not adding a `2of3` class will make it full width anyway)

config/landing_pages/default.yml
```diff
@@ -0,0 +1,104 @@
...
+    - width: 2 
+      entries:
+      - type: call_to_action
+      call_to_action:
+        header: "Sign Up"
+        icon: "sign-up"
+        subheader: "Try it for free"
+    - width: 1
+      entries: 
+      - type: call_to_action
```
mheap: It'd be good to have a `join_slack` block type so that we can update it in a single place if needed

config/landing_pages/default.yml
```diff
@@ -0,0 +1,104 @@
...
+      call_to_action:
+        header: "Find us on GitHub"
+        icon: "github-icon"
+        subheader:
+
+rows: 
+  - columns: 
+    - width: 1
+    entries:
+      - type: developer_resources
```
mheap: `developer_resources` handles tutorials and SDKs thanks to the `type` inside the `developer_resources` key. This means we'll need an `if` statement in the rendering, which makes me think that this is actually two different types of block

config/landing_pages/default.yml
```diff
@@ -0,0 +1,104 @@
...
+        developer_resources:
+          type: sdk
+          name: ruby
+
+rows:
+  - columns: 
+    - width: 1
+    entries: 
+      - type: text
```
mheap: Will `text` also support markdown? If so, should it be renamed? If not, should we have a `markdown` block type?

config/landing_pages/default.yml
benhayehudi: As discussed, if `width:` is absent from the config then it will default to full width

config/landing_pages/default.yml
lornajane: Would separating the blocks mean that I couldn't include a list of resources including a tutorial, two building blocks and a link to the relevant language lib? Perhaps rather than starting from yaml we could start from some more sample landing pages and work out how to make those reusable?

config/landing_pages/default.yml
benhayehudi: And as we discussed also a `join_github` block type as well

config/landing_pages/default.yml
benhayehudi: As discussed, `text` will support markdown.

config/landing_pages/default.yml
benhayehudi: I don't think it would impact your ability to do that, it's a question about structuring the code for rendering, but not what you can include. I could be wrong though.

config/landing_pages/default.yml
benhayehudi: Added

config/landing_pages/default.yml
mheap: At the moment `developer_resources` only renders a single item. If you wanted a list of a tutorial, two building blocks and a link to the lib in a single block I think you'd have to create a `html` block for that.

If they were separate blocks you can definitely include them all on a page

config/landing_pages/default.yml
```diff
@@ -0,0 +1,97 @@
...
+    - width: 1
+      entries: 
+      - type: join_slack
+        join_slack:
+          header: "Join Slack"
+          icon: "slack-icon"
+          subheader:
+    - width: 1
+      entries: 
+      - type: go_to_github
```
mheap: What's this intended to link to? An organisation? A repo?

config/landing_pages/default.yml
```diff
@@ -0,0 +1,97 @@
...
+        go_to_github:
+          header: "Find us on GitHub"
+          icon: "github-icon"
+          subheader:
+  - columns: 
+    - width: 1
+    entries:
```
mheap: Indentation here

config/landing_pages/default.yml
```diff
@@ -0,0 +1,97 @@
...
+
+title: "Welcome to Nexmo, Residents of Jupiter"
+meta_title: "A longer title goes here."
+
+rows:
+  - columns:
+    - entries: 
+      - type: header
+        header:
+          lead: "Thanks for joining us!"
```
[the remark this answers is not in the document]
benhayehudi: No, not at all. I agree that sounds clearer.

config/landing_pages/default.yml
benhayehudi: 👍

config/landing_pages/default.yml
```diff
@@ -0,0 +1,97 @@
...
+          sublead: "So glad you made it to OneHack Jupiter!"
+          icon: "onehack-icon"
+  - columns:
+    - width: 2
+      entries:
+      - type: building_block
+        building_block:
+          name: messaging/sms
+          language: ruby
+          pretext:
```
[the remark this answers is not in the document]
benhayehudi: Sure

config/landing_pages/default.yml
benhayehudi: I had in mind the Nexmo gh organization, but it could be a link to a specific repo

benhayehudi: See the following suggestion in the latest commit:

```
      - type: go_to_github
        go_to_github:
          title: "Find us on GitHub"
          subtitle:
          icon: "github-icon"
          reference_type: repo # or org
          link:
```

This gives us the flexibility to direct to either the org or a specific repo, and provide the link. Thoughts?

mheap: Would we want to render a link to a repo the same way as a link to an org? If they're different (which based on the existing repo card-like style I think they would be) we probably want a `github_repo` and a `github_org` type

config/landing_pages/default.yml
```diff
@@ -51,13 +51,10 @@ rows:
...
     - width: 1
       entries: 
-      - type: go_to_github
-        go_to_github:
-          title: "Find us on GitHub"
-          subtitle:
-          icon: "github-icon"
-          reference_type: repo # or org
-          link:
+      - type: github_org
```
mheap: We need to specify the org name here:

```
- type: github_org
  github_org:
    name: nexmo
```

config/landing_pages/default.yml
```diff
@@ -51,13 +51,10 @@ rows:
...
+    - width: 1
+      entries:
+      - type: github_repo
```
mheap: And the repo name here

```
- type: github_repo
  github_repo:
    org: nexmo
    name: nexmo-php
```

config/landing_pages/default.yml
```diff
@@ -0,0 +1,104 @@
...
+        html:
+          content: <a href="/find/me/here">Amazing things</a>
+    - width: 1
+      entries:
+      - type: structured_text
+        structured_text:
+          header: "This is the header text"
+          icon: "icon-lightbulb"
+          text:
+            - main_text: "This is bigger text."
```
mheap: @benhayehudi Did we want to refactor this too?

benhayehudi: Indeed, I hadn't got there yet, expect the commit shortly 🙂

config/landing_pages/hansel_new.yml
```diff
@@ -0,0 +1,54 @@
...
+rows:
+  - columns:
+    - entries: 
+      - type: header
+        header:
+          title: "Thanks, Scott. We'll take it from here!"
+          subtitle: "Welcome, Hanselminutes listeners. Here is everything you need to build your connected applications."
+          icon: "icon-star"
+  - columns:
+    - width: 2
```
mheap: Is having these both be `width: 2` is the same as having them both `width: 1`?

config/landing_pages/hansel_new.yml
```diff
@@ -0,0 +1,54 @@
...
+      - type: call_to_action
+        call_to_action:
+          title: "Nexmo & .Net in action"
+          subtitle: "The Nexmo libraries allow you to get up and running with Nexmo APIs quickly in your language of choice. Examples and inspiration in .NET form"
+          icon: "icon-blog"
+  - columns:
+    - width: 2
+      entries:
+      - type: tutorial
+        tutorial:
```
mheap: How does the `tutorial` type work with images, title + body content?

config/landing_pages/spotlight_new.yml
```diff
@@ -0,0 +1,103 @@
+# Developer Spotlight Landing Page Config
+
+title: "Write for Nexmo"
+meta-title: "meta title goes here"
+
+rows:
+  - columns:
+    - width: 1
+      entries:
+      - type: header
```
mheap: The `Apply Now` button seems to be missing. How would we handle this?

config/landing_pages/spotlight_new.yml
```diff
@@ -0,0 +1,103 @@
...
+          icon:
+    - width: 3
+      entries:
+      - type: html
+        html:
+          content: "<img src=\"/path/to/spotlight-logo.png\" />"
+  - columns:
+    - width: 2
+      entries:
+      - type: html
```
mheap: An image that links to a URL feels like a common block to me. Should we have an abstraction here?

config/landing_pages/spotlight_new.yml
```diff
@@ -0,0 +1,103 @@
...
+      - type: structured_text
+        structured_text:
+          header: "What's in it for You?"
+          icon: "icon-head"
+          text:
+            - content: "We pay $400 USD per post. Your tutorial will be published on our blog, and may be shared in our newsletter or showcased as part of our \"Tutorial Tuesday\" weekly feature."
+              type: large
+            - content: "We make every effort to amplify the reach of each and every blog post on our site, and yours will be no exception; you'll have the Nexmo marketing and social media teams behind you getting the word out about your work. Your success is our success."
+              type: large
+  - columns:
```
mheap: Why do we start a new column here? All of the images/structured text can be in the same two columns

config/landing_pages/hansel_new.yml
benhayehudi: I was imagining it would do a lookup for the tutorial based on the `name` and get the content from the tutorial itself, so that the title, snippet and header image all came from that content. Would that be feasible?

config/landing_pages/spotlight_new.yml
benhayehudi: Perhaps a `call_to_action` block after the `header` block?

config/landing_pages/spotlight_new.yml
benhayehudi: Yes, 100% agree. I can create something and we can go from there.

config/landing_pages/spotlight_new.yml
benhayehudi: Addressed this.

config/landing_pages/spotlight_new.yml
benhayehudi: I addressed this in my latest commits, take a look and let me know what you think.

config/landing_pages/spotlight_new.yml
benhayehudi: Check out what I added in the latest commits and let me know what you think.

config/landing_pages/hansel_new.yml
mheap: Yep, that would work. We don't currently have header images for all tutorials so we need to think about what happens in that case.

We'll also need to make sure we add X with Y blog posts as tutorials in NDP (we have a few examples of this) to make sure we can reference a tutorial by name

config/landing_pages/spotlight_new.yml
```diff
@@ -0,0 +1,117 @@
...
+      - type: action_button
+        action_button:
+          text: "Apply Now"
+          url: "https://developer.nexmo.com/spotlight#submit-your-idea"
+    - width: 3
+      entries:
+      - type: html
+        html:
+          content: "<img src=\"/path/to/spotlight-logo.png\" />"
+  - columns:
```
mheap: Wouldn't this render 8 columns, all with a width of 2 each?

We want two columns, each with multiple entries in it

PR #1345, opened 2019-01-04
Upgrade octicons dependency

mheap:
## Description

This unlocked rack and activejob which had security vulnerabilities

### Rack

CVE-2018-16471
moderate severity
Vulnerable versions: >= 2.0.0, < 2.0.6
Patched version: 2.0.6

There is a possible XSS vulnerability in Rack before 2.0.6 and 1.6.11. Carefully crafted requests can impact the data returned by the scheme method on Rack::Request. Applications that expect the scheme to be limited to 'http' or 'https' and do not escape the return value could be vulnerable to an XSS attack. Note that applications using the normal escaping mechanisms provided by Rails may not impacted, but applications that bypass the escaping mechanisms, or do not use them may be vulnerable.

CVE-2018-16470
moderate severity
Vulnerable versions: >= 2.0.4, < 2.0.6
Patched version: 2.0.6

There is a possible DoS vulnerability in the multipart parser in Rack before 2.0.6. Specially crafted requests can cause the multipart parser to enter a pathological state, causing the parser to use CPU resources disproportionate to the request size.

### Activejob

CVE-2018-16476
moderate severity
Vulnerable versions: >= 5.1.0, < 5.1.6.1
Patched version: 5.1.6.1

A Broken Access Control vulnerability in Active Job versions >= 4.2.0 allows an attacker to craft user input which can cause Active Job to deserialize it using GlobalId and give them access to information that they should not have.

## Deploy Notes

N/A


lornajane:
We approved, but didn't merge. If there's a reason, can someone update and say why?

mheap:
No reason. @benhayehudi if you approve a PR and it's not marked as `don't merge`, please squash or rebase and merge as required

PR #1346, opened 2019-01-04
Updates to the "Running locally" section of the Readme & node version update

pardel:
## Description

This PR includes 2 minor changes and 1 show-stopper (I've installed everything from scratch on a new machine).

**Minor changes:**

- updated required min version for Ruby as defined in `.ruby-version`.
- updated the Rails project setup instructions to use the new [rake proxy for rails cli](https://github.com/rails/rails/pull/22288). I've also compressed all 3 operations into one.

**Show-stopper** 

An incompatibility exists between yarn and node versions as specified in `package.json`:

```
{
  ...
  "engines": {
    "node": "> 8.0.0 < 10.0.0",
    "yarn": ">=0.25.2"
  }
}

```

Error:

```
$ ./bin/yarn install
yarn install v1.12.3
[1/5] 🔍  Validating package.json...
error @1.1.0: The engine "node" is incompatible with this module. Expected version "> 8.0.0 < 10.0.0". Got "11.6.0"
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

The fix was to upgrade the max version of node to 12.0.0.

## Deploy Notes

Not sure if the `package.json` change will have an impact on the deployment process (my JS experience is rather limited).


package.json
```diff
@@ -59,7 +59,7 @@
   },
   "version": "1.1.0",
   "engines": {
-    "node": "> 8.0.0 < 10.0.0",
+    "node": "> 8.0.0 < 12.0.0",
```
mheap: This is a requirement due to some of our dependencies. I can recommend [nvm](https://github.com/creationix/nvm) for managing local node versions. Once installed, run `nvm use 9` before running `yarn install`

PR #1352, opened 2019-01-08
Configurable Landing Pages Feature: Rendering Rows and Columns

benhayehudi:
## Description

Begun work on a configurable landing page rendering. In this PR we add the ability to take in a configuration YAML file and render the correct rows and columns with their appropriate `div` `class` names.

## Deploy Notes


mheap:
@benhayehudi I'd like to see this refactored in to a controller + associated controller tests. The following code should be in `static_controller`, not the test

```
    @config["rows"].each do |row|
      some_columns_have_widths = row['columns'].select {|c| c['width']}.count > 0
      if some_columns_have_widths
        row['columns'] = row['columns'].map { |c| c['width'] ||=1; c }
        row['column_count'] = row['columns'].map { |c| c['width'] }.sum
      end
    end
```

benhayehudi:
@mheap - I've gone ahead and transitioned the view tests into controller tests and moved the logic out of the test. The construction of the `row['column_count]` logic was not passing its value from the controller action to the view, and moving it to the view solved that.

PR #1372, opened 2019-01-25
Configurable Landing Pages: Render Developer Spotlight from YAML Config

benhayehudi:
In this PR we use the new developer spotlight config YAML file and render the Developer Spotlight landing page from it with all of its correct content block types and with the correct rows and columns.


app/views/static/default_landing.html.erb
```diff
@@ -1,30 +1,17 @@
...
         <% end %>
         <% info["entries"].each do |data| %>
-          <% if data["type"] == "text" %>
-            <%= data["text"]["content"] %>
-          <% elsif data["type"] == "html" %>
-            <%= data["html"]["content"] %>
-          <% else%>
-            Unsupported block: <%= data['type'] %>
-          <% end %>
+          <%= render partial: "layouts/partials/landing_pages/#{data['type']}", locals: data["#{data['type']}"] %>
```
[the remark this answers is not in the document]
benhayehudi: That's a great point. I went ahead and added some exception handling for that scenario. 

I would recommend keeping these partials in `layouts/partials` as that is the conventional place `partials` are kept, in the same way, the `api` partial is in `layouts/partials`.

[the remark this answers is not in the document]
mheap: I'd say that the `api` partial is in the wrong place in this instance. Thoughtbots have a great article on application-wide partials (https://thoughtbot.com/blog/directory-for-shared-partials-in-rails) but for page specific partials I believe they should live with the views for that specific endpoint. Having them in `layouts` is misleading to me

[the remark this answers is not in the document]
benhayehudi: Okay, you convinced me! :) 

I've gone ahead and moved them as suggested.

app/views/static/default_landing/partials/_action_button.html.erb
```diff
@@ -0,0 +1,5 @@
+<% raise 'Could not find action button content' if (local_assigns['url'] && local_assigns['text']).nil? %>
```
mheap: This error doesn't tell me how to fix the issue if it occurs. Whilst I appreciate the attempt to keep error checking terse, I think explicit errors would improve this:

```
<%
raise "missing 'url' key in action_button landing page block" unless local_assigns['url']
raise "missing 'text' key in action_button landing page block" unless local_assigns['text']
%>
```

mheap: The `.nil?` check is quite specific. It evaluates if it's not set at all, but what about if it's set with an empty string? Is this a state that we want to raise an error on?

```
irb(main):001:0> true.nil?
=> false
irb(main):002:0> false.nil?
=> false
irb(main):003:0> nil.nil?
=> true
irb(main):004:0> "".nil?
=> false
irb(main):005:0> (true && nil).nil?
=> true
irb(main):006:0> (true && "").nil?
=> false
```

app/views/static/default_landing/partials/_header.html.erb
```diff
@@ -0,0 +1,9 @@
+<% raise 'Could not find header content' if 
+(local_assigns['icon_color'] && local_assigns['icon'] && local_assigns['title']).nil? %>
+<h1 class="Vlt-title--icon">
+    <svg class="Vlt-<%= local_assigns['icon_color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon']%>"></use></svg>
+    <%= local_assigns['title'] %>
+</h1>
+<p>
+<%= local_assigns['subtitle'] %>
```
mheap: We don't want the `<p>` tags if a subtitle isn't provided

app/views/static/default_landing.html.erb
```diff
@@ -1,30 +1,21 @@
...
           <div class="Vlt-col Nxd-landing-col">
         <% end %>
         <% info["entries"].each do |data| %>
-          <% if data["type"] == "text" %>
-            <%= data["text"]["content"] %>
-          <% elsif data["type"] == "html" %>
-            <%= data["html"]["content"] %>
-          <% else%>
-            Unsupported block: <%= data['type'] %>
+          <% if lookup_context.exists?(data['type'], "static/default_landing/partials", true) %>
```
mheap: As `raise` stops execution, we can remove an `else` block from this:

```
<%
unless lookup_context.exists?(data['type'], "static/default_landing/partials", true)
  raise "Could not find the #{data['type']} partial file."
end
%>

<%= render partial: "static/default_landing/partials/#{data['type']}", locals: data[data['type']] %>
```

app/views/static/default_landing/partials/_section_header.html.erb
```diff
@@ -0,0 +1,6 @@
+<% raise 'Could not find Section Header title' if local_assigns['title'].nil? %>
```
mheap: Are `icon_color` and `icon` required in this block?

app/views/static/default_landing/partials/_structured_text.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <svg class="Vlt-<%= local_assigns['icon_color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon'] %>"></use></svg>
+    <%= local_assigns['header'] %>
+</h2>
+<% local_assigns['text'] ||= [] %>
+<% local_assigns['text'].each do |text| %>
+    <% if text['type'] == 'large' %>
+        <p class="p-large">
+            <%= text['content'].render_markdown({skip_paragraph_surround: true}) %>
+        </p>
+    <% else %>
```
mheap: We can fail fast here by explicitly checking for `type == small` too. Otherwise if someone provides `lrge` as a typo it'll be rendered in small text and it may be difficult to debug:

```
<% if text['type'] == 'large' %>
  <p class="p-large">
    <%= text['content'].render_markdown({skip_paragraph_surround: true}) %>
  </p>
<% elsif text['type'] == 'small' %>
  <%= text['content'].render_markdown %>
<% else 
  raise "Unknown text type: #{text['type']}"
end %>
```

config/landing_pages/spotlight.yml
```diff
@@ -5,43 +5,51 @@ meta_title: "meta title goes here"
...
+        - type: line_divider
   - columns:
     - width: 2
       entries:
+      - type: section_header
+        section_header:
+          title: "Examples"
+          icon: icon-app
+          icon_color: yellow
       - type: linked_image
```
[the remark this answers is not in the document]
mheap: This still needs updating

spec/filters/indent_filter_spec.rb
```diff
@@ -4,12 +4,7 @@
   it 'turns four whitespaces followed by "->" into indented text' do
```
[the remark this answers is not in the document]
mheap: Was this change intentional?

spec/views/static/default_landing/partials/_action_button.html.erb_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    render partial: '/static/default_landing/partials/action_button.html.erb', locals: {
+      'url' => url_input, 'text' => text_input, 'another_variable' => another_variable
+    }
+
+    expect(rendered).to include('<a class="Vlt-btn Vlt-btn--primary" href="#a-sample-url">')
+    expect(rendered).to include('Click here!')
+    expect(rendered).to_not include('Ignore me')
+  end
+
+  it 'raises an error if all data is not provided' do
```
mheap: This condition will be triggered by `url` or `text` not existing, so we can remove the test

app/views/static/default_landing/partials/_section_header.html.erb
benhayehudi: I think so, should we make them optional?

mheap: I think that they're required too. In that case, there's a missing `raise` check for `icon` and `icon_color`

benhayehudi:
The YAML has been updated for the developer spotlight with the new `linked_image` content.

PR #1386, opened 2019-01-29
Reorder building blocks.

tbedford:
## Description

Building blocks reordered as follows:

1. Start a call
2. Start a server managed call
3. Register to incoming events
4. Answer
5. Decline
6. Hangup
7. Mute
8. Earmuff
9. Listen to call status


brittBarak:
I don't have permissions to approve this, but that's perfect! thank you!

PR #1388, opened 2019-01-29
Client SDK - Tidy up of Voice building blocks format.

tbedford:
## Description

For Client SDK building blocks:

1) Dependencies. Rather than a simple list we need to link to the tutorial which gives details on how to install dependencies.
2) Rather than specify a filename in which to add the code, just say add to your code as the developer will be adding to a complex application consisting of multiple files.
3) Create application - now provides more details and adds section on using existing application with links to tutorials.

### DIFF
--- a/_documentation/client-sdk/in-app-voice/building-blocks/answer-call.md
+++ b/_documentation/client-sdk/in-app-voice/building-blocks/answer-call.md
@@ -8,12 +8,6 @@ In this building block you will see how to answer a Call.
 
 ## Example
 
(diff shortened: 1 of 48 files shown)


tbedford:
x@mheap: I feel there might be a better way to do this than add the link to the Ruby code. I might be missing a feature of the building block config? Also, not sure how to remove the code block of the dependencies section - even if there are no dependencies there is an empty code block displayed.

tbedford:
@brittBarak let me know if the latest changes are OK please. There is also a review app available in Heroku if that helps. Thanks.

PR #1392, opened 2019-01-30
Added Decline Call example for JS

garann:
## Description

Added a JavaScript code example for Declining a Call, to address DEVX-782.

## Deploy Notes

Depends on the example code in https://github.com/Nexmo/client-sdk-javascript-building-blocks


brittBarak:
@tbedford  re: the naming - This entire block is supposed to be called reject(). iOS named it decline() by mistake and they are renaming it for the upcoming release. @pardel is in charge of the rename on the iOS side. 
in Android it will be the same method as the hangup().

PR #1399, opened 2019-01-31
Updated firewall details

peter-nexmo:
Added Dallas DC firewall details, and removed Washington DC firewall details.

## Description

## Deploy Notes


mheap:
Thanks @peter-nexmo! We were thinking about removing these IP addresses and linking to the knowledgebase as the source of truth for this info. Any thoughts @lornajane?

lornajane:
Yes there is an issue open to remove them since we don't get notified of changes. We should point to the knowledgebase resource for the list of IPs

peter-nexmo:
That's a whole other process we are working on, so that Support is notified of these changes also, which we currently are not.

mheap:
Would your current preference be to merge this PR and revisit with a link to the KB in the future?

PR #1416, opened 2019-02-05
Configurable Landing Pages: Render Hanselminutes Landing Page

benhayehudi:
## Description

In this PR we will introduce the Hanselminutes landing page with the new configurable landing page feature. In so doing, we will build the remaining renderer block types for the landing page feature and incorporate additional parameters to existing renderer block types as required.


app/views/static/default_landing/partials/_action_button.html.erb
```diff
@@ -1,8 +1,20 @@
 <% 
     raise "missing 'url' key in action_button landing page block" unless local_assigns['url']
     raise "missing 'text' key in action_button landing page block" unless local_assigns['text']
+    raise "missing 'large' key in action_button landing page block" unless local_assigns['large']
 %>
-
-<a class="Vlt-btn Vlt-btn--primary" href="<%= local_assigns['url'] %>">
-    <%= local_assigns['text'] %>
+<% if local_assigns['center_button'] == 'true' %>
```
mheap: I'd expect these to be booleans, not strings. If we want something 'truthy' we can do:

```
<% if local_assigns['center_button'] %>
```

app/views/static/default_landing/partials/_action_button.html.erb
```diff
@@ -1,8 +1,20 @@
 <% 
     raise "missing 'url' key in action_button landing page block" unless local_assigns['url']
     raise "missing 'text' key in action_button landing page block" unless local_assigns['text']
+    raise "missing 'large' key in action_button landing page block" unless local_assigns['large']
```
mheap: I think this can default to false rather than being required

app/views/static/default_landing/partials/_action_button.html.erb
```diff
@@ -1,8 +1,20 @@
...
-
-<a class="Vlt-btn Vlt-btn--primary" href="<%= local_assigns['url'] %>">
-    <%= local_assigns['text'] %>
+<% if local_assigns['center_button'] == 'true' %>
+    <div class="Vlt-center">
+<% end %>
+<% if local_assigns['large'] == 'false' %>
+    <a class="Vlt-btn Vlt-btn--<%= local_assigns['type'] %>" href="<%= local_assigns['url'] %>">
+<% elsif local_assigns['large'] == 'true' %>
+    <a class="Vlt-btn Vlt-btn--<%= local_assigns['type'] %> Vlt-btn--large" href="<%= local_assigns['url'] %>">
```
mheap: This is almost identical to the prior line. It can be refactored inline

```
<a class="Vlt-btn Vlt-btn--<%= local_assigns['type'] %> <%= local_assigns['large'] ? 'Vlt-btn--large' : '' %>" href="<%= local_assigns['url'] %>">
```

app/views/static/default_landing/partials/_action_button.html.erb
mheap: Do you think being able to center the button is something we want to allow, or one customisation step too far? If we feel that not allowing it for consistency reasons is the better path, we can do that

app/views/static/default_landing/partials/_header.html.erb
```diff
@@ -2,13 +2,31 @@
     raise "Missing 'icon_color' key in header landing page block" unless local_assigns['icon_color']
     raise "Missing 'icon' key in header landing page block" unless local_assigns['icon']
     raise "Missing 'title' key in header landing page block" unless local_assigns['title']
+    raise "Missing 'center_title' key in header landing page block" unless local_assigns['center_title']
```
mheap: Same as above comments

* Should we standardise on either centered or left aligned titles?
* If we allow this as an option, should we have a default of false?

app/views/static/default_landing/partials/_tutorial.html.erb
```diff
@@ -0,0 +1,24 @@
+<% 
+    raise "Missing 'url' key in tutorial landing page block" unless local_assigns['url']
```
mheap: For tutorials, the `url` should be auto-generated as we have deterministic paths for them

app/views/static/default_landing/partials/_tutorial.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    raise "Missing 'image' key in tutorial landing page block" unless local_assigns['image']
```
mheap: Not all tutorials have images

app/views/static/default_landing/partials/_tutorial.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    raise "Missing 'title' key in tutorial landing page block" unless local_assigns['title']
```
mheap: The `title` is all we need for a tutorial. Everything else should be read out of it's front matter

app/views/static/default_landing/partials/_action_button.html.erb
benhayehudi: See comment below. I believe we should allow this as a design choice, but a default is set to `false`.

app/views/static/default_landing/partials/_call_to_action.html.erb
```diff
@@ -0,0 +1,24 @@
+<%
+    raise "Missing 'icon_color' key in call_to_action landing page block" unless local_assigns['icon_color']
+    raise "Missing 'icon' key in call_to_action landing page block" unless local_assigns['icon']
+    raise "Missing 'title' key in call_to_action landing page block" unless local_assigns['title']
+%>
+
+<div class="center">
+    <h2>
+        <a href="/tools">
```
mheap: Do we always link to the tools page in a CTA?

app/views/static/default_landing/partials/_call_to_action.html.erb
```diff
@@ -0,0 +1,24 @@
...
+                <use xlink:href="<%= local_assigns['icon'].start_with?("Brand") ? "/symbol/volta-brand-icons.svg##{local_assigns['icon']}" : "/symbol/volta-icons.svg#Vlt-#{local_assigns['icon']}" %>"></use>
+                </svg>
+            </div>
+    <%= local_assigns['title'] %>
+        </a>
+    </h2>
+<% if local_assigns['subtitle'] %>
+    <p><%= local_assigns['subtitle'].render_markdown %></p>
+<% end %>
+<% if local_assigns['large_text'] %>
```
mheap: As this is the only text in the block, should it be called `text` rather than `large_text`? The fact that it's large is an implementation detail (that we may change in the future to make it bold rather than large)

app/views/static/default_landing/partials/_github_repo.html.erb
```diff
@@ -0,0 +1,32 @@
+<%
+    raise "Missing 'repo_url' key in github_repo landing page block" unless local_assigns['repo_url']
+    raise "Missing 'github_repo_title' key in github_repo landing page block" unless local_assigns['github_repo_title']
+    raise "Missing 'language' key in github_repo landing page block" unless local_assigns['language']
+%>
+
+<div class="Vlt-grid Vlt-grid--center">
+  <div class="Vlt-col Vlt-col--M-1of3 Vlt-col--1of3">
+    <a class="Vlt-card Nxd-github-card Vlt-left" href="<%= local_assigns['repo_url'] %>" data-github="Nexmo/<%= local_assigns['repo_url'].split('/').last %>">
```
mheap: Could we want to link to non-Nexmo repos here? A safer option is `local_assigns['repo.url'].gsub(%r{https://(www)?github.com/}, '')`

app/views/static/default_landing/partials/_structured_text.html.erb
```diff
@@ -1,12 +1,16 @@
 <%
-    raise "Missing 'icon_color' key in structured_text landing page block" unless local_assigns['icon_color']
-    raise "Missing 'icon' key in structured_text landing page block" unless local_assigns['icon']
     raise "Missing 'header' key in structured_text landing page block" unless local_assigns['header']
-    raise "Missing 'text' key in structured_text landing page block" unless local_assigns['text'] 
+    raise "Missing 'text' key in structured_text landing page block" unless local_assigns['text']
+    raise "Missing 'icon' key in structured_text landing page block" if local_assigns['center_text'] == false && (local_assigns['icon'] == '' || local_assigns['icon'].nil?)
```
mheap: Why does the icon depend on `center_text` being false?

app/views/static/default_landing/partials/_tutorial.html.erb
```diff
@@ -0,0 +1,23 @@
...
+%>
+
+<% file = File.read("_tutorials/#{local_assigns['name']}.md") %>
+<% tutorial = YAML.safe_load(file) %>
+
+<a class="Vlt-card" href="<%= tutorial['external_link'] ? tutorial['external_link'] : '/tutorials/' + local_assigns['name'] %>">
+    <% if tutorial['image_url'] %>
+        <img 
+            src="<%= tutorial['image_url'] %>" 
+            class="img-full wp-post-image"
```
mheap: `wp-post-image` isn't a class that we support, so it can be removed

config/landing_pages/hansel.yml
```diff
@@ -1,14 +1,102 @@
...
-sdk:
-  highlight: "true"
-  library_name: "dotnet"
-url:
-  reference: "hansel-dr"
\ No newline at end of file
+# Hanselminutes Landing Page Config
+
+title: "Thanks, Scott. We'll take it from here!"
+meta_title: "Hanselminutes listeners meta-title"
```
mheap: Where is this used?

config/landing_pages/hansel_old.yml
```diff
@@ -0,0 +1,14 @@
+# Config for the Hanselminutes landing page
```
mheap: We can delete this file

config/landing_pages/spotlight.yml
```diff
@@ -9,13 +9,19 @@ rows:
       - type: header
         header:
           title: "Write for Nexmo"
+          center_title: false
```
mheap: This should be the default and omitted from this config

config/landing_pages/spotlight.yml
```diff
@@ -9,13 +9,19 @@ rows:
...
+          center_subtitle: false
+          subtitle_type: "small"
           icon: icon-rocket
           icon_color: blue
       - type: action_button
         action_button:
           text: "Apply Now"
           url: "#submit-your-idea"
+          type: "primary"
+          large: false
```
mheap: Default values that can be removed

config/routes.rb
```diff
@@ -47,9 +47,7 @@
 
   get '/documentation', to: 'static#documentation'
 
-  get '/hansel', to: 'static#podcast'
```
mheap: Let's leave these routes in for now. Using `/welcome/:name` doesn't give us the flexibility that we need (but there is a way to achieve what you intended here. We can implement it together once this PR is merged)

spec/views/static/default_landing/partials/_call_to_action.html.erb_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'rails_helper'
+
+RSpec.describe 'rendering _call_to_action landing page partial' do
+  it 'renders correctly with local variables' do
+    icon_color = 'orange'
```
mheap: We don't need to define variables here. Pass the values directly in to `locals`

spec/views/static/default_landing/partials/_github_repo.html.erb_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'rails_helper'
+
+RSpec.describe 'rendering _github_repo landing page partial' do
+  it 'renders correctly' do
+    repo_url = 'https://sample.url/org/repo-name'
```
mheap: Any time we use `sample.url` we should be using `example.com`, which is reserved for example data. At some point, `sample.url` may be a valid TLD

app/views/static/default_landing/partials/_call_to_action.html.erb
benhayehudi: No, not at all. Good catch!

benhayehudi:
@mheap, I believe I have incorporated all the changes requested in the last review. Please let me know what you think. Thanks!

app/views/static/default_landing/partials/_header.html.erb
```diff
@@ -2,13 +2,24 @@
...
-<h1 class="Vlt-title--icon">
+    <h1 class="Vlt-title--icon <%= 'Vlt-title--center' if local_assigns['center_title'] %>">
     <svg class="Vlt-<%= local_assigns['icon_color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon']%>"></use></svg>
     <%= local_assigns['title'] %>
 </h1>
 <% if local_assigns['subtitle'] %>
-    <p>
-        <%= local_assigns['subtitle'].render_markdown %>
-    </p>
+    <% if local_assigns['center_subtitle'] == false %>
```
mheap: This has duplication of formatting. We should be consistent with how we've handled this earlier e.g. only outputting `<div class="Vlt-center">` `<% if local_assigns['center_subtitle'] %>`

We don't need exact checks here, a truthy value is enough which would let us remove the `else`

app/views/static/default_landing/partials/_header.html.erb
```diff
@@ -2,13 +2,24 @@
...
+        <% if local_assigns['subtitle_type'] == 'small' %> <p> <% else %> <p class="p-large"> <% end %>
+            <%= local_assigns['subtitle'].render_markdown({skip_paragraph_surround: true}) %>
+        </p>
+    <% elsif local_assigns['center_subtitle'] == true %>
+        <div class="Vlt-center">
+            <% if local_assigns['subtitle_type'] == 'small' %> <p> <% else %> <p class="p-large"> <% end %>
```
mheap: Would this be better as a ternary?

```
<p<%= local_assigns['subtitle_type'] == 'small' ? '' : ' class="p-large"' %>>
```

config/landing_pages/spotlight.yml
mheap: This still seems to be here

spec/views/static/default_landing/default_landing.html.erb_spec.rb
```diff
@@ -22,6 +24,7 @@
                         'action_button' => {
                             'text' => 'Test Action Button',
                             'url' => '#test-url',
+                            'large' => false,
```
mheap: This should be a default

spec/views/static/default_landing/default_landing.html.erb_spec.rb
```diff
@@ -12,7 +12,9 @@
                         'type' => 'header',
                         'header' => {
                             'title' => 'Test Title',
+                            'center_title' => 'false',
```
mheap: This should be a default

spec/views/static/default_landing/default_landing.html.erb_spec.rb
```diff
@@ -12,7 +12,9 @@
...
                             'subtitle' => 'Test Subtitle',
+                            'center_subtitle' => false,
```
mheap: This should be a default

spec/views/static/default_landing/partials/_action_button.html.erb_spec.rb
```diff
@@ -2,38 +2,48 @@
...
 RSpec.describe 'rendering _action_button landing page partial' do
   it 'renders correctly with local variables' do
-    url_input = '#a-sample-url'
-    text_input = 'Click here!'
-
-    render partial: '/static/default_landing/partials/action_button.html.erb', locals: { 'url' => url_input, 'text' => text_input }
+    render partial: '/static/default_landing/partials/action_button.html.erb', locals: {
+      'url' => '#a-sample-url',
+      'text' => 'Click here!',
+      'large' => false,
```
mheap: This should be a default

spec/views/static/default_landing/partials/_action_button.html.erb_spec.rb
```diff
@@ -2,38 +2,48 @@
...
+      'type' => 'primary',
```
mheap: This should be a default

config/landing_pages/hansel.yml
```diff
@@ -1,14 +1,107 @@
...
+
+rows:
+  - columns:
+    - entries: 
+      - type: header
+        header:
+          title: "Thanks, Scott. We'll take it from here!"
+          center_title: true
```
mheap: Would this be better as a complex type for title? e.g.

```
title:
  text: "Thanks, Scott. We'll take it from here!"
  align: center
```

This would allow us to align left/center/right (default left if `align` isn't present) without adding multiple different keys.

We'd have to do the same for subtitle

config/landing_pages/hansel.yml
```diff
@@ -1,14 +1,107 @@
...
+          subtitle: "Welcome, Hanselminutes listeners. Here is everything you need to build your connected applications."
+          center_subtitle: true
+          subtitle_type: "large"
+          icon: "icon-star"
```
mheap: If we choose a complex type for `title`, should we do the same for icon / color?

config/landing_pages/hansel.yml
```diff
@@ -1,14 +1,107 @@
...
+    - entries:
+      - type: products
+  - columns:
+    - entries: 
+      - type: line_divider
+  - columns:
+    - entries:
+      - type: structured_text
+        structured_text:
+          header: "Do you have a question?"
```
mheap: I think this will be a common block. Could we extract to something reusable?

config/landing_pages/spotlight.yml
```diff
@@ -10,12 +10,15 @@ rows:
         header:
           title: "Write for Nexmo"
           subtitle: "Teach others, grow as a writer, and help us build the next go-to destination for developers in search of [high quality technical content](https://www.nexmo.com/blog/category/developer/) and inspiring development stories."
+          center_subtitle: false
```
mheap: This should be a default

config/landing_pages/spotlight.yml
```diff
@@ -10,12 +10,15 @@ rows:
...
+          subtitle_type: "small"
```
mheap: Are we defaulting to `small` or `large` here?

spec/views/static/default_landing/partials/_action_button.html.erb_spec.rb
```diff
@@ -2,38 +2,48 @@
...
   it 'ignores local variables provided to it in the rendering if not used' do
-    url_input = '#a-sample-url'
-    text_input = 'Click here!'
-    another_variable = 'Ignore me'
-
     render partial: '/static/default_landing/partials/action_button.html.erb', locals: {
-      'url' => url_input, 'text' => text_input, 'another_variable' => another_variable
+      'url' => '#a-sample-url',
+      'text' => 'Click here!',
+      'large' => false,
```
mheap: Check for defaults

spec/views/static/default_landing/partials/_action_button.html.erb_spec.rb
```diff
@@ -2,38 +2,48 @@
...
   end
 
   it 'raises an error if url is not provided' do
-    text_input = 'Click here!'
-
-    expect { render partial: '/static/default_landing/partials/action_button.html.erb', locals: { 'text_input' => text_input } }.to raise_error("missing 'url' key in action_button landing page block")
+    expect do
+      render partial: '/static/default_landing/partials/action_button.html.erb', locals: {
+      'text' => 'Click here!',
+      'large' => false,
```
mheap: Check for defaults

app/views/static/default_landing/partials/_structured_text.html.erb
benhayehudi: The presence or absence of an icon alters the composition of the `<h2></h2>` block of content. If there is an icon then the `<h2>` tag gets assigned a class of `Vlt-title-icon`, which sets up a text + icon layout. If there is no icon and `center_text` was set to `true` then the `<h2>` block gets a `<div></div>` around it with a class of `Vlt-center`.

config/landing_pages/hansel.yml
benhayehudi: It's not used now, this was added based on your advice to create some meta content for eventual SEO usage.

app/views/static/default_landing/partials/_header.html.erb
benhayehudi: That's true. I've updated it accordingly.

config/landing_pages/hansel.yml
benhayehudi: I like that. It's a much more elegant and flexible way of creating customization.

config/landing_pages/hansel.yml
benhayehudi: I am leaning towards leaving it as two keys, since its only a name and a color, without much more customization. Thoughts?

config/landing_pages/hansel.yml
benhayehudi: I don't see why we need an additional block type for this. How is it significantly different than `structured_text`? Wasn't that the point of `structured_text` to be configurable for all types of structure, including this one, which doesn't deviate much from the use case we've seen so far?

config/landing_pages/hansel.yml
```diff
@@ -1,14 +1,114 @@
...
+title: "Thanks, Scott. We'll take it from here!"
+meta_title: "Hanselminutes listeners meta-title"
+
+rows:
+  - columns:
+    - entries: 
+      - type: header
+        header:
+          title: 
+            - text: "Thanks, Scott. We'll take it from here!"
```
mheap: These should be objects, not lists. Remove the `-` and you can access it as `entry['title']['text'] rather than `entry['title'][0]['text']`

config/landing_pages/hansel.yml
```diff
@@ -1,14 +1,114 @@
...
+              align: center
+          subtitle:
+            - text: "Welcome, Hanselminutes listeners. Here is everything you need to build your connected applications."
```
mheap: Objects, not lists as above

config/landing_pages/hansel.yml
```diff
@@ -1,14 +1,114 @@
...
+              align: center
+              type: large
+          icon: 
+            - name: "icon-star"
```
mheap: Objects, not lists as above

app/views/static/default_landing/partials/_header.html.erb
```diff
@@ -1,14 +1,23 @@
...
     <%= local_assigns['title'] %>
 </h1>
 <% if local_assigns['subtitle'] %>
-    <p>
-        <%= local_assigns['subtitle'].render_markdown %>
-    </p>
+    <% local_assigns['subtitle'] ||= [] %>
+    <% local_assigns['subtitle'].each do |text| %>
+        <% if text['align'] == 'center' %>
+            <div class="Vlt-center">
```
mheap: We could use `<div class="Vlt-<%= text['align'] %>">` and this would support `left`, `right` and `center` with no other changes, plus remove a conditional (at the expense of more markup)

app/views/static/default_landing/partials/_header.html.erb
```diff
@@ -1,14 +1,23 @@
...
-    <svg class="Vlt-<%= local_assigns['icon_color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon']%>"></use></svg>
+    <h1 class="Vlt-title--icon <%= 'Vlt-title--center' if local_assigns['title'][0]['align'] == 'center' %>">
+    <svg class="Vlt-<%= local_assigns['icon'][0]['color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon'][0]['name'] %>"></use></svg>
     <%= local_assigns['title'] %>
 </h1>
 <% if local_assigns['subtitle'] %>
-    <p>
-        <%= local_assigns['subtitle'].render_markdown %>
-    </p>
+    <% local_assigns['subtitle'] ||= [] %>
```
mheap: This is redundant thanks to the `if` check on line 10

app/views/static/default_landing/partials/_structured_text.html.erb
```diff
@@ -1,12 +1,16 @@
 <%
-    raise "Missing 'icon_color' key in structured_text landing page block" unless local_assigns['icon_color']
-    raise "Missing 'icon' key in structured_text landing page block" unless local_assigns['icon']
     raise "Missing 'header' key in structured_text landing page block" unless local_assigns['header']
-    raise "Missing 'text' key in structured_text landing page block" unless local_assigns['text'] 
+    raise "Missing 'text' key in structured_text landing page block" unless local_assigns['text']
+    raise "Missing icon 'name' key in structured_text landing page block" if local_assigns['center_text'] == false && (local_assigns['icon'][0]['name'] == '' || local_assigns['icon'][0]['name'].nil?)
```
mheap: This still feels wrong to me. Why can't we support a centered header with an icon? They feel like separate config items to me

config/landing_pages/hansel.yml
mheap: We should set it to a real value so that when it is used we don't forget about it

config/landing_pages/hansel.yml
mheap: It's about consistency. We use it on multiple pages and want them to be in sync like the Slack block

app/views/static/default_landing/partials/_call_to_action.html.erb
```diff
@@ -0,0 +1,36 @@
...
+                </svg>
+            </div>
+    <%= local_assigns['title'] %>
+        </a>
+    </h2>
+<% if local_assigns['subtitle'] %>
+    <p><%= local_assigns['subtitle'].render_markdown %></p>
+<% end %>
+<% if local_assigns['text'] %>
+    <% local_assigns['text'] ||= [] %>
```
mheap: This is redundant due to the `if` check. We can use one or the other

app/views/static/default_landing/partials/_header.html.erb
```diff
@@ -1,14 +1,20 @@
...
-    raise "Missing 'icon_color' key in header landing page block" unless local_assigns['icon_color']
-    raise "Missing 'icon' key in header landing page block" unless local_assigns['icon']
+    raise "Missing icon 'color' key in header landing page block" unless local_assigns['icon']['color']
+    raise "Missing icon 'name' key in header landing page block" unless local_assigns['icon']['name']
     raise "Missing 'title' key in header landing page block" unless local_assigns['title']
 %>
-<h1 class="Vlt-title--icon">
-    <svg class="Vlt-<%= local_assigns['icon_color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon']%>"></use></svg>
-    <%= local_assigns['title'] %>
+    <h1 class="Vlt-title--icon <%= 'Vlt-title--center' if local_assigns['title']['align'] == 'center' %>">
```
mheap: Do we want to check explicitly for `center` here, or use `Vlt-title--<%= local_assigns['title']['align'] %>`?

app/views/static/default_landing/partials/_header.html.erb
```diff
@@ -1,14 +1,20 @@
...
+    <svg class="Vlt-<%= local_assigns['icon']['color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon']['name'] %>"></use></svg>
+    <%= local_assigns['title']['text'] %>
 </h1>
 <% if local_assigns['subtitle'] %>
-    <p>
-        <%= local_assigns['subtitle'].render_markdown %>
+    <% if local_assigns['subtitle']['align'] %>
+        <div class="Vlt-<%= local_assigns['subtitle']['align'] %>">
+    <% end %>
+    <p class="<%= local_assigns['subtitle']['type'] != 'small' ? 'p-large' : '' %>">
```
mheap: Is `large` the most common default here, or should we default to small?

app/views/static/default_landing/partials/_section_header.html.erb
```diff
@@ -1,10 +1,10 @@
...
     raise "Missing 'title' key in section_header landing page block" unless local_assigns['title']
-    raise "Missing 'icon_color' key in section_header landing page block" unless local_assigns['icon_color']
-    raise "Missing 'icon' key in section_header landing page block" unless local_assigns['icon']
+    raise "Missing icon 'color' key in section_header landing page block" unless local_assigns['icon'][0]['color']
+    raise "Missing icon 'name' key in section_header landing page block" unless local_assigns['icon'][0]['name']
 %>
 
 <h2 class="Vlt-title--icon">
-    <svg class="Vlt-<%= local_assigns['icon_color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon'] %>"></use></svg>
+    <svg class="Vlt-<%= local_assigns['icon'][0]['color'] %>"><use xlink:href="/symbol/volta-icons.svg#Vlt-<%= local_assigns['icon'][0]['name'] %>"></use></svg>
```
mheap: This is still expecting a list

config/landing_pages/hansel.yml
mheap: We don't want any parameters in this block, as we want it to be consistent across all pages. Hard code the content, and render with just `type: contact_support`

PR #1432, opened 2019-02-08
MMS LVN update clarity on Short Codes

martyndavies:
## Description

Text change to clarify that T-Mobile and Verizon can be sent to on US Short Code but not US LVN.

## Deploy Notes

Text only change.


mheap:
Are T-Mobile and Verizon supported when using shortcodes? If so I'd call that out specifically

PR #1436, opened 2019-02-11
Text Content Renderer for Config Driven Landing Pages

benhayehudi:
## Description

Added last content renderer type to config-driven landing pages: `text`, along with accompanying tests.


spec/views/static/default_landing/partials/_text.html.erb_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe 'rendering _text landing page partial' do
```
mheap: Could we add a test to ensure that it renders markdown too?

benhayehudi: Done!

PR #1439, opened 2019-02-11
Documentation for Config Driven Landing Pages

benhayehudi:
## Description

In this pull request, we add documentation for config-driven landing pages, describing the process and each content block.


LANDING_PAGE_GUIDE.md
```diff
@@ -0,0 +1,288 @@
...
+```yaml
+
+- type: action_button
+    action_button:
+      text: # Text for the button (i.e. "Click here!")
+      url: # The URL for the button (i.e. "https://developer.nexmo.com")
+
+```
+
+The `action_button` also takes optional parameters. **You do not need to define these optional parameters, unless you want their value to be `true` or the non-default.**:
```
marklewin: This is confusing and the instructions for how to work with these optional parameters are implicit in the word "optional" and in the description of the default values. I think a simple _The action_button supports the following optional parameters:_ will suffice here.

LANDING_PAGE_GUIDE.md
```diff
@@ -0,0 +1,288 @@
...
+### Call To Action
+
+The `call_to_action` content block creates a specific layout of text with an icon. You might want to use the `call_to_action` to draw attention to a featured area of Nexmo Developer, like our Community page, for example. The `call_to_action` block requires the following parameters:
+
+```yaml
+
+- type: call_to_action
+  call_to_action:
+    title: # Title for the content (i.e. "SDKs and Tools")
+    icon:
```
marklewin: Can we link or otherwise point to lists of available icons, etc?

LANDING_PAGE_GUIDE.md
```diff
@@ -0,0 +1,288 @@
...
+
+- type: tutorial
+  tutorial:
+    name: # Name of tutorial (i.e. mobile-app-invites)
+
+```
+
+### Unordered List
+
+The `unordered_list` content block lets you create a list of items and specify the shape of the list bullet item demarcator. It accepts the following required parameters:
```
marklewin: demarcator => symbol

LANDING_PAGE_GUIDE.md
```diff
@@ -0,0 +1,288 @@
...
+  unordered_list:
+    list:
+      - item: # First item in list
+      - item: # Second item in list
+      - item: # Third item in list
+      - item: # Fourth item in list
+
+```
+
+As mentioned, you can also specify the shape of the list bullet with the `bullet_shape` parameter:
```
marklewin: "shape of the list bullet" => "the bullet symbol"

LANDING_PAGE_GUIDE.md
benhayehudi: Yes, thank you, I added a link!

benhayehudi:
@mheap - Any thoughts on why this is now failing in the Smoke Tests?

mheap:
```
(<unknown>): invalid leading UTF-8 octet at line 1 column 1
```

Looks like an invisible leading character to me. Try opening it in `vim` and see what you can see as the very first character of the file.

Then, when the cursor is over it press `x` to delete it, then `:x` to save and exit vim. Should work fine after that

benhayehudi:
This is the error that comes up for me now, not an unknown character, but this:

```
     Failure/Error: class_name = "#{class_name} navigation-item--#{normalised_title(child).parameterize}"
     
     ActionView::Template::Error:
       undefined method `parameterize' for nil:NilClass
```

mheap:
@benhayehudi It was a weird combination of images in the `views` folder and our navigation generator. All fixed now.

I rebased thinking that it was something fixed on master, but it wasn't. It does however mean you'll need to pull down the latest branch by running:

```
git checkout documentation-for-landing-pages
git fetch
git reset --hard origin/documentation-for-landing-pages
```

I've got a couple of requests before merge:

* Apply @marklewin's suggestions from above
* When providing an example, have the example match the config required to product the screenshot that you're showing
* Have screenshots for all block types

lornajane:
I tried to use this documentation to create a landing page and I now have questions:
 - what is `title` for? How about `meta_title`? I see them in one of the example pages but not in the documentation.
 - I think I need more context in the screenshots, sometimes (e.g. section header) it would have been more helpful to see it in use rather than just a picture of it by itself.

The YAML format is pretty longwinded, and I'm completely confused by the naming. To create a row? Type `columns`. I'll try to make some alternative suggestions on some of this stuff before we build many/any more pages.

Could we include a diagram of how things will be laid out? I really didn't understand that I was essentially creating a table layout, creating a row and then a cell, then putting things vertically below one another in that cell, then optionally adding more cells. Maybe the words are there that describe this but I'm not a web developer and I was very lost for a while here. It didn't feel simpler than writing a markdown page although bits of it do look a lot nicer than they would have if I had done that!

benhayehudi:
Are we good to approve this and get it merged?

PR #1446, opened 2019-02-14
JWT Generator

mheap:
## Description

This PR adds a JWT generator at `/jwt` that takes a private key, application ID and optional `sub` and creates a JWT that's valid for 6 hours (customisable in the interface). It's all done client side, so the private key never leaves your browser.

It definitely needs refactoring, but what we have here is useful so I'm in favour of shipping and fixing later

## Deploy Notes

N/A

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -18,6 +18,10 @@ class StaticController < ApplicationController
     render layout: 'landing'
   end
 
+  def jwt
+    render layout: 'landing'
+  end
+
   def landing
     render layout: 'landing'
   end
--- a/app/javascript/packs/JWTGenerator.jsx
+++ b/app/javascript/packs/JWTGenerator.jsx
@@ -1,0 +1,262 @@
+import React, { Component } from 'react';
+import { KJUR, KEYUTIL} from 'jsrsasign';
+
+class JWTGenerator extends Component {
+
+    constructor(args) {
+        super(args)
+        var tNow = KJUR.jws.IntDate.get('now');
+        var tEnd = tNow + (3600*6)
+        this.state ={
+            sub: '',
+            acl: '',
+            iat: tNow,
+            exp: tEnd,
+            jti: this.generateJti(12),
+            nbf: tNow,
+            application_id: "",
+            invalidPrivateKey: false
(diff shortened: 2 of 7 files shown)


lornajane:
I can't tell from this PR what this change will implement. From the title, I think I'm going to like it but I have questions:
 - what is this? what do we do with it?
 - how can users discover it? Do we need to add more words, links or something?
 - is it a tool? A page? Something else?

mheap:
You're too quick for me! This is a demo branch for the team call later. It's definitely not ready to merge

lornajane:
Sorry to jump in! Just I did have questions and we don't have a strict review process so I thought I should speak now

mheap:
Ok, I think we're there!

> what is this? what do we do with it?

It's a client side JWT generator, which removes the need for customers to install the CLI to generate JWTs

> how can users discover it? Do we need to add more words, links or something?

At the moment they can't. It's available at `/jwt` and it's intended as a debugging tool to be sent to customers by hand for now. Over time we'll add links to relevant pages (such as Client SDK getting started).

> is it a tool? A page? Something else?

It's an embeddable component which we can put on any page. For now, it's on a single page which is located at `/jwt`

leggetter:
Feels like a tools & sdks type thing to me. Or maybe a `/labs` resource.

It should clearly indicate that all work is done within the web browser and not data is transmitted outside of the page.

As mentioned in non GitHub conversations it does need some UX input. That said, it shouldn't be a complete blocker so as long as some feedback from the DevRel team has been provided (or at least sought) then it shouldn't stop it going live.

If we ever consider using HotJar on NDP then we should disable for the JWT page.

A clear mechanism for feedback from NDP users within the page would be valuable.

mheap:
I had @msach22 review and +1. It likely needs more work going forwards and isn't linked from anywhere, but I'm going to merge so that we can start integrating it in to our existing pages

msach22:
I didn't get a chance to review the code, I just did a UI review as requested by Michael! 

Happy to do a code review as well.

mheap:
I think it'd be good to do a pair programming session on the code at some point as it's far from ideal. But, functional > perfect 🚢

PR #1455, opened 2019-02-18
Remove the "nuget package" label from the github_repo landing page content block

benhayehudi:
This pull request removes the `nuget package` label that sits at the top of the `github_repo` content block. 

Before:
screen shot 2019-02-18 at 8 45 39

After:
screen shot 2019-02-18 at 8 46 18


mheap:
We want to be able to show this on the .NET package. Could we make it configurable?

benhayehudi:
> We want to be able to show this on the .NET package. Could we make it configurable?

Do we only want to show it on the .NET package? If so, could we just make it visible for .NET, or would you prefer it to be a user defined option?

mheap:
User defined. We may want to show the npm badge for node, packagist for PHP etc

mheap:
After looking in to how this works (links to Github, not nuget) and shows an inaccurate version, we've agreed that removing is the best way forward

PR #1458, opened 2019-02-18
Update connect-an-inbound-call.md

Namr-ta:
## Description

## Deploy Notes


mheap:
Thanks for this @Namr-ta. It looks as though this is an issue with the node building block, rather than the docs. I'll get it fixed in the code

PR #1482, opened 2019-02-25
Slim sidebar navigation

mheap:
## Description

This PR scopes the sidebar navigation to the current active product, and reformats the top bar links to contain a list of all products

## Deploy Notes

N/A

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -38,7 +38,15 @@ class ApplicationController < ActionController::Base
 
   def set_code_language
     return unless request.params[:code_language]
-    @code_language = CodeLanguage.find(request.params[:code_language])
+
+    # If it errors, assume that there is no code language and it's just a greedy regex
+    # And use what we captured as part of the document
+    begin
+        @code_language = CodeLanguageResolver.find(request.params[:code_language])
+      rescue StandardError
+        @code_language = nil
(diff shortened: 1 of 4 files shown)


cr0wst:
I like this idea. A few notes right now that I mentioned in Slack:

Mobile is broken. The extra links don't scroll like they do on the current deployment.
image
image

Where did the search bar go?

Overall I like the idea. I'm not a fan of the current left navigation and this improves it quite a bit.

mheap:
Search isn't enabled on review apps as there's no Algolia key but I can enable for this app specifically so that we can see how it looks.

The nav needs work on mobile, but if we think this is a good direction to go in I'll spend some time polishing up this PR

lornajane:
Do we need a WIP flag on this or something? I sat down to review without realising this is a starting point rather than a finished offering.  Personally I really like it - if we can call the "Overview" page the name of the product/section that would really help. And we need a "Back to full list" link or something at the top of the nav IMO. It seems to make the left hand menu behave/scroll differently but possibly better.

Do we need to rebase to bring in the changes to tutorial navigation to consider that as well as this?

mheap:
Yeah this is WIP. The overview page titles need updating, I agree.

This is parked for now as the top navigation needs some love, but I'll pick it up towards the end of March

mheap:
Closing until the design time has time to build the required components

PR #1487, opened 2019-02-27
Add Task-based functionality

mheap:
## Description

Add the ability to define task flows based on reusable subtasks. There are no tasks defined in this PR, only the functionality and documentation.

Documentation at https://nexmo-developer-pr-1487.herokuapp.com/contribute/guides/tasks

## Deploy Notes

N/A


app/controllers/markdown_controller.rb
```diff
@@ -17,6 +17,7 @@ def show
     @content = MarkdownPipeline.new({
       code_language: @code_language,
       current_user: current_user,
+      disable_label_filter: params[:namespace].present? # Disable if we're in the contribute section
```
benhayehudi: Should we use `#exists?` here instead of `#present?`. It seems a major difference between the two is that the former does not initialize ActiveRecord, if a record is found, whereas the latter does. (see: https://stackoverflow.com/questions/13186722/what-is-the-difference-between-using-exists-and-present-in-ruby)

mheap:
@benhayehudi Fixed! Please merge whenever you're ready

mheap:
@benhayehudi Looks like `.present?` is the correct check to use on strings. See https://travis-ci.org/Nexmo/nexmo-developer/builds/500028655?utm_source=github_status&utm_medium=notification

PR #1494, opened 2019-03-01
Simplify ApplicationHelper#search_enabled?

timcraft:
~~ALGOLIA_CONFIG is only defined if ENV['ALGOLIA_SEARCH_KEY'] is set so the first line is redundant.~~ Double negation simplifies the boolean logic.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -7,9 +7,7 @@ COLLAPSIBLE = ['Messages API', 'Dispatch API', 'Messaging', 'SMS', 'Conversion A
 
 module ApplicationHelper
   def search_enabled?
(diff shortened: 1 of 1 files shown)


mheap:
@timcraft The original behaviour is correct - we need both `ALGOLIA_CONFIG` *and* `ENV['ALGOLIA_SEARCH_KEY']` to enable search. If either are missing then we don't have enough information to run the search

timcraft:
@mheap Yep, my mistake I didn't look at the rescue block.

So just `defined?(ALGOLIA_CONFIG)` would be correct logic?

timcraft:
I've updated to match the logic in `config/initializers/algolia.rb`

PR #1499, opened 2019-03-04
Refactored ScreenshotFilter to check if file exists

benhayehudi:
Presently, `ScreenshotFilter` does not check if the file provided by the user exists in the file system. In this PR, we add the functionality to check for the presence of the file. Additionally, the spec testing for the filter has been updated to mock both conditions: when the file exists and when it does not.

### DIFF
--- a/spec/filters/screenshot_filter_spec.rb
+++ b/spec/filters/screenshot_filter_spec.rb
@@ -2,9 +2,10 @@ require 'rails_helper'
 
 RSpec.describe ScreenshotFilter do
   it 'renders image markdown if image location is present in input' do
+    expect(File).to receive(:file?).and_return(true)
     input = <<~HEREDOC
       ```screenshot
-      image: /a/path/to/an/image.png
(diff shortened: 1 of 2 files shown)


spec/filters/screenshot_filter_spec.rb
```diff
@@ -41,4 +42,24 @@
     # .chop to remove trailing \n from input
     expect(described_class.call(input.chop)).to eql(expected_output)
   end
+
+  it 'provides instructions if image cannot be found on disk' do
```
mheap: This test doesn't match the description. I'd expect the following:

`expect(File).to receive(:file?).and_return(true)`

```
input = <<~HEREDOC
      ```screenshot
      image: '/a/path/to/an/image.png'
      ```
```

Where a path is provided, but it doesn't exist on disk

benhayehudi: I believe I've now addressed this in the latest commit.

PR #1503, opened 2019-03-04
 Client SDK tutorial Outbound PSTN

pardel:
## Description

Added "How to Make Phone Calls with the Nexmo Client SDK on iOS" tutorial.

## Deploy Notes

No changes.

### DIFF
--- a/_tutorials/client-sdk-ios-outbound-pstn.md
+++ b/_tutorials/client-sdk-ios-outbound-pstn.md
@@ -29,7 +29,7 @@ Both URLs need to return JSON and follow the [Nexmo Call Control Object (NCCO)](
 
 A [Nexmo virtual number](https://developer.nexmo.com/numbers/overview) will be associated with the app and serve as the "reference point" to it - this is the number that will be the callee when you'll be testing the application.
 
-For more information on Nexmo applications please visit the Nexmo [API Reference](https://developer.nexmo.com/api/application).)
+For more information on Nexmo applications please visit the Nexmo [API Reference](https://developer.nexmo.com/api/application).
 
 
 ## Prerequisites
@@ -74,6 +74,7 @@ Once created, add the gist raw URL (make sure you're using the raw version) to y
 source: '_tutorials_tabbed_content/client-sdk/get-started/in-app-voice/outbound-pstn-ios/started-project'
 ```
 
+
 ## Login
 
 Using the Nexmo Client SDK should start with logging in to `NexmoClient`, using a `jwt` user token.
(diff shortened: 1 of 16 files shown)


_tutorials/client-sdk-ios-outbound-pstn.md
```diff
@@ -0,0 +1,153 @@
...
+
+You can now make an App-to-Phone call.
+
+The `Call` button press is already connected to `ViewController`.
+
+```tabbed_content
+source: '_tutorials_tabbed_content/client-sdk/get-started/in-app-voice/outbound-pstn-ios/start-call'
+```
+
+You are expected to replace `CALLEE_PHONE_NUMBER` with the number to be called but, ultimately, the number called is the one supplied in the `Answer URL` webhook. In a real-life environment, you would create a server component to server as the `Answer URL`; this will receive the `CALLEE_PHONE_NUMBER`, validate it and then supply it in the JSON returned.
```
brittBarak: I'd put this comment somewhere above, closer to the code. because by now the reader will probably copy and paste the code without noticing....

brittBarak: I'd write it a little clearer like:

```
Make sure to update `CALLEE_PHONE_NUMBER ` here, with the same number you placed on the gist provided to the answer_url above. 

```


also, I see now the comment on the real life scenario. it's very good! in my opinion, it's better to be added above though.

_tutorials/client-sdk-ios-outbound-pstn.md
```diff
@@ -0,0 +1,153 @@
...
+
+**NB:** Whilst the default HTTP method for the `Answer URL` is `GET`, `POST` can also be used.
+
+
+### Call Type
```
brittBarak: `Call Type` will be renamed in the upcoming version to Call Handler. we might want to update it already here?

_tutorials/client-sdk-ios-outbound-pstn.md
```diff
@@ -0,0 +1,153 @@
...
+
+Note the use of `NXMCallTypeServer` as the `callType` in the `client`'s `call:` method above - this is required for outbound PSTN calls; the other `callType` is `NXMCallTypeInApp`, useful for making simple calls as shown in [this tutorial](/tutorials/client-sdk-ios-in-app-calling).
```
brittBarak: you can add about `NXMCallTypeServer `, that this means that the logic of the call is defined by the server, which will execute the NCCO defined on the answer url, as defined above. This is rather using of a NXMCallTypeInApp which can only execute a simple in app to in app call.

_tutorials/client-sdk-ios-outbound-pstn.md
```diff
@@ -0,0 +1,153 @@
...
+For the call to happen, `Audio Permissions` are required. In the `appDelegate` of the sample project, you can find an implementation for the permissions request in `application:didFinishLaunchingWithOptions`.  
+
+To read more about the permissions required, [see the setup tutorial](/tutorials/client-sdk-ios-add-sdk-to-your-app#add-permissions).
+
+
+## Conclusion
+
+You have implemented your first App to Phone Voice application with the Nexmo Client SDK for iOS.
+
+Run the app on a simulator and see that you place a call, cancel and hangup that call to a PSTN phone number from the phone number associated with your Nexmo application.
```
brittBarak: cancel = reject?

_tutorials/client-sdk-ios-outbound-pstn.md
pardel: The method parameter is called callType so it might be confusing. I say, let's update it once the change is live, together with updates to the code.

PR #1506, opened 2019-03-04
Add spec for CodeLanguage model

timcraft:
Adds specs for the CodeLanguage model (increases coverage from 80% to 100%).

The CodeLanguage#linkable? method is always returning true—if that's correct then the method and the attribute are redundant and could be removed; if not then that's a bug. The `linkable: false` lines in the config suggest the latter but I'm not 100% sure what the correct behaviour is meant to be.

### DIFF
--- a/app/models/code_language.rb
+++ b/app/models/code_language.rb
@@ -8,7 +8,8 @@ class CodeLanguage
   end
 
   def linkable?
-    @linkable || true
+    return true if @linkable.nil?
+    @linkable
   end
 
   def lexer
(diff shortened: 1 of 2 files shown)


mheap:
@timcraft Yep, definitely a bug! `linkable?` should resolve to false for some entries in that file. Could you take a look?

timcraft:
@mheap Sure! I've added a commit to fix the bug and updated the spec to match.

```
$ bin/rails console
> CodeLanguageResolver.find('cli').linkable?
=> false
> CodeLanguageResolver.find('curl').linkable?
=> true
```

PR #1510, opened 2019-03-05
Add app endpoint and update possible `from` values in NCCO ref

mheap:
## Description

With the addition of IP calls, `from` can now be a string and we have an additional endpoint type of `app` (with a single additional parameter, `user`).

## Deploy Notes

N/A

### DIFF
--- a/_documentation/voice/voice-api/ncco-reference.md
+++ b/_documentation/voice/voice-api/ncco-reference.md
@@ -137,11 +137,11 @@ You can use the following options to control a `connect` action:
 Option | Description | Required
 -- | -- | --
 `endpoint` | Connect to a single endpoint. [Available endpoint types](#endpoint-types-and-values) | Yes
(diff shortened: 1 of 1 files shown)


JamieChapman8462:
@mheap please speak to @cmurphyn regarding this change, as it looks like this isn't what the user will experience.

AlexLakatos:
I'm not sure if `from` accepts a user or it has to still be a Nexmo number.

mheap:
@cmurphyn Can you clarify the behaviour here as this is what I understood from reading the jira ticket

JamieChapman8462:
Ok so I think there is some confusion around this, the `from` in the NCCO isn't changing. This is the `from` the invite for the callee. The `from` for the invite for the callee is populated using the `from` in the `app:knocking` event.

mheap:
@JamieChapman8462 So for IP calls, the `from` in the NCCO is ignored?

cmurphyn:
@mheap The `from` in the connect action NCCO, if present, will be used as a phone number. For the purpose described in the ticket (IP->IP), it should not be included if we want to pass the IP endpoint's `from`.

PR #1511, opened 2019-03-05
Added TOC link to JavaScript SDK.

tbedford:
## Description

Adds link to JavaScript SDK docs back into the TOC.

### DIFF
--- a/_documentation/client-sdk/sdk-documentation/javascript.md
+++ b/_documentation/client-sdk/sdk-documentation/javascript.md
@@ -1,0 +1,10 @@
+---
+title: JavaScript
+redirect: true
+---
+
+<center>
+  <img src="/assets/images/lost.svg" alt="" width="120">
+  <br><br>
+  <h2>Go to the <a href="/sdk/client-sdk/javascript/" target="_blank">JavaScript Client SDK Reference</a></h2>
+</center>


brittBarak:
@tbedford is it possible to open the SDK generated docs on a separate tab?

tbedford:
@brittBarak we use the following link:

``` html
<h2>Go to the <a href="/sdk/client-sdk/android/" target="_blank">Android Client SDK Reference</a></h2>
```
That _should_ open the link in a new tab. I think because of either the redirect or security it does not work. I was looking at [this page](https://www.thesitewizard.com/html-tutorial/open-links-in-new-window-or-tab.shtml), which discusses some security issues around opening content in a new tab, so possibly best avoided.

@kapekost Regarding look and feel of generated docs - I'm guessing that is set by the tools that generate the docs from the code rather than NDP. Not sure how that side of things are done. Perhaps @AlexLakatos knows for JavaScript?

kapekost:
Indeed that's how we generate them at the moment, I remember there were talks that maybe they will be ported to another process and make a unified format across platforms, 
otherwise please let me know if you have an idea, or we could pick a random theme to generate them .. 🤔 doesn't sound quite right.
@brittBarak remember the large words overlapping that is happening in some pages? that's the fix-story for that one ;)

PR #1512, opened 2019-03-06
Add VBC extension endpoint

marklewin:
## Description

There is a new `vbc` endpoint for the connect action which I have documented here. I need to revisit and add links to the VBC Number Programmability docs when they are merged.

### DIFF
--- a/_examples/voice/guides/ncco-reference/connect/vbc-extension-connect.md
+++ b/_examples/voice/guides/ncco-reference/connect/vbc-extension-connect.md
@@ -1,0 +1,23 @@
+---
+title: VBC Extension
+menu_weight: 6
+---
+
+```json
+[
+  {
+    "action": "talk",
+    "voiceName": "Russell",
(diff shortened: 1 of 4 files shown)


marklewin:
Marking as don't merge for now as need to update API ref and also answer the question: If an application routes back to an extension, can you then send DTMF from that extension back to the app? (The VBC extension is a SIP endpoint but DTMF only works for phone endpoints?)

marklewin:
No need to change the API reference. Ready for review/merge.

_examples/voice/guides/ncco-reference/connect/vbc-extension-connect.md
```diff
@@ -0,0 +1,23 @@
+---
+title: VBC Extension
+menu_weight: 6
```
tbedford: Interesting - I'd only seen `navigation_weight` used before.

PR #1513, opened 2019-03-07
Tutorials for making and receiving phone calls with the client sdk and js

AlexLakatos:


_tutorials/client-sdk-js-receive-calls.md
```diff
@@ -0,0 +1,193 @@
...
+```javascript
+.then(application => {
+    notification.textContent = `You've logged in with the user ${application.me.user.name}`;
+    application.on("member:call", (member, call) => {
+      notification.textContent = `You're receiving a call`;
+    })
+})
+```
+
+The listener method takes as a parameter an `member` object that contains information about who's calling, and a `call` object, that lets you interact with the call in progress.
```
garann: This kind of makes it sound like the event handler should be called directly. It might be clearer to change "takes as a parameter" to "receives"?

_tutorials/client-sdk-js-receive-calls.md
```diff
@@ -0,0 +1,193 @@
...
+  notification.textContent = `You're receiving a call`;
+  ...
+  document.getElementById("reject").addEventListener(click, () => {
+    call.reject();
+    notification.textContent = `You rejected the call`;
+  })
+})
+```
+
+In order to reject an incoming call, you'll have to use the `reject()` method on the call object that you received when the `member:call` event was triggered.
```
garann: Should this explanation–and the one for `hangUp`–maybe go before the code, to introduce it a little?

AlexLakatos: I was looking at the iOS ones, that have the explanation under them. I'm fine with either choice, but I'd like to be consistent. I'll change them for now, maybe we should look at updating the other ones as well.

_tutorials/client-sdk-js-receive-calls.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Prerequisites
+
+- Use an existing Nexmo Application or [create a new one](/tutorials/client-sdk-generate-test-credentials).
+- Have a user named `jamie` or [create one](/tutorials/client-sdk-generate-test-credentials#create-a-user) for your Nexmo Application, with a [valid JWT](/tutorials/client-sdk-generate-test-credentials).
+
+
+## Application webhook
+
+For your application to connect an incoming phone call to an app user, you'll need to provide a URL as the `Answer URL` webhook - we've created a [gist](https://gist.github.com/NexmoDev/6534f91245c7bc78ff83e9984c1bffb3) for you to use.
```
brittBarak: can we please use the gist and example Paul created and used on his tutorial? for consistency, as well as potentially to save the end developer work, in case s.he followed the tutorial of another platform.
I'll shortly push my Android tutorial that does the same.

Paul called the user `Jane` and used the gist:
https://gist.github.com/NexmoDev/ed91ac99a0b278fbdcbde72ca3599ac7


Here's the full tutorial: https://developer.nexmo.com/tutorials/client-sdk-ios-inbound-pstn

pardel: FYI, I've used Jane & Joe as per the existing tutorial - not sure where the original naming came from.

AlexLakatos: The JS guides were using jamie and alice. Because we still have the old messaging guides around I'd rather keep using jamie and alice, for consistency. Could I argue for you to make the switch to jamie and alice?

_partials/client-sdk/voice/javascript/login.md
```diff
@@ -0,0 +1,29 @@
+Using the Nexmo Client SDK should start with logging in to `ConversationClient`, using a `jwt` user token.
+
+In production apps, your server would authenticate the user, and would return a [correctly configured JWT](/client-sdk/concepts/jwt-acl) to your app.
+
+For testing and getting started purposes, you can use the Nexmo CLI to [generate JWTs](/tutorials/client-sdk-generate-test-credentials).
+
+We'll update the empty `<script>` tag that's at the bottom of your `<body>` tag to login with a `USER_JWT`, and update the `notification` paragraph when the login was successful. We'll create a new `ConversationClient`, and then call the `login` method, with a string for the user JWT. That returns a promise containig an `application` object, and we'll use that to check we're logged in with the correct user.
```
garann: sp "containing" 

(I'm adding copy edits to these partials so we're not having to edit them for little tweaks every time they're used, hope that's within scope.)

_partials/client-sdk/voice/javascript/login.md
```diff
@@ -0,0 +1,29 @@
...
+  new ConversationClient()
+      .login(USER_JWT)
+      .then(application => {
+          notification.textContent = `You've logged in with the user ${application.me.name}`
+      })
+      .catch(console.log);
+</script>
+```
+
+At this point you should already be able to run the app and see that you can login successfully with the SDK. Because the Nexmo Client uses IndexedDB for it's sync mechanism, we suggest you don't serve `index.html` using the `file://` protocol. That corrupts the IndexedDB for any other users and makes the SDK behave inconsistently. You can use any HTTP server, like `http-server`. If you want to install it from NPM and then run it with cache disabled, here are the terminal commands:
```
garann: "its sync mechanism"

_partials/client-sdk/voice/outbound-pstn-application-webhook.md
```diff
@@ -0,0 +1,13 @@
+For your application to an app user to a phone call, you'll need to provide a URL as the `Answer URL` webhook - we've created a [gist](https://gist.github.com/NexmoDev/f11d10f21fe426ddac4e30b94c6b28dc) for you to fork and update with your Nexmo phone number and personal phone number.
```
garann: This seems like it's missing a verb somewhere: "For your application to an app user to a phone call"

_tutorials/client-sdk-js-make-calls.md
```diff
@@ -0,0 +1,114 @@
...
+```partial
+source: _partials/client-sdk/voice/javascript/login.md
+```
+
+## Make a call
+
+In order to make a call, you'll have to use the `callPhone()` method on the application object that is returned by the `login()` promise. The method accepts a string as parameter that it passes along to your `answer_url`, but since ours is a static file, we'll leave the method parameter empty.
+
+```javascript
+.then(application => {
```
garann: `.then` without the beginning of its chain of functions seems a little confusing, even though I can understand the value of only focusing on the code being discussed. Would it make sense to show the rest of the chain, but use ellipsis for parameter lists and function bodies to reduce the amount of repeated code?

_partials/client-sdk/voice/javascript/hangup.md
```diff
@@ -0,0 +1,10 @@
+```javascript
+application.on("member:call", (member, call) => {
+  notification.textContent = `You're receiving a call`;
+  ...
+  document.getElementById("hangup").addEventListener(click, () => {
```
garann: Should "click" be a string here?

_tutorials/client-sdk-js-receive-calls.md
```diff
@@ -0,0 +1,134 @@
...
+
+
+## Listen for incoming call
+
+When the phone number associated with your Nexmo app receives a call, the application should notify the user `Jane` so that they can decide whether to answer or reject it.
+
+This is done by listening for `member:call` events on the `application` object returned by the `login()` promise.
+
+```javascript
+.then(application => {
```
garann: Same question here about whether there's a way to show the code preceding `.then` without it being too much.

PR #1514, opened 2019-03-07
How to Make Phone Calls with the Nexmo Client SDK in JavaScript

AlexLakatos:


_tutorials/client-sdk-js-make-calls.md
```diff
@@ -0,0 +1,167 @@
...
+``https://gist.githubusercontent.com/NexmoDev/f11d10f21fe426ddac4e30b94c6b28dc/raw/b14c4087097ab67503b4fdd996269502e107bce6/call-phone.json``
+
+and click 'Save changes'.
+
+NB: This gist is specific to this tutorial and in a real-life scenario, the `answer_url` should be provided by a purposely built web solution that can serve custom NCCO's if required.
+
+
+## Adding the Nexmo Client to your application
+
+We'll start with a blank HTML page with two buttons for calling a phone and hanging up a call. Create an `index.html` file with, and initialize it with a boilerplate, for example:
```
garann: I think this is the same as in the other PR, but the sentence "Create an `index.html` file with, and initialize it with a boilerplate, for example:" doesn't make sense to me. Is there another way to word that?

_tutorials/client-sdk-js-make-calls.md
```diff
@@ -0,0 +1,167 @@
...
+      })
+      .catch(console.log);
+</script>
+```
+
+At this point you should already be able to run the app and see that you can login successfully with the SDK. Because the Nexmo Client uses IndexedDB for it's sync mechanism, we suggest you don't serve `index.html` using the `file://` protocol. That corrupts the IndexedDB for any other users and makes the SDK behave inconsistently. You can use any HTTP server, like `http-server`:
+
+```bash
+$ npm install -g http-server
+$ http-server -c-1
```
garann: As in the other PR, it seems nice to explain what this is doing.

AlexLakatos:
@garann thanks for the review. Because most of it is common with the other tutorial, I'll merge the 2 branches and use partials for the common bits.

PR #1520, opened 2019-03-11
Remove unused #search_error action and view

timcraft:
Unused since 14e12324


timcraft:
Agreed, welcome! For some reason the status check was stuck in pending on this one even though the tests were passing (ditto #1523), so I've just rebased and pushed it up again.

PR #1523, opened 2019-03-11
Combine CodeLanguageResolver with CodeLanguage

timcraft:
Merges CodeLanguageResolver methods into the CodeLanguage model (because it doesn't really add anything besides indirection). Removes a couple of related DocumentationConstraint methods.


spec/models/code_language_spec.rb
```diff
@@ -43,4 +43,89 @@
...
+
+  describe '.find' do
+    it 'returns the code language for the given key' do
+      expect(CodeLanguage.find('ruby')).to be_kind_of(CodeLanguage)
+      expect(CodeLanguage.find('ios')).to be_kind_of(CodeLanguage)
+      expect(CodeLanguage.find('curl')).to be_kind_of(CodeLanguage)
+    end
+
+    context 'when given an invalid key' do
+      it 'returns nil' do
```
mheap: The description doesn't seem to match the behaviour here

spec/models/code_language_spec.rb
```diff
@@ -43,4 +43,89 @@
...
+
+  describe '.linkable' do
+    let(:languages) { CodeLanguage.linkable }
+
+    it 'returns an array of code languages' do
+      expect(languages).to be_kind_of(Array)
+      expect(languages[0]).to be_kind_of(CodeLanguage)
+    end
+
+    it 'excludes code languages that are not linkable' do
```
mheap: It'd be good to assert that something that we know isn't linkable isn't in `languages` here

spec/models/code_language_spec.rb
timcraft: Yep, good catch 🙈

spec/models/code_language_spec.rb
timcraft: Added a couple of assertions.

PR #1537, opened 2019-03-14
Code snippets for Server SDKs become submodules

lornajane:
## Description

We relocated and renamed all our "building blocks" repositories - they are now "code snippets" repositories and they live in the main `nexmo` org. This PR brings these repos in as submodules and updates all references to their previous paths (I hope).

## Deploy Notes

I don't _think_ new submodules will impact deployment but a review app would be a good move for sure


benhayehudi:
@lornajane - Are we also renaming the category to "Code Snippets" from "Building Blocks"? Right now, in this review app the top-level links under each product still identify as Building Blocks.

benhayehudi:
Additionally, should we rename the methods in NDP currently called `BuildingBlocksFilter` to `CodeSnippetsFilter` in this PR as well?

lornajane:
I suggest we get the copy changes added in a later PR - this one is kind of bigger than I'd like as it is and the two steps are quite separate. This one just does "plumbing", doesn't change much on NDP. I'll open tickets for the next steps and would appreciate input

benhayehudi:
> I suggest we get the copy changes added in a later PR - this one is kind of bigger than I'd like as it is and the two steps are quite separate. This one just does "plumbing", doesn't change much on NDP. I'll open tickets for the next steps and would appreciate input

That makes sense. I can work on the two comments I raised once this is merged.

lornajane:
The curl ones are not included because you asked me not to include them. Since the team has clicked around and tested this changeset, let's only merge these changes for now. I'll raise another PR to bring the curl repos in line and that can be reviewed and tested separately

marklewin:
I have spoken to @lornajane about this and the plan is to implement this change without the curl repo, because this PR has been widely tested by the team. So, changing my review to approved.

PR #1438, opened 2019-02-11
WIP - Client SDK getting started/setup section mockup

tbedford:
## Description

**ABANDONED** -- DO NOT DELETE

* The Ruby code changes / Task related configs have been salvaged in PR #1538.
* TBD

----
A test restructuring that moves the set up material from tutorials to its own Getting Started section.

- [x] Moved setup material to Getting Started section
- [x] Used tabbed content for Getting Started material.
- [x] Moved other task-oriented material to Guides.

Update:

- [x] Add task-based config
- [ ] Add landing pages

## Review 

A Review App is available via Heroku.

PR #1538, opened 2019-03-14
Task related fixes salvaged from PR1438.

tbedford:
## Description

Ruby code related to Tasks salvaged from PR #1438.

Includes some Rubocop fixes.


tbedford:
@benhayehudi : Michael made some amendments/additions to Task related Ruby code just before his break - these were in PR1438 which has to be abandoned for various reasons. Would it be possible to have a quick check through please in Michael's absence? Thanks!

PR #1547, opened 2019-03-15
Document VBC extension endpoint

marklewin:
## Description

Added new `vbc` endpoint to `/api/voice#createCall`

## Deploy Notes

View at https://nexmo-developer-pr-1547.herokuapp.com/api/voice/#createCall


marklewin:
Removing this as Vonage don't want this to go live before Monday 18th March 2019 because of an exclusive with TechCrunch

PR #1556, opened 2019-03-18
Updated node-quickstart for package vulnerability

benhayehudi:
## Description

Updated the version of `extend` to `3.0.2` in `package.json`.


lornajane:
We'll need to re-open this after the change to submodules. @benhayehudi I'll leave this with you if that is OK?

PR #1566, opened 2019-03-19
Fix incorrect formatting of array elements as labels

marklewin:
## Description

This is an attempt to fix the issue shown [here](https://files.slack.com/files-pri/T02NNHD8S-FH0UAHUD6/send-sms-python-bad-label.png) (in the as-yet uncommitted [PR 1565](https://github.com/Nexmo/nexmo-developer/pull/1565)), where array elements are being rendered as labels. I'm just checking for the presence of a whitespace character before the square bracket before it is considered to be a label.

## Deploy Notes

I've checked for `[POST]`, `[GET}`, `[200]` etc and all appears to be okay, but till need further testing in the [review app](https://nexmo-developer-pr-1566.herokuapp.com/).

### DIFF
--- a/app/filters/label_filter.rb
+++ b/app/filters/label_filter.rb
@@ -1,7 +1,7 @@
 class LabelFilter < Banzai::Filter
   def call(input)
     return input if options[:disable_label_filter]
-    input.gsub(/\[([a-zA-Z0-9\s:\-\.]+)\]/) do |_s|
+    input.gsub(/\s\[([a-zA-Z0-9\s:\-\.]+)\]/) do |_s|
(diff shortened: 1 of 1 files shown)


marklewin:
Marking as WIP for the time being as needs more thought :/

marklewin:
My first idea as above was to look for a leading space. This, however, broke the deploy tests on `LabelFilter`.

The approach in my second commit is this: consider a label a label only when it is longer than 1 char. I couldn't find any instances of labels shorter than that and I can't find any example of referencing array elements in positions > 9 either.

marklewin:
Removing due to discussions around doing away with the labels filter which seems to cause more problems than it solves.

PR #1570, opened 2019-03-20
[Take 2] Migrated all references on NDP from building blocks to code snippets

lornajane:
## Description

Possible successor to #1564 : this has been rebased to take account of the moved/renamed curl code snippets. It aims to transform all use of the phrase "building block" on NDP to "code snippet"


marklewin:
Here we are capitalising `Code Snippet`, whereas before we had `building block`. I personally prefer lower case for this, but let's see what @tbedford and @lornajane say.

benhayehudi:
> Here we are capitalising `Code Snippet`, whereas before we had `building block`. I personally prefer lower case for this, but let's see what @tbedford and @lornajane say.

Can you reference where is the "here"? On the main NDP page it's called "Building Blocks" and throughout the site it's capitalized in some places, while not in others. Do we have a style guide that tells us when to vs. not to? Thanks @marklewin!

lornajane:
Yeah I saw the capitalisation and I think it's incorrect in some places, particularly in prose where we previously had "building blocks" and now we have "Code Snippets" but I don't think the snippets are a proper noun!

lornajane:
Also: If we decide we need a change, I can probably help to fix it up quite quickly so let's not be reluctant to change it because work would be involved. I can think of a few shortcuts.

benhayehudi:
In a post-modern world, what truly is a "proper noun" in any case? 😆 

On a serious note, I would recommend approving it as such, and I can open a separate PR to go back through the prose and de(?)-capitalize the places it should not be. I'd rather make sure nothing is broken here, and then make that fix after.

lornajane:
I'd like to get this correct and then apply it in a single change. The "Code Snippet" is a bit jarring and I think it is worth getting it right. I think we can test as-is because I'm only proposing the case change in prose context so the lion's share of the work won't need a change and in theory this last bit of polish shouldn't be functional so won't break things ...

benhayehudi:
> I'd like to get this correct and then apply it in a single change. The "Code Snippet" is a bit jarring and I think it is worth getting it right. I think we can test as-is because I'm only proposing the case change in prose context so the lion's share of the work won't need a change and in theory this last bit of polish shouldn't be functional so won't break things ...

I'm okay with that, especially if you can manage it in a relatively painless shortcut of a way.

marklewin:
Yes, basically in the `.md` files that describe the code snippets. For example: https://github.com/Nexmo/nexmo-developer/pull/1570/files#diff-0b63035501f81f4f1748698b4f321e1e

As Lorna says, "code snippets" isn't a proper noun, so I'd be happier if it wasn't capitalised :)

> Can you reference where is the "here"? On the main NDP page it's called "Building Blocks" and throughout the site it's capitalized in some places, while not in others. Do we have a style guide that tells us when to vs. not to? Thanks @marklewin!

lornajane:
Try it now? Although it was a bit of a blunt `sed` instrument and only about 20 hand edits so do check it makes sense ....

PR #1578, opened 2019-03-21
Redirect issue from 'building-blocks' to 'code-snippets' solution

benhayehudi:
Check in controller if `building-blocks` is present in the URL parameters, and, if so, replace it with `code-snippets`.

### DIFF
--- a/app/controllers/markdown_controller.rb
+++ b/app/controllers/markdown_controller.rb
@@ -5,6 +5,9 @@ class MarkdownController < ApplicationController
   before_action :set_namespace
 
   def show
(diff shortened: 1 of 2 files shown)


benhayehudi:
As discussed with @tbedford, this fix is temporary. Tomorrow morning I will add the relevant `building-blocks/` URLs into a YAML file and add that to the Redirector service functionality. This will mean that this condition check in this PR can be safely removed at that point and the functionality of the existing service will redirect all the old `building-blocks/` to `code-snippets/`

PR #1582, opened 2019-03-22
Verify request limit one per second

marklewin:
## Description

Added note that verify requests are limited to a maximum of one per second to:

- "Verification Events" guide
- "Send Verification" code snippet
- OAS "Verify Request" endpoint description under [PR 90](https://github.com/Nexmo/api-specification/pull/90)


lornajane:
I'd strongly recommend against opening a pull request that points to anything other than the tip of a master branch inside a submodule, because for example if I review and merge this, I'd remove a load of changes that are in master but not in this branch.

It is totally fine and actually useful to do it for testing BUT in that case please clearly mark the PR as "Do Not Merge" and write us a note for why? I typically prefer to get the change in to the submodule before opening the PR on the parent repo to try to avoid incidents.

marklewin:
Good point.

**Note**: This PR references a branch for the OAS submodule repo. This is so that we can create a test app. Adding **don't merge** label.

> I'd strongly recommend against opening a pull request that points to anything other than the tip of a master branch inside a submodule, because for example if I review and merge this, I'd remove a load of changes that are in master but not in this branch.
> 
> It is totally fine and actually useful to do it for testing BUT in that case please clearly mark the PR as "Do Not Merge" and write us a note for why? I typically prefer to get the change in to the submodule before opening the PR on the parent repo to try to avoid incidents.

marklewin:
Closing because of unwanted commits.

PR #1586, opened 2019-03-22
Mention POST method in NI API ref but use GET in curl snippets

marklewin:
## Description

This PR brings in the change to the NI OAS in https://github.com/Nexmo/api-specification/pull/88 (mention that `POST` requests are also accepted) and the `curl` code snippets (use `GET` instead of `POST` so we don't advertise that we accept `POST` too).


marklewin:
Test app: https://nexmo-developer-pr-1586.herokuapp.com

Check that all NI curl snippets use `GET` rather than `POST` and that each endpoint in the API ref says that `POST` is also accepted.

PR #1591, opened 2019-03-25
Fix Account/Secret Management redirect

mheap:
## Description

https://developer.nexmo.com/api/account/secret-management currently throws an error as `secret-management` is detected as a code language. This PR makes the route stricter by only accepting code languages that we know about

This means that the `code_language` filter succeeds, allowing NDP to get to the redirect filter and redirect as intended

## Deploy Notes

N/A


lornajane:
The secret management spec shouldn't be there any more, it was combined into the account one. So I'm not sure what will happen with this? The URL should redirect to account now I think

mheap:
It should redirect, but it doesn't due to a bug. This PR fixes that bug

PR #1594, opened 2019-03-26
Create Nexmo application with VBC capability

marklewin:
## Description

Replaces the existing description of creating a Nexmo Application for VBC Number Programmability. You must enable the VBC capability for number programmability calls to be zero-rated in Nexmo. This is only available in version 2 of the Application API, which the Nexmo CLI and Dashboard do not yet use.


marklewin:
Adding don't merge, because I want to include the updated Application API v2 OAS. Would appreciate a review before I do this.

marklewin:
Removed "don't merge" and adding "ready for review" because I have submitted the OAS changes under a separate PR: https://github.com/Nexmo/nexmo-developer/pull/1595

PR #1595, opened 2019-03-26
Update Applications v2 OAS with VBC capability

marklewin:
## Description

This brings in the OAS changes in the following PR: https://github.com/Nexmo/api-specification/pull/94


marklewin:
Reopening for merge of Account API OAS change from submodule. Removed incorrect commits from earlier.

PR #1596, opened 2019-03-26
Update ncco-reference.md

sammachin:
Updated the description for conversation names as they are (now) only scoped to an application level not the account level


sammachin:
FYI I've confirmed this is correct with testing and talking to Neil

PR #1597, opened 2019-03-26
Update verify OAS with one per second limit

marklewin:
## Description

Updated Verify OAS with one per second limit. That was merged in https://github.com/Nexmo/api-specification/pull/90. Bringing that change into NDP.


mheap:
I've updated to latest master which pulled this in too

PR #1602, opened 2019-03-27
NCCO connect guide | updated terminology for app

mariatechmaniac:
we should not use the term 'ip' externally. the leg type: app (both in the code and in the docs). 
no need to mention 'legs' here as well. we connect TO users VIA legs(legs is the means). 

can either use: application or app
i used app here so that people dont get confudes with the concept of Nexmo application :/

we connect to a denoted user, not a member. the member is created as a result of this ncco. the vapi user is also unaware of the concept of the member so lets not add confusion ;)


mariatechmaniac:
@brittBarak 
sorry mate wont let me add you as a reviewer =/

PR #1606, opened 2019-03-28
Add new Advanced NI IP matching guide

marklewin:
## Description

Added a new guide showing developers how to IP match in the Number Insight Advanced API.

I've added a note about the `ip_warnings` response property after discussion with James Wren:

> ...this field is pretty much defunct, as we’re always returning unknown. unknown means “we don’t know whether the there are warnings for this IP being an anonymous proxy / satellite provider, or whether there are no warnings”.

If reviewers agree it is OK to mention this here, I'll submit another PR to add the same to the `ip_warnings` response field description in the API reference.


mheap:
👍 for adding the warning to the API reference too

PR #1616, opened 2019-04-03
Use new Numbers OAS

marklewin:
## Description

Use new Numbers OAS. Also changed link on docs homepage to point to the new API reference.

Please test in [review app](https://nexmo-developer-pr-1616.herokuapp.com/).


app/views/static/_products.html.erb
```diff
@@ -211,7 +211,7 @@
       <nav>
         <h3 class="Vlt-grey-dark">Developer API</h3>
         <a href="/api/developer/messages">Messages</a> |
-        <a href="/api/developer/numbers">Numbers</a> |
```
mheap: Can we add a redirect for this?

marklewin: Yes, done.

mheap:
This PR looks good, but there are a couple of things to change in the OAS to ensure that we don't lose information before merging this:

### General
Keep `api_key` and `secret` in URL but move other fields e.g. `country` and `msisdn` to body

### Search Available
type: The type of number to search for. Accepted values: landline, landline-toll-free and mobile-lvn.	

### Update a number

messagesCallbackType and messagesCallbackValue are missing. These may be missing from the list endpoint response too

marklewin:
> Keep api_key and secret in URL but move other fields e.g. country and msisdn to body

You can only include the API key and secret in the path for the `/account/numbers` endpoint. Changed it there.

PR #1626, opened 2019-04-12
Send and Receive DTMF for iOS

pardel:
## Description

Added iOS Swift & Objective-C snippets for sending and receiving DTMF using the Client SDK.

## Deploy Notes

No changes.

### DIFF
--- a/_documentation/client-sdk/in-app-voice/guides/send-and-receive-dtmf.md
+++ b/_documentation/client-sdk/in-app-voice/guides/send-and-receive-dtmf.md
@@ -5,8 +5,6 @@ description: This topic shows you how to send and receive DTMF in your app.
 
 # Send and Receive DTMF
 
-> **NOTE**: This feature is currently supported only on Nexmo Client Android SDK.
-
 ## Overview
 
 Dual Tone Multi Frequency (DTMF), is a form of signalling used by phone systems to transmit the digits `0`-`9` and the `*` and `#` characters. Typically a caller presses these buttons on their telephone keypad and the phone then generates a tone made up of two frequencies played simultaneously (hence Dual Tone).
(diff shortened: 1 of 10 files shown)


mheap:
@brittBarak I renamed Java to Kotlin and added `ObjC` highlighting.

@pardel Use `objective_c` as the language to get highlighting on code blocks

pardel:
@brittBarak could we push this live?

brittBarak:
1. I think we should move the files to another location, since it's not a part of get-started at all. I created in my PR a "guides" folder, we can move them there. 

2. I think, for consistency in NDP, we should rename the files to lower case. so `Swift.md` --> `swift.md` etc.

3. I don't really like the inconsistency between the snippets tbh. I know it's only the params, and quite small. I personally like clean samples and small samples when it comes to official docs. either way, I strongly feel we should keep the platforms snippets as consistent as possible. 

4. are we sure the name of the callback is `DTMFreceived()`? was supposed to be `onDTMF()` and consistent with Android.

pardel:
@brittBarak 
re: 1 - do you mean the subfolder at:  _tutorials_tabbed_content/client-sdk/guides ?
re: 2 - 👍
re: 3 - do you mean the method params? Those are coming from the SDK.
re: 4 - as per the Slack chat, here is the NXMCallDelegate protocol definition:

```
@protocol NXMCallDelegate <NSObject>
- (void)statusChanged:(nonnull NXMCallMember *)callMember;
@optional
- (void)DTMFReceived:(nonnull NSString *)dtmf callMember:(nonnull NXMCallMember *)callMember;
@end
```

brittBarak:
re: 1 -  _tutorials_tabbed_content/client-sdk/guides  --> yes. I merged mine so you can pull and use that. 
re: 3 - no, I meant the dtmf string example. I'd love all of us to have a clean consistent implementation. 
re: 4 yes. following slack: 🤦‍♀️😭

ok I'm approving this PR, as I don't want to be a blocker. if you can change the folder and merge, or do another PR, as you wish. 

thank you!!!

mheap:
@pardel What do we need to do to get this merged?

pardel:
@mheap there were some nice-to-have changes but I'll create another PR

PR #1627, opened 2019-04-12
Fix OAS array response field rendering

mheap:
## Description

If it's an array, look at the items for the response schema as all items within the array are expected to be identical

## Deploy Notes

N/A

### DIFF
--- a/app/views/open_api/_response_description_parameters.html.erb
+++ b/app/views/open_api/_response_description_parameters.html.erb
@@ -6,6 +6,12 @@
   end
 %>
 
+<%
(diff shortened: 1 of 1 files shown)


brittBarak:
@mheap I assume that it's for viewing the response fields. I don't really have an idea on how OAS works, in order to approve / disapprove it 🤔

mheap:
@brittBarak Yep! You should be able to inspect `List Conversations for Members` in the `Conversation` OAS to see that it's displaying fields (whilst current production isn't)

PR #1637, opened 2019-04-16
Add Framework Libraries to NDP Tools Page

benhayehudi:
## Description

Add Nexmo framework libraries to the `/tools` page on NDP

### DIFF
--- a/app/views/static/tools.html.erb
+++ b/app/views/static/tools.html.erb
@@ -170,6 +170,72 @@
 <hr>
 <br>
 
+
+
+<h2>Framework Libraries</h2>
+
+<p>The Nexmo framework libraries can get you up and running with the Nexmo APIs quickly and easily in your framework of choice.</p>
+
+<div class="Vlt-grid Vlt-grid--margin-top3">
+  <div class="Vlt-col Vlt-col--1of3">
+    <a href="https://github.com/Nexmo/nexmo-laravel" class="Vlt-card Nxd-github-card" data-github="Nexmo/nexmo-laravel">
+      <h3 class="Vlt-blue-dark">nexmo-laravel</h3>
+      <p>Nexmo framework library for Laravel</p>
+      <img src="//badge.fury.io/rb/nexmo.svg" class="Nxd-github-card__badge" height="18">
+      <p>
+        <small class="Nxd-github-card__meta">
+          <span>
+            <span class="Vlt-red-dark">●</span> PHP
+          </span>
+          <span>
+            <svg class="Vlt-icon Vlt-icon--smaller Vlt-icon--text-bottom Vlt-grey-darker" aria-hidden="true"><use xlink:href="/symbol/volta-icons.svg#Vlt-icon-star-full"/></svg>
+            <span data-stars>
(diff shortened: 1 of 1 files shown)


app/views/static/tools.html.erb
```diff
@@ -170,6 +170,72 @@
...
+          </span>
+        </small>
+      </p>
+    </a>
+  </div>
+
+  <div class="Vlt-col Vlt-col--1of3">
+    <a href="https://github.com/Nexmo/nexmo-rails" class="Vlt-card Nxd-github-card" data-github="Nexmo/nexmo-rails">
+      <h3 class="Vlt-blue-dark">nexmo-rails</h3>
+      <p>Nexmo framework library for Ruby on Rails</p>
```
mheap: Let's add this the day of Railsconf so that we can shout about it as being brand new

benhayehudi: 👍 Makes sense to me!

lornajane:
Can we keep the CLI above the frameworks? I think it's useful to a wider range of developers.

Also if the rails lib is ready then I'd prefer to see it listed, we don't have a way of timing content release and it could be getting indexed by things for a week. If we aren't doing anything else before release then I don't know why we'd keep this cool thing a secret from anyone who cared to go looking (sorry, maybe I'm missing an important point here)

benhayehudi:
> Can we keep the CLI above the frameworks? I think it's useful to a wider range of developers.
> 
> Also if the rails lib is ready then I'd prefer to see it listed, we don't have a way of timing content release and it could be getting indexed by things for a week. If we aren't doing anything else before release then I don't know why we'd keep this cool thing a secret from anyone who cared to go looking (sorry, maybe I'm missing an important point here)

I'm happy to move the CLI above the frameworks, I don't feel strongly either way.

Re: second point -- I'll let @mheap comment on timing, etc.

mheap:
We're going to release, add to NDP and publish a blog post on the first day of RailsConf to try and capitalise on the social going on around the event. Seeing that it was published X hours ago is more likely to draw people in than seeing "Published 15 days ago" and we're just shouting about it now imo

lornajane:
Definitely not my call, forget I mentioned it!

app/views/static/tools.html.erb
```diff
@@ -198,8 +198,47 @@
...
+<h2>Framework Libraries</h2>
+
+<p>The Nexmo framework libraries can get you up and running with the Nexmo APIs quickly and easily in your framework of choice.</p>
+
+<div class="Vlt-grid Vlt-grid--margin-top3">
+  <div class="Vlt-col Vlt-col--1of3">
+    <a href="https://github.com/Nexmo/nexmo-laravel" class="Vlt-card Nxd-github-card" data-github="Nexmo/nexmo-laravel">
+      <h3 class="Vlt-blue-dark">nexmo-laravel</h3>
+      <p>Nexmo framework library for Laravel</p>
+      <img src="https://poser.pugx.org/nexmo/laravel/v/stable" class="Nxd-github-card__badge" height="18">
```
mheap: Why did we choose to use `pugx` rather than `https://badge.fury.io/ph/nexmo%2Flaravel.svg` like all of the other cards?

benhayehudi: Because that's what is reference in `nexmo-laravel`'s README: https://raw.githubusercontent.com/Nexmo/nexmo-laravel/master/README.md

mheap: Ah, I think that was a hacktoberfest contribution. I've switched to using a `fury` badge for consistency and will merge

PR #1643, opened 2019-04-16
Conversation API OAS Fixes

brittBarak:
## Description

## Deploy Notes


mheap:
@brittBarak I've resolved the issue you were having, but I'm not sure that this is correct. Can `from` really be an array of items? The text also needs work as the descriptions are all in the context of placing a call

brittBarak:
@mheap  thank you!

1. I still can't seem to see a review app... :-/ 

2. sorry so maybe I didn't understand you in Slack: `to ` and `from` have a `type`, that can one of "app, phone, sip, websocket, vbc". for each type there's a second field to fit. for example if `type = app`  there will be a field `user=username`. if `type = sip` the other field is `uri=...`. Does that make sense? is there another way to represent it?

3. it is indeed the context of placing or receiving a call. happy for phrasing suggestions. 

4. I'd need to make a small change in `conversation.yaml` line 657: user_name --> user, I got a little lost. but it can wait for a bit.


Thank you for the help!!

mheap:
> I still can't seem to build a review app... :-/

It's building for me now

> sorry so maybe I didn't understand you in Slack: to and from have a type, that can one of "app, phone, sip, websocket, vbc". for each type there's a second field to fit. for example if type = app there will be a field user=username. if type = sip the other field is uri=.... Does that make sense? is there another way to represent it?

The way we have it now using `oneOf` is the only way. The styling needs work as this is the most complex API we've encountered so far, but it is showing accurately

> it is indeed the context of placing or receiving a call. happy for phrasing suggestions.

I'll add a followup ticket for this as it's not a quick task

> I'd need to make a small change in conversation.yaml line 657: user_name --> user, I got a little lost. but it can wait for a bit.

Looks like you already got to this!

PR #1647, opened 2019-04-17
Add JWT dependency to curl building blocks

marklewin:
## Description

The curl voice code snippets assumed knowledge of how to generate a JWT for authentication but we don't ever show how to do this.

After discussion with Michael we agreed that the best approach would be to add a `JWT` dependency to curl code snippets.

If this dependency is present, we also change the title  of the `_dependencies.html.erb` view from "Install dependencies" to "Generate your JWT"

I have added the JWT dependency to the Voice API "Make an outbound call" curl code snippet for testing.

```
title: cURL
language: curl
dependencies:
  - JWT
code:
    source: .repos/nexmo/nexmo-curl-code-snippets/voice/voice-api/make-an-outbound-call.sh
    from_line: 5
    to_line: 10
file_name: make-an-outbound-call.sh
run_command: bash make-an-outbound-call.sh
```

### DIFF
--- a/app/services/code_snippet_renderer/curl.rb
+++ b/app/services/code_snippet_renderer/curl.rb
@@ -6,7 +6,7 @@ module CodeSnippetRenderer
       {
         'text' => 'Run the following <code>curl</code> command to create the <a href="/concepts/guides/authentication#json-web-tokens-jwt">JWT</a> for authentication:',
         'code' => 'export JWT=\'$(nexmo jwt:generate $PATH_TO_PRIVATE_KEY application_id=$NEXMO_APPLICATION_ID)\'',
-      } if dependencies.include?('JWT')    
+      }    
     end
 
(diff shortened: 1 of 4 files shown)


app/services/code_snippet_renderer/curl.rb
```diff
@@ -1,7 +1,12 @@
 module CodeSnippetRenderer
   class Curl
-    def self.dependencies(_deps)
-      raise 'No dependency support for Curl'
+    def self.dependencies(deps)
+      dependencies = deps.map(&:upcase)
+      raise 'The only permitted curl dependency is `jwt`' unless dependencies.include?('JWT')
+      {
+        'text' => 'Run the following <code>curl</code> command to create the <a href="/concepts/guides/authentication#json-web-tokens-jwt">JWT</a> for authentication:',
```
mheap: This is a `shell` command, not a `curl` command

PR #1648, opened 2019-04-17
Add inbound SIP answer_url docs

mheap:
## Description

We've added the capability to forward SIP headers to an `answer_url` if they are provided in the SIP invite.

## Deploy Notes

N/A


mheap:
Deniz has reviewed. We'll update in the future but for now this gets something out there

PR #1651, opened 2019-04-17
Add Ringback key to onAnswer in NCCO ref

tbedford:
## Description

Adds mention of `ringback` value in `onAnswer`.

I have some questions about `onAnswer`. What exactly is the `url` for - the URL in the example I saw looks more like an event webhook.  

We also probably need a building block on this.


tbedford:
@mheap reworded - hopefully clearer?

mheap:
Much! Cheers @tbedford

PR #1652, opened 2019-04-18
Upgrade webpack to v4

fabianrbz:
## Description

[Upgrade to Webpack 4](https://nexmoinc.atlassian.net/browse/DEVX-1030).
* dependencies and config files were updated
* config files were replaced with their new versions
* files that were not entries were removed from `packs` to follow the conventions and react components were placed under `components`

## Deploy Notes


fabianrbz:
@mheap I took the liberty to re-arrange some files, we can change the structure though.
One thing that would be nice to have would be to include `RAILS_ENV=test bundle exec rake webpacker:compile` to travis so at least we know that we didn't brake the build when we merge stuff. WDYT?

mheap:
👍 on adding a compilation test to Travis.

Both the asset pipeline and `bin/webpack-dev-server` both work for me locally, as does `/jwt` and code in `application.js`.

I noticed that if you use `webpack-dev-server` and then turn it off you need to run `assets:clobber` to force Rails to recompile all the assets. This has an unwanted side effect of deleting lots of files (which are currently in the wrong place). 

I'm happy to merge this once we add the test to Travis and revisit the `clobber` issue in a future PR

fabianrbz:
👍 I'll add that to travis and take a look at the `assets:clobber` issue.

PR #1653, opened 2019-04-18
Clarify when completed status returned by a call

marklewin:
## Description

This change is in response to the following feedback received by @fauna5:

> "completed" is always returned unless the call was rejected. "unanswered", "timeout", "failed", "busy" and "cancelled" all proceed to "completed". The description of the "completed" state probably needs to change to something like "When a call flow has been terminated. This is always received unless the status is 'rejected'"

### DIFF
--- a/_documentation/voice/voice-api/guides/call-flow.md
+++ b/_documentation/voice/voice-api/guides/call-flow.md
@@ -24,10 +24,10 @@ You may choose to provide your NCCO as part of the request you send to create a 
 
 Each call goes through a sequence of states in its lifecycle:
 
-1. Created
-2. Ringing
-3. Answered
-4. Completed
+1. `started`
+2. `ringing`
+3. `answered`
+4. `completed`
 
 ## Event objects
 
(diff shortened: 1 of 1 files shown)


_documentation/voice/voice-api/guides/call-flow.md
```diff
@@ -27,7 +27,7 @@ Each call goes through a sequence of states in its lifecycle:
 1. Created
```
fauna5:
```suggestion
1. Started
```

marklewin: I've reformatted the sequence of events to match the actual event names. I've also added a note saying that the events usually arrive in order but that there are no guarantees and their applications should handle out-of-order messages.

marklewin:
Marking as don't merge while @fauna5 is looking at the timestamps of some out-of-order events he has experienced.

fauna5:
Unfortunately not :( 

CON-0caa8d24-254f-4a33-9bc4-1f4ddb97b220
   status           payload timestamp
4. completed 	2019-04-16T16:24:30.576Z
3. ringing    	2019-04-16T16:24:25.753Z
2. started    	2019-04-16T16:24:25.753Z
1. busy   		2019-04-16T16:24:30.613Z

marklewin:
OK, so are we saying that there is no way to order the events? In which case I'd better remove the suggestion that their application handles it, @mheap?

fauna5:
Best to leave it out i think

mheap:
Which timestamp is out of order? `busy`/`completed`?

fauna5:
Yep

mheap:
I've checked with @JamieChapman8462 and neither order or generation nor order of delivery are guaranteed.

Let's take Rich's suggestion and omit the ordering from the docs

PR #1654, opened 2019-04-18
Check for migration pending Rails error in CI rake tasks

benhayehudi:
… into rake tasks

## Description

Up to this point, the CI Rake tasks do not provide feedback if it fails because migrations are pending. This PR adds a new class for `CommonErrors` and adds a class method in it to check for migration pending errors. That method is then incorporated in `./lib/ci.rake` tasks.

### DIFF
--- a/lib/common_errors.rb
+++ b/lib/common_errors.rb
@@ -1,5 +1,5 @@
 class CommonErrors
   def self.check_for_migration_error(body)
-    raise 'DB Migration is pending. Please run RAILS_ENV=development bin/rake db:migrate' if body && body.downcase.include?('migrations are pending')
+    raise 'DB Migration is pending. Please run RAILS_ENV=development bin/rake db:migrate' if body&.downcase&.include?('migrations are pending')
   end
 end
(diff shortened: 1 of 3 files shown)


lib/common_errors.rb
```diff
@@ -0,0 +1,5 @@
+class CommonErrors
+  def self.check_for_migration_error(session)
+    errors.push('DB Migration is pending. Please run rake db:migrate RAILS_ENV=development.') if session.response.body.include?('Migrations are pending.')
```
mheap: This should `raise` rather than push in to `errors` if an error is encountered

benhayehudi:
Could use some help in simulating the session for the spec tests. I've pushed up what I got so far... 🙏

mheap:
@benhayehudi Instead of passing `session` pass `session.response.body`. At that point your `CommonErrors` class is dealing with strings and is easy to test.

The general rule here is that you should *tell*, not *ask* when passing function parameters. The fancy name is the law of Demeter (https://hackernoon.com/object-oriented-tricks-2-law-of-demeter-4ecc9becad85)

benhayehudi:
> @benhayehudi Instead of passing `session` pass `session.response.body`. At that point your `CommonErrors` class is dealing with strings and is easy to test.
> 
> The general rule here is that you should _tell_, not _ask_ when passing function parameters. The fancy name is the law of Demeter (https://hackernoon.com/object-oriented-tricks-2-law-of-demeter-4ecc9becad85)

This might be the most helpful and educational comment on a PR in a while. Thanks @mheap!

PR #1669, opened 2019-04-29
Add code examples to message signing docs

lornajane:
## Description

The PHP lib now has signature generation and verification functionality and we have building block spec for both of these - so let's add the code to the docs.

### DIFF
--- a/_documentation/concepts/guides/signing-messages.md
+++ b/_documentation/concepts/guides/signing-messages.md
@@ -14,6 +14,14 @@ You use a signature to:
 * Ensure that the message has not been tampered with en-route
 * Defend against interception and later replay
 
+## Contents
+
+This document covers how to use signatures with messages.
+
+  - [Send messages with signatures](#use-signatures-when-sending-messages)
+  - [Verify signatures on incoming messages](#validate-the-signature-on-incoming-messages) to ensure authenticity of the incoming webhook
+  - [Manually generate a signature](#manually-generate-a-signature) if you cannot use the existing libraries, the manual process for signature generation is included here
+
 ## Use signatures when sending messages
 
 To send a message with a signature, you will need to use the `SIGNATURE_SECRET` instead of your `API_SECRET` when sending the message. You can find the signature secret and choose which signing algorithm to use by visiting the [dashboard](https://dashboard.nexmo.com). The default algorithm is'[MD5 hash](https://en.wikipedia.org/wiki/MD5)' and we also support `MD5 HMAC`, `SHA1 HMAC`, `SHA-256 HMAC` and `SHA-512 HMAC`.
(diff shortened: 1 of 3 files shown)


_documentation/concepts/guides/signing-messages.md
```diff
@@ -14,6 +14,14 @@ You use a signature to:
 * Ensure that the message has not been tampered with en-route
 * Defend against interception and later replay
 
+## Contents
+
+This document covers how to use signatures with messages.
+
+  - [Send messages with signatures](#use-signatures-when-sending-messages)
+  - [Verify signatures on incoming messages](#validate-the-signature-on-incoming-messages) to ensure authenticity of the incoming webhook
+  - [Manually generate a signature](#manually-generate-a-signature) if you cannot use the existing libraries, the manual process for signature generation is included here
```
marklewin: In the third bullet, we're talking about client libs before we've defined them. Maybe change the first bullet to read: 
- [Send messages with signatures](#use-signatures-when-sending-messages) Use the [client libraries](/tools) to generate and send the signed message

_documentation/concepts/guides/signing-messages.md
```diff
@@ -14,6 +14,14 @@ You use a signature to:
 * Ensure that the message has not been tampered with en-route
 * Defend against interception and later replay
 
+## Contents
+
+This document covers how to use signatures with messages.
+
+  - [Send messages with signatures](#use-signatures-when-sending-messages)
+  - [Verify signatures on incoming messages](#validate-the-signature-on-incoming-messages) to ensure authenticity of the incoming webhook
```
marklewin: Are we ensuring the authenticity of the webhook itself, or the incoming request on the webhook? I believe the latter. This also applies to the description under the "Validate the signature" section below.

_documentation/concepts/guides/signing-messages.md
```diff
@@ -31,12 +39,24 @@ If you did not generate the signature correctly the [status](/messaging/sms/guid
 
 > It is possible to *require* that all outgoing SMS use message signing. Contact support@nexmo.com to enable or disable this setting on your account.
```
marklewin: Not really part of your PR, but I'd much prefer this said "Nexmo can configure your account so that all outgoing SMS are signed. Contact support@nexmo.com if you require this."

lornajane: It's more like "all outgoing SMS must be signed" - there's one setting to enable getting signatures on incoming messages, you can send with signature any time you want to without any setting changes, and a second setting means that you can only send if the message is signed.

_documentation/concepts/guides/signing-messages.md
lornajane: Thanks!

_documentation/concepts/guides/signing-messages.md
lornajane: Authenticity of the contents ... I've reworded it a bit.

PR #1674, opened 2019-05-02
Fix NDP sidebar keyboard navigation

mheap:
## Description

Make the NDP sidebar nav keyboard accessible again

## Deploy Notes

N/A

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -143,7 +143,7 @@ module ApplicationHelper
 
       ss = []
       ss << "<li class='navigation-item--#{folder[:title]} navigation-item'>"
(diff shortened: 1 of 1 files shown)


mheap:
@lornajane I get that label when I press `f` on this branch. Any ideas why it's showing for me?

Screen Shot 2019-05-03 at 14 57 19

lornajane:
Ah, I see. It only seems to work on Chrome and I switched to Firefox since we added this feature. Also something has changed about how scrolled the bar is I think, it used to always be scrolled to the top so I could focus the div by selecting something on the top left but now it centres on the actual content so I can't access the scroll thing.

Maybe I could just have a sitemap? :)

PR #1682, opened 2019-05-08
Make Messages Code Snippets collapsible

tbedford:
## Description

As there are many new media types supported, the code snippet TOC was becoming crowded (and new types still to add for messenger and Viber channels). 

Each channel now has its own collapsible TOC entry.

Also tried to simplify naming of topics and make them more consistent.

Added redirects.


_documentation/messages/code-snippets/messenger/send-text.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: Send with Facebook Messenger
+title: Send a Text Message
```
mheap: Could we also set `meta_title` so that they show up in search results? e.g. `Send a message with Facebook Messenger`

See https://github.com/Nexmo/nexmo-developer/blob/master/_documentation/messaging/sms/overview.md#L3 for an example

tbedford: Thanks I had not seen that before - very useful! I will add in other areas to.

PR #1693, opened 2019-05-12
Testing homepage layouts

mheap:
## Description

*DO NOT MERGE*

I was playing around with the homepage design on my flight and landed on this. It still needs a lot of work, but I'm raising a PR to get a review app and solicit feedback

* Puts Messaging, Voice and Video front and center
* Hotjar shows that most people use the overview link, so all sub-links removed
* "All docs" link goes to `/api` for now but I think we'd need a different page showing all available products

## Deploy Notes

N/A

### DIFF
--- a/app/assets/stylesheets/objects/_card.scss
+++ b/app/assets/stylesheets/objects/_card.scss
@@ -94,6 +94,11 @@
     width: 25px;
   }
 
+  &:last-child {
+    border: 0;
+    margin-bottom: 0;
+  }
+
   p {
     color: $grey-darker;
   }
--- a/app/views/static/_products.html.erb
+++ b/app/views/static/_products.html.erb
@@ -1,399 +1,89 @@
 <div class="Vlt-grid Vlt-grid--center Vlt-grid--margin-top2">
   <div class="Vlt-col Vlt-col--1of3 Vlt-col--M-1of2">
     <div class="Vlt-card Vlt-article--reverse Vlt-card--image Nxd-card--products">
-      <div class="Vlt-card__image Vlt-card__image--app Vlt-card__image--purple">
-        <div class="Vlt-card__image__icon">
-          <svg>
-            <use xlink:href="/symbol/volta-icons.svg#Vlt-icon-message" /></svg>
-        </div>
+      <div class="Vlt-card__corner Vlt-bg-purple">
+        <svg class="Vlt-icon"><use xlink:href="/symbol/volta-icons.svg#Vlt-icon-message"/></svg>
       </div>
+
       <div class="Vlt-card__content">
-        <small class="Vlt-grey-dark">Send and receive</small>
         <h2>
           <a href="/messaging/sms/overview">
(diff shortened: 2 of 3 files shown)


benhayehudi:
Hi @mheap,

This PR is a great opportunity to discuss a bit more at length a larger conversation around the NDP home page. 

I think your work here illustrates a few points:

(a) Data-driven: Let's put forward what is most accessed prominently
(b) Ease of use: How do we get people to where they want to go with the greatest ease?
(c) Messaging: Who are we? Why should you care? How can we convey that in as few words as possible?

In regards to each one:

(a) You have clearly reorganized according to what our indicators are pointing to is the most frequently accessed information.
(b) I think we still have some work to do here: 
* I know this is something that needs more people to make happen, but the difference between *SMS API* and *Messages API* is not immediately clear. 
* We make prominent several of our APIs, but what about the tooling & resources the DevRel team has built to make those APIs easier to integrate? Under "Let's build something..." header, I wonder what it would look like to add more quick connections to specific resources we've created in this regard? Maybe to the Laravel framework library, for example, knowing that it is quite popular?
(c) This new homepage design still does not answer the question succinctly of who we are and why you should care. The banner we use at conferences says "Nexmo, The Vonage API Platform," which is accurate but falls victim to the same messaging issue. I would like to see something heading this redesign that clearly captures who we are and why a developer should care about us in 1 sentence (maybe 2 sentences). 

Just some thoughts for now!

Ben

marklewin:
I applaud the thought behind this. Our existing homepage is really cluttered and breaking it out into our core capabilities (Messaging, Voice, Video) is a great idea. However, I think is still a bit cluttered and "Need something else" looks like an afterthought. The stuff "below the fold" (Server SDKs, etc) might be missed entirely.

I think generally I prefer the [Amazon AWS approach](https://docs.aws.amazon.com/) approach. I think we're trying to say too much about each API on the home page instead of linking through to a landing page that says specifically what that API is intended for and provides access to the API ref, tutorials, etc.

For those who know what they want to do (Messaging, Voice, Video, Admin, etc) but don't know which API to use you could link from the group itself to a separate page that summarizes all the APIs in the group and helps them decide which one they need.

That way we have a central API "index" so that people who know exactly what they want can quickly navigate to it. And those that don't can click on, say, Messaging and see why they might want to use Messages API over SMS.

I agree with everything @benhayehudi says about having something that defines us and the confusion around Messaging/SMS. The only way I think we can differentiate their functionality in the name alone is to call one 'SMS API' and the other 'MultiChannel Messaging API' or something else that captures the spirit of what Messages API offers. "Messages" might be a catchier name, but it doesn't tell us much about it when we're offering an "SMS" API right next to it.

mheap:
I'm going to close this for now as there's a product grouping project happening that will clear up the messages/messaging confusion and inform our decisions here.

The thoughts above are very valuable and we're aiming to go in an interactive direction in the longer term. For now, we're looking to focus on information architecture only

PR #1707, opened 2019-05-20
Add nexmo oas renderer

fabianrbz:
## Description

Adds the new `nexmo-oas-renderer` and removes related code.
@mheap this is still a wip, the gem is added with a local path because it hasn't been merged/released yet.

A couple of things to note:
* there is some code duplication, we should extract all the filters and constraints into a gem since they are used both by the rails and the sinatra app
* -there are some [old routes](https://github.com/Nexmo/nexmo-developer/blob/master/config/routes.rb#L82) that are preventing me from removing `api#show`, are those still valid?-
* there's probably a lot of cleanup that we can do on the `assets` folder but first I want to be sure that everything is working

## EDIT:
- [x] `/api/verify/templates` is currently broken
- [x] `/api/developer/numbers` redirects to `api/numbers` and it is currently broken


mheap:
> there is some code duplication, we should extract all the filters and constraints into a gem since they are used both by the rails and the sinatra app

The duplication is ok for now I think. This is the first iteration and those files don't tend to change very often.

> there are some old routes that are preventing me from removing api#show, are those still valid?

Sadly they are, but we're moving to OAS for everything, so if we can document the old APIs using OAS we can remove all of this code.

> there's probably a lot of cleanup that we can do on the assets folder but first I want to be sure that everything is working

Agreed, let's do the cleanup later

mheap:
@fabianrbz There doesn't seem to be a gemspec in `nexmo-oas-renderer` which means I can't install it with the following line:

```
gem 'nexmo-oas-renderer', github: 'Nexmo/nexmo-oas-renderer'
```

fabianrbz:
@mheap have you merged the PR in `nexmo-oas-renderer` already? if not you should point to the [right branch](https://github.com/Nexmo/nexmo-oas-renderer/tree/encapsulate-into-gem)

PR #1716, opened 2019-05-28
Updated and added missing Voice code snippets

RabebOthmani:
Updated and added Voice code snippets for CSharp


marklewin:
Checked with @RabebOthmani and she is happy with the snippets, so approving and merging.

PR #1728, opened 2019-06-02
Add AltConf event page

mheap:
## Description

Add AltConf event page for @myrsiningos and @pardel 

## Deploy Notes

N/A


mheap:
Build failures are due to a GCP outage (so we can't download Chromium). Will check again in the morning

PR #1737, opened 2019-06-04
Prog. Conv. set up and architecture guides

brittBarak:
Added overview and setup guides with diagrams.

- system-overview 
- set-up
- set-up-authentication


brittBarak:
@tbedford  Hi :) would appreciate your feedback on the guides themselves, as well as where to locate them. 
Ideally, it should be under Programmable Conversation entry in the left menu. For the time being, I put it under Conv. api. Still, I'm thinking how to arrange that. maybe add a "get started" collapsible title? and add the [How To generate JWT there as well](https://developer.nexmo.com/conversation/concepts/jwt-acl) ? 

Also, maybe the new System Overview should be merged with the existing Overview? I'm not sure. 

What do you think?

mheap:
The deploy error will be due to the submodule removals. If we merge `master` in to this branch they should be resolved

PR #1768, opened 2019-06-17
EuRuKo landing page

benhayehudi:
This PR adds a landing page for EuRuKo 2019


benhayehudi:
> @benhayehudi looks good to me, one question though, why the `e` namespace in`/e/euruko`?

That's for our namespacing of landing pages to keep them manageable. We will use `nexmo.dev/euruko` for all public facing linking.

PR #1781, opened 2019-06-23
Add MermaidJS diagrams

mheap:
## Description

> The generated diagrams need some styling work before this is merged

Adds the ability to use any [Mermaid](https://mermaidjs.github.io/) diagram within the docs e.g.

````
```mermaid
graph TD;
    A-->B;
    A-->C;
    B-->D;
    C-->D;
```
````

There is a helper for sequence diagrams as they're the most common type:

````
```sequence_diagram
```
```

## Deploy Notes

N/A


lornajane:
The left hand menu looks weird to me here https://nexmo-developer-pr-1781.herokuapp.com/contribute/overview (compared to https://developer.nexmo.com/contribute/overview) and I can't navigate to where the diagrams are. Am I doing something wrong?

mheap:
I managed to reproduce this locally today. The issue occurs when `assets:compile` is used but not when webpack dev server is run. I'm still investigating

mheap:
This should all be rendering correctly now, using the `dark` theme which means blue colours. This is good enough to ship imo, and we can add another colour scheme later if required

PR #1783, opened 2019-06-23
Add Split for A/B testing

mheap:
## Description

This adds the `Split` gem to perform A/B testing. If `REDIS_URL` is not provided as an environment variable then Split is not loaded.

To track a click on an element, add `data-ab="experiment_name"` to it. This will track a completion for `:experiment_name` on click

## Deploy Notes

This needs Redis to run. Without `REDIS_URL` being set, Split will not be loaded

### DIFF
--- a/config/initializers/split.rb
+++ b/config/initializers/split.rb
@@ -1,0 +1,3 @@
+Split.configure do |config|
+  config.enabled = !!ENV['REDIS_URL']
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -90,3 +90,6 @@ Style/Next:
 
 Style/ConditionalAssignment:
   Enabled: false
+
+Style/DoubleNegation:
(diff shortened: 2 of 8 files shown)


config/initializers/split.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    def ab_finished(_name); end
+  end
+end
+
+if ENV['REDIS_URL']
+  require 'split'
+  require 'split/dashboard'
+
+  # Not sure why I have to include these. `split/engine` should do
```
fabianrbz: @mheap you should be able to get rid of this if you remove the `require: false` from the Gemfile and changing this initializer to:
```
Split.configure do |config|
  config.enabled = !!ENV['REDIS_URL']
end
```

mheap:
@fabianrbz Don't write code after midnight - lesson learned!

PR #1784, opened 2019-06-24
bringing in olympus typo

pardel:
## Description

Bringing in Olympus type - showing Hugh how to use submodules

## Deploy Notes

no changes to deployment process


pardel:
used as a test to show Hugh how merging works when adding changes to external repos

PR #1788, opened 2019-06-27
Correct E.164 statement

peter-nexmo:
The statement suggested Nexmo handled E.164 format incorrectly, whereas international access codes including + are not part of E.164 formatting.

## Description

## Deploy Notes

### DIFF
--- a/_documentation/concepts/guides/glossary.md
+++ b/_documentation/concepts/guides/glossary.md
@@ -118,11 +118,11 @@ A measure of the impact of inbound phone calls in digital marketing efforts. For
 
 ## E.164 format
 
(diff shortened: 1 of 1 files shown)


mheap:
Looks like the `+` is optional (see [Freeswitch](https://freeswitch.org/confluence/display/FREESWITCH/Regular+Expression) which is a reasonable source of truth)

peter-nexmo:
The access codes are used by carriers but do not form part of the E.164 numbering format. The ITU doc/Wikipedia explain that the `+` or other international access code is usually added so that the recipient party can call the caller back easily, but does not form part of the E.164 formatted number.

PR #1796, opened 2019-07-01
API Specification Update

NexmoDev:


mheap:
`application` and `conversion` are included in this PR but will not be loaded as they don't exist in the `open_api_constraint` so it's safe to merge

PR #1797, opened 2019-07-02
API Specification Update

NexmoDev:


lornajane:
Something went wrong! Visit https://nexmo-developer-pr-1797.herokuapp.com/api and none of the API reference pages seem to work ... it's not just the Messages API one that I broke

mheap:
The issue with with review app configs. I resolved and checked and it's working great

PR #1801, opened 2019-07-03
Remove Turbolinks

mheap:
## Description

Turbolinks has been causing issues where JS isn't executed as it's not
a real page refresh. This means we keep adding workarounds for third
party services

This commit removes Turbolinks, and as far as I can tell it doesn't have
a negative performance impact.

## Deploy Notes

Check things thoroughly with this deploy as it's quite a big change


app/assets/javascripts/application.js
```diff
@@ -12,14 +12,14 @@
...
 //= require jquery_ujs
-//= require turbolinks
 //= require jquery-scrolltofixed
 //= require volta/volta.core.js
 //= require volta/components/volta.accordion.js
 //= require volta/components/volta.tab.js
 //= require volta/components/volta.tooltip.js
 //= require volta/components/volta.modal.js
 //= require volta/components/volta.dropdown.js
+//= require volta/components/volta.menu.js
```
benhayehudi: Are we also including the Volta menu scripts as part of this PR? Is that a separate issue?

app/javascript/packs/application.js
```diff
@@ -67,15 +56,7 @@ let refresh = () => {
...
-  Volta.init(['accordion', 'tooltip', 'tab', 'modal', 'dropdown'])
-
-  // Fix for Turbolinks scrolling to in-page anchor when navigating to a new page
-  if(window.location.hash){
-    const tag = document.getElementById(window.location.hash.slice(1))
-    if(tag){
-      tag.scrollIntoView(true);
-    }
-  }
+  Volta.init(['accordion', 'tooltip', 'tab', 'modal', 'dropdown', 'menu'])
```
benhayehudi: Same question as above.

app/pipelines/markdown_pipeline.rb
```diff
@@ -35,8 +35,7 @@ def initialize(options = {})
       BreakFilter,
       UnfreezeFilter,
       IconFilter,
-      ExternalLinkFilter,
-      TutorialLinkFilter
```
benhayehudi: How is this connected to the removal of Turbolinks?

app/assets/javascripts/application.js
mheap: This is intended. We used to have a `Navigation.js` file that contained a copy of the Volta menu code with some Turbolinks enhancements. Now that we've removed Turbolinks we can use the standard Volta menu code

app/pipelines/markdown_pipeline.rb
mheap: The `TutorialLinkFilter` detected if any link starts with `/tutorial` and added an attribute to disable turbolinks for that link, which is no longer needed

vendor/assets/javascripts/volta/components/volta.menu.js
```diff
@@ -0,0 +1,377 @@
+/**
```
fabianrbz: @mheap can we use [this one](https://github.com/Nexmo/nexmo-developer/blob/master/vendor/assets/javascripts/volta/components/side-navigation/volta.menu.js) instead?

PR #1808, opened 2019-07-03
Fix nested navigation after Turbolinks removal

mheap:
## Description

Fix nested navigation in sidebar by restoring the old `navigation.js` file

## Deploy Notes

N/A

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -240,8 +240,9 @@ module ApplicationHelper
   def output_nested_dropdown(item, title)
     ss = []
     ss << "<li class=\"js--collapsible navigation-item--#{title.parameterize} navigation-item\">"
-    ss << "<a class=\"Vlt-sidemenu__trigger\" href=\"/concepts\"><span class=\"Vlt-sidemenu__label\">#{title}</span></a>"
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -240,8 +240,9 @@ def output_link(item, title, options)
   def output_nested_dropdown(item, title)
     ss = []
     ss << "<li class=\"js--collapsible navigation-item--#{title.parameterize} navigation-item\">"
-    ss << "<a class=\"Vlt-sidemenu__trigger\" href=\"/concepts\"><span class=\"Vlt-sidemenu__label\">#{title}</span></a>"
+    ss << "<a class=\"Vlt-sidemenu__trigger\" href=\"#{path_to_url(item[:path])}\"><span class=\"Vlt-sidemenu__label\">#{title}</span></a>"
```
fabianrbz: @mheap I think we don't need this part

app/helpers/application_helper.rb
```diff
@@ -240,8 +240,9 @@ def output_link(item, title, options)
...
     ss << output_children(item[:children]) if item[:children]
+    ss << '<li>'
```
fabianrbz: same here

mheap: This was a bug, updated to `</li>`

PR #1809, opened 2019-07-04
Add flash messages when submitting to spotlight.

fabianrbz:
## Description

Add flash messages when submitting to spotlight in order to provide feedback to the user.


fabianrbz:
yeah, it accepts everything you threw at it :p, the error check is just in case something happens

PR #1810, opened 2019-07-04
CodeSnippetsFilter Testing

benhayehudi:
## Description

Create tests for `CodeSnippetsFilter`

## Deploy Notes


mheap:
This is a great start, but it relies on state on disk. The next step is to mock the calls to `Dir`, `File.read` etc and ensure that if you remove the existing code snippet the tests continue to pass

mheap:
Merging as this is valuable and we can refactor to mocks going forwards

PR #1811, opened 2019-07-04
Point NDP at latest node code snippets

tbedford:
## Description

Point NDP at latest master of Node code snippets (was pointing at branch).


tbedford:
Updates were made in another PR.

PR #1827, opened 2019-07-10
Remove initialization files from OAS and auto-detect multiple response schemas

mheap:
## Description

We can automatically detect when an OAS document contains JSON and XML. Let's do this instead of having to create an `initialization` config file

## Deploy Notes

N/A


benhayehudi:
⭐️ 🌟 👍 💙 🚀

PR #1832, opened 2019-07-11
Use greenhouse's API for jobs section

fabianrbz:
## Description

- [x] Cleanup old code
- [x] Pull DevRel jobs from Greenhouse
- [x] Cache jobs 

## Deploy Notes

This PR contains two migrations that remove two tables `careers` and `friendly_id_slugs`
Adds `redis`, `truncato` and `greenhouse_io`
We need to set up the redis add-on in heroku.

# EDIT:
I tried the add-on in the review app and worked fine, we need to select a plan for production.
Also, if we want to schedule the cache expiration we need to set up [this](https://devcenter.heroku.com/articles/scheduler)


benhayehudi:
Spent some time reviewing this, and just want to say it looks really well done! 💪

mheap:
@fabianrbz It seems to be showing more than just DevRel jobs on https://nexmo-developer-pr-1832.herokuapp.com/team

Looking at the code it should be filtering to our department (or are all these roles tagged as our department?)

fabianrbz:
@mheap sadly they are all tagged as our department

mheap:
Ok, I think we'll need to filter on title too in that case. Select them if they contain any of the following (case insensitive):

* SDK
* Advocate
* Community Manager

fabianrbz:
@mheap cool, also, should I automatically expire the cache after a while?

fabianrbz:
@mheap I'm expiring them after one hour according to a Jira ticket that I found, shouldn't it be a bit longer?

mheap:
I’m happy to take recommendations here. An hour felt like a good balance
(24 API hits/day) but we could go up to every 3 hours or so without much
impact

fabianrbz:
yeah, I'm not sure how often Greenhouse's job board is updated, I thought that hitting the API once a day would be ok, but if you are fine with 1 hour, let's leave it like that

PR #1865, opened 2019-07-25
Rename VBC Number Programmability to Smart Numbers

marklewin:
## Description

Request from Brian Dorry. He claims that everything is the same apart from the name change.


marklewin:
Marking as don't merge because I want Brian to provide a new screenshot for the Provisioning guide

marklewin:
Brian Dorry has approved these changes in the review app, so we can merge once there's an approving review.

PR #1871, opened 2019-07-30
API Specification Update

NexmoDev:


mheap:
This was pulled in as part of #1868

PR #1885, opened 2019-08-07
Create YAML files for App v2 Ruby Snippets

benhayehudi:
## Description

Add the YAML config files for the application v2 Ruby snippets

## Deploy Notes


benhayehudi:
Hi @fabianrbz - Can I get a second set of 👀 on this? I've reviewed the demo app and checked each Application snippet and they are all outputting correctly, so I think it's good to go!

PR #1890, opened 2019-08-07
add yaml config for two remaining ruby voice snippets

benhayehudi:
## Description

* Outbound call with NCCO
* Track NCCO progress

## Deploy Notes


benhayehudi:
@fabianrbz - If you have a moment, another quick one for you, thank you!

All you need to do is verify (I've done it also) that the Ruby code snippets at the following URLs come up alright: 

* `/voice/voice-api/code-snippets/make-an-outbound-call-with-ncco/`
* `/voice/voice-api/code-snippets/track-ncco-progress`

PR #1918, opened 2019-08-21
Convert Check Redirects action to YAML

mheap:
## Description

Github is moving from HCL to YAML for their actions config, which means that we need to update our configs

## Deploy Notes

N/A


mheap:
Closing for now as it's a much bigger change than initially intended

PR #1923, opened 2019-08-22
Add feature specs

fabianrbz:
## Description

Add basic feature specs for `community`, `team`, `extend`  and `landing` pages


mheap:
These tests feel really brittle to me. Do we need to assert the exact structure, or would looking for key elements be enough?

fabianrbz:
@mheap yeah I know, it's kind of up to us. I relied heavily on the structure because I wanted to test that the order of the elements was correct, but we can drop that and just check the content.

Issue #1143, opened 2018-10-22
Check Authentication topic and review JWT creation


tbedford:
## Description

From Tony Chan:

1) https://developer.nexmo.com/concepts/guides/authentication

    Should this table also mention “Stitch” in the list of APIs, or rename “Voice” to be “Voice/Stitch”?

2) The developer site has lots of information on how to generate JWT for Stitch but almost none on how to generate JWT for Voice and what parameters are necessary. That seems to be a hindrance for developers wanting to use Voice API vs. Voice SDKs. Could we improve that?

---

tbedford:
I have it on my TODO.

tbedford:
See https://github.com/Nexmo/nexmo-developer/pull/1925

PR #1925, opened 2019-08-23
Add Conversation API. Add info on generating JWTs.

tbedford:
## Description

* Fix issue #1143
* Adds additional link from webhooks topic to Application topic.

### DIFF
--- a/_documentation/concepts/guides/authentication.md
+++ b/_documentation/concepts/guides/authentication.md
@@ -123,7 +123,7 @@ If you are not using a Nexmo library you should refer to [RFC 7519](https://tool
 
 ### Using the Nexmo CLI to generate JWTs
 
-The Nexmo CLI is one way of creating a JWT. The general syntax is:
+The Nexmo CLI provides a command for generating a JWT. The general syntax is:
 
 ``` shell
 nexmo jwt:generate [options] <private_key> [claim=value...]
--- a/_documentation/concepts/guides/authentication.md
+++ b/_documentation/concepts/guides/authentication.md
@@ -125,19 +125,19 @@ If you are not using a Nexmo library you should refer to [RFC 7519](https://tool
 
 The Nexmo CLI is one way of creating a JWT. The general syntax is:
 
-```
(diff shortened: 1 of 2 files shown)


_documentation/concepts/guides/authentication.md
```diff
@@ -120,6 +121,30 @@ Name | Description | Required
 
 If you are not using a Nexmo library you should refer to [RFC 7519](https://tools.ietf.org/html/rfc7519) to implement JWT.
 
+### Using the Nexmo CLI to generate JWTs
+
+The Nexmo CLI is one way of creating a JWT. The general syntax is:
+
+``` shell
```
mheap: I don't believe there should be a space here

tbedford: Emacs GFM mode puts a space in automatically so I assume that works. Is the codeblock failing to render correctly in some way?

_documentation/concepts/guides/authentication.md
```diff
@@ -120,6 +121,30 @@ Name | Description | Required
 
 If you are not using a Nexmo library you should refer to [RFC 7519](https://tools.ietf.org/html/rfc7519) to implement JWT.
 
+### Using the Nexmo CLI to generate JWTs
+
+The Nexmo CLI is one way of creating a JWT. The general syntax is:
```
marklewin: I'm cringing at my pedantry here, but I think we should try and be as clear as possible at all times, particularly now that we're writing for translation.

The Nexmo CLI isn't one way of creating a JWT, it's a suite of commands for doing all sorts of Nexmo goodness. One way to create a JWT is to use the Nexmo CLI. Or the Nexmo CLI provides a command for creating a JWT.

Hope that makes sense!

_documentation/concepts/guides/authentication.md
mheap: You're right! It does support a space in the commonmark spec https://spec.commonmark.org/0.12/#fenced-code-blocks

PR #1932, opened 2019-08-27
update node code snippets

AlexLakatos:
## Description

Add missing node code snippets for:
https://developer.nexmo.com/voice/voice-api/code-snippets/make-an-outbound-call-with-ncco
https://developer.nexmo.com/voice/voice-api/code-snippets/record-a-call-with-split-audio
https://developer.nexmo.com/voice/voice-api/code-snippets/record-a-call
https://developer.nexmo.com/voice/voice-api/code-snippets/record-a-conversation
https://developer.nexmo.com/verify/code-snippets/search-verify-request
https://developer.nexmo.com/verify/code-snippets/send-verify-request-with-workflow
https://developer.nexmo.com/account/code-snippets/create-a-secret
https://developer.nexmo.com/account/code-snippets/revoke-a-secret
https://developer.nexmo.com/account/code-snippets/list-all-secrets
https://developer.nexmo.com/account/code-snippets/fetch-a-secret
https://developer.nexmo.com/voice/voice-api/code-snippets/track-ncco-progress
https://developer.nexmo.com/voice/voice-api/code-snippets/transfer-a-call-inline-ncco


_examples/account/secret-management/create-a-secret/node.yml
```diff
@@ -0,0 +1,15 @@
+---
+title: Node
```
mheap: Only `client` and `code` are required thanks to some defaults that we added

_examples/voice/make-an-outbound-call-with-ncco/node.yml
```diff
@@ -0,0 +1,16 @@
+---
+title: Node
+language: node
+dependencies:
+    - nexmo
+    - dotenv
```
mheap: Why do we need `dotenv`? It should be a snippet implementation detail, with the actual usage using one of the placeholder variables

_examples/voice/record-a-call-with-split-audio/node.yml
```diff
@@ -0,0 +1,13 @@
+---
+title: Node
+language: node
+dependencies:
+    - express
+    - body-parser
```
mheap: We need to leave `dependencies` in this one as it uses more than just `nexmo`

AlexLakatos:
@mheap the client / code thing seems to not be the default across node, I'm going to go ahead and clean up the rest, it's confusing when working on new one.

AlexLakatos:
The last two ones won't render until https://github.com/Nexmo/nexmo-node-code-snippets/pull/68 is merged and the code snippets pulled in.

tbedford:
On the blocks that don't render - what we usually do is point the main PR at the correct working branch in the code snippet repo. You then have a fully working review app. At that point it's possible to merge the code snippet repo and your main PR can be updated to point at the latest code snippet. Also, if the code snippets are updated, a PR is automatically created to prompt you to update NDP.

AlexLakatos:
I've updated the code with @marklewin 's feedback and pointed the submodule to the working branch in the code snippets repo, so now everything should build ok.

For Verify(request, and request-with-workflow), the code snippet specs say that `BRAND` should be a variable, but everybody is hardcoding it in the snippets, slightly different, mostly variations of Acme or Nexmo, and the text on NDP doesn't list `BRAND` as a variable. 

I've added `BRAND_NAME` as a variable in the text and snippet, but I'm not sure if that's the correct approach here. Should we update the specs to agree with the existing code snippets, or should we change the existing snippets to agree with the spec? I personally think we should have the variable `BRAND_NAME` in the snippets, it's a bit more descriptive than `BRAND`, and update the spec and the other snippets to use it. But I'll need someone else to weigh in here, I don't have a strong opinion either way.

lornajane:
I agree that the `BRAND_NAME` should be the variable. I think the correct process is probably to:
 - update the spec
 - make your code snippet match
 - open tickets on the `code_snippets` repos for anything that now doesn't match the updated spec

AlexLakatos:
I've gone ahead an opened a PR for the specs repo. The code snippets here already use BRAND_NAME.

https://github.com/Nexmo/code-snippet-specs/pull/23

PR #1936, opened 2019-08-27
Basic text formatting note

tbedford:
## Description

Add note on text formatting.


tbedford:
Not totally sure of the best way to format this.

PR #1938, opened 2019-08-28
Mock Filesystem in CodeSnippetsFilter Testing

benhayehudi:
## Description

Remove the dependency on the current state of the filesystem in the `CodeSnippetsFilter` test

## Deploy Notes


fabianrbz:
@benhayehudi can you update the snapshot file?

mheap:
I found why the tests failed in the previous PR. The old title was `Node` and the new defaults title is `Node.js`, so the tab titles changed.

Screen Shot 2019-09-03 at 13 55 08

The reason that the tests still pass when we change the contents on disk of the node code snippet, is that we mock the method that loads the contents so that we're only testing tab generation rather than file loading in this test

mheap:
Closing this for now as the test failed correctly and no mocking is required

PR #1941, opened 2019-08-29
Use Message Search OAS 3 spec.

tbedford:
## Description

* Try to use the OAS3 spec for Message Search API.
* Remove the old API pages for Account API (now replaced by OAS3 spec), and Message Search API.

Now awaiting renderer release for link `/api/developer/messages` to work. (Note the renderer has been fixed, just waiting for a new release)

## Review

Review app available.


tbedford:
@fabianrbz : any ideas Fabian why the spec is not being picked up? The definition is present at master `api-specification/definitions/developer/messages.yml`. Also NDP master should be pointing at the latest master. Is there something else that needs changing for the spec to be picked up? Thanks.

fabianrbz:
@tbedford I know what's going on, you need to update the [constraints file](https://github.com/Nexmo/nexmo-oas-renderer/blob/master/lib/nexmo/oas/renderer/constraints/open_api.rb) to include `developer/messages`

PR #1942, opened 2019-08-29
Update snippet to remove unneeded dependency

lornajane:
## Description

We had Slim framework as a dependency for a snippet that didn't need it.


lornajane:
This one https://nexmo-developer-pr-1942.herokuapp.com/voice/voice-api/code-snippets/make-an-outbound-call (the file/snippet names are a bit not-matching)

PR #1956, opened 2019-09-05
Fix Geocoding for events

mheap:
## Description

The Google maps geocoder was intepreting "Tel Aviv, IL" as a place called Tel-Aviv in Chicago, Illinos. This PR provides the full country name to the geocoder to make the results more accurate

## Deploy Notes

N/A

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -150,6 +150,7 @@ gem 'terminal-table'
 gem 'lograge'
 
 # Country picker for ActiveAdmin
+gem 'countries'
 gem 'country_select', '~> 4.0'
 
 gem 'nexmo-oas-renderer', '~> 0.4.1', require: false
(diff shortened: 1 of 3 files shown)


benhayehudi:
Thanks, @mheap for this PR. It also needed an increase in the character limit for `country` in the DB, so I have gone ahead and incorporated that, and it now works successfully. I am going to go ahead and approve it, and let you confirm this change I have incorporated and merge when ready.

mheap:
@benhayehudi Why does it need the column increase? The country should be stored as a 2 character code, then converted to a full country name at geocoding time (unless I've misunderstood how things work)

benhayehudi:
> @benhayehudi Why does it need the column increase? The country should be stored as a 2 character code, then converted to a full country name at geocoding time (unless I've misunderstood how things work)

Oops... I misunderstood how it works then, not you. This is what happens on jetlag brain. Let me have another look given your comment.

benhayehudi:
@mheap - review is done, I've verified it works as expected w/o any addl changes

PR #1991, opened 2019-09-17
Create send media mtm

hughhopkins:
## Description

how to send a rich media WhatsApp message

## Review

Review app will be available via Heroku


tbedford:
Approving with the proviso that we come back to this and do the code snippet properly, that is, the code should be in the corresponding code snippet repo and independently tested from there. I have created a JIRA ticket to revisit this one.

PR #1994, opened 2019-09-18
Add support for rendering folders as Tabbed Content

benhayehudi:
## Description

This PR refactors the process for generating tabs for content on NDP. Whereas heretofore tabs were determined by a YAML file pointing to another folder elsewhere with the content, now there will be a folder of markdown files each representing a tab with a `.config.yml` file in the folder with `tabbed` option either set to `true` or `false`.

## Deploy Notes


benhayehudi:
@mheap - Can you please review this at this point?

app/controllers/markdown_controller.rb
```diff
@@ -71,13 +73,45 @@ def set_document_path_when_file_name_is_the_same_as_a_linkable_code_language
     @code_language = nil
   end
 
+  def check_if_path_is_folder
```
mheap: This starts with `check` but has side effects.

In general, I'm a fan of methods returning values to be used rather than populating instance variables

app/controllers/markdown_controller.rb
```diff
@@ -9,16 +9,18 @@ def show
     redirect = Redirector.find(request)
     return redirect_to redirect if redirect
 
-    @frontmatter = YAML.safe_load(document)
+    if check_if_path_is_folder == false
```
mheap: This could be refactored in to something similar to:

```
if path_is_folder?
  [@content, @frontmatter] = content_from_folder
else
  [@content, @frontmatter] = content_from_file
end
``

The statement in the `if` shouldn't have a side effect

File: app/filters/tab_filter.rb [hunk 3]
```diff
@@ -1,10 +1,18 @@
 class TabFilter < Banzai::Filter
   def call(input)
-    input.gsub(/^(\s*)```tabbed_(examples|content)(.+?)```/m) do |_s|
+    input.gsub(/^(\s*)```tabbed_(examples|content|folder)(.+?)```/m) do |_s|
       @indentation = $1
       @mode = $2
       @config = YAML.safe_load($3)
-      validate_config
+
+      if @mode == 'folder'
```
mheap: It looks like the other options have `is_tabbed_X?` methods. Should we add one here for consistency?

app/helpers/application_helper.rb
```diff
@@ -34,7 +34,13 @@ def directory_hash(path, name = nil)
       next if IGNORED_PATHS.include? entry
       full_path = File.join(path, entry)
       if File.directory?(full_path)
-        data[:children] << directory_hash(full_path, entry)
+        # if its a tabbed folder
+        if File.exist?("#{full_path}/.config.yml") && YAML.safe_load(File.read("#{full_path}/.config.yml"))['tabbed'] == true
```
mheap: We're reading the config twice here. Can we combine it in to one read?

spec/filters/tab_filter_spec.rb
```diff
@@ -27,4 +27,155 @@
...
+        ```tabbed_folder
+        source: /path/to/a/directory
+        ```
+      HEREDOC
+      expected_output = "FREEZESTARTPGRpdiBjbGFzcz0iVmx0LXRhYnMiPgogIDxkaXYgY2xhc3M9IlZsdC10YWJzX19oZWFkZXIgVmx0LXRhYnNfX2hlYWRlci0tYm9yZGVyZWQiIGRhdGEtaGFzLWluaXRpYWwtdGFiPSJmYWxzZSI-PGRpdiBjbGFzcz0iVmx0LXRhYnNfX2xpbmsgVmx0LXRhYnNfX2xpbmtfYWN0aXZlIiBkYXRhLWxhbmd1YWdlPSJqYXZhc2NyaXB0IiBkYXRhLWxhbmd1YWdlLXR5cGU9Imxhbmd1YWdlcyIgZGF0YS1sYW5ndWFnZS1saW5rYWJsZT0idHJ1ZSI-PHNwYW4-PHN2Zz48dXNlIHhsaW5rOmhyZWY9Ii9hc3NldHMvaW1hZ2VzL2JyYW5kcy9qYXZhc2NyaXB0LnN2ZyNqYXZhc2NyaXB0Ij48L3VzZT48L3N2Zz48c3Bhbj5GaXJzdCBTYW1wbGUgTWFya2Rvd248L3NwYW4-PC9zcGFuPjwvZGl2PjwvZGl2PgogICAgPGRpdiBjbGFzcz0iVmx0LXRhYnNfX2NvbnRlbnQiPgogICAgPGRpdiBjbGFzcz0iVmx0LXRhYnNfX3BhbmVsIFZsdC10YWJzX19wYW5lbF9hY3RpdmUiPjxwIGFyaWEtbGFiZWxsZWRieT0nIklEMTIzNDU2IicgYXJpYS1oaWRkZW49InRydWUiPjxwPiAjIyBIZWFkaW5nCiBTYW1wbGUgY29udGVudDwvcD48L3A-PC9kaXY-CjwvZGl2Pgo8L2Rpdj4KFREEZEEND\n"
+      expect(described_class.new.call(input)).to eq(expected_output)
+    end
+
+    it 'renders content with two markdown files in input' do
+      expect(File).to receive(:directory?).with('/path/to/a/directory').and_return(true)
```
mheap: `/path/to/a/directory` can be stored in a variable and used everywhere rather than repeated definition

spec/filters/tab_filter_spec.rb
```diff
@@ -27,4 +27,155 @@
...
+      expect(File).to receive(:read).with('/path/to/a/directory/.config.yml').and_return(config_tabbed_true)
+      expect(Dir).to receive(:glob).with('/path/to/a/directory/*.md').and_return(['/path/to/a/directory/javascript.md', '/path/to/a/directory/android.md'])
+      expect(File).to receive(:exist?).with('/path/to/a/directory/javascript.md').and_return(true)
```
mheap: This expectation could be extracted e.g.

```ruby
def mock_content(name, content)
  expect(File).to receive(:exist?).with("/path/to/a/directory/#{name}.md").and_return(true)
  expect(File).to receive(:read).with("/path/to/a/directory/#{name}.md").and_return(content)
end

mock_content('javascript', first_sample_markdown)
mock_content('android', second_sample_markdown)
```

spec/filters/tab_filter_spec.rb
```diff
@@ -27,4 +27,155 @@
...
+      expect(File).to receive(:exist?).with('/path/to/a/directory/android.md').and_return(true)
+      expect(File).to receive(:read).with('/path/to/a/directory/javascript.md').and_return(first_sample_markdown)
+      expect(File).to receive(:read).with('/path/to/a/directory/android.md').and_return(second_sample_markdown)
+      expect(SecureRandom).to receive(:hex).at_least(:once).and_return('ID123456')
```
mheap: `at_least` is a code smell when we know exactly how many times it should be called (twice, in this case)

app/helpers/application_helper.rb
benhayehudi: We're reading the file once with `File.read`, and checking if it exists once with `File.exist?` prior to reading it. Removing the check for its existence breaks the code.

app/filters/tab_filter.rb
benhayehudi: I am not sure what you mean by this comment. All the other points have been addressed in your review. Could you explicate this further, or demonstrate? Thanks! 🙏

PR #2008, opened 2019-09-24
add info on media templates

hughhopkins:
## Description

helpful info

### DIFF
--- a/_documentation/messages/code-snippets/whatsapp/send-media-mtm.md
+++ b/_documentation/messages/code-snippets/whatsapp/send-media-mtm.md
@@ -8,6 +8,8 @@ navigation_weight: 2
 
 In this code snippet you will see how to send a WhatsApp Template using the Messages API. To send the Media Message Template you need to use the Messages custom object. The custom object takes a partial section of the original WhatsApp API request and sends it directly though.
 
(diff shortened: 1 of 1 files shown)


_documentation/messages/code-snippets/whatsapp/send-media-mtm.md
```diff
@@ -8,6 +8,8 @@ navigation_weight: 2
 
 In this code snippet you will see how to send a WhatsApp Template using the Messages API. To send the Media Message Template you need to use the Messages custom object. The custom object takes a partial section of the original WhatsApp API request and sends it directly though.
 
+WhatsApp Media Templates consist of a Header, Body and Footer structure. The Header contains the media which can be text, location, video, image or file. The Body contains the text message. This is currently limited to 160 characters when displayed to the end user; the reason being that WhatsApp don’t want the message to extend the fold. The Footer is optional and contains a static text option.
```
marklewin: I just want to be sure that we know what you mean by "extend the fold" here, Hugh, especially as this might not translate very well. Do you mean that the entire message body should be visible without scrolling?

tbedford: Also, I would appreciate some clarification on what exactly a "static text option" is please.

hughhopkins: 1. below the fold is below the screen. i.e. if it's a super long message then the user is going to have to scroll below the fold. 
2. static text meaning that no variable can be added into the template in the footer. The footer is always going to say the same text.

PR #2023, opened 2019-10-02
Remove Alumni Section

benhayehudi:
## Description

This PR removes the alumni section from the Team information on NDP.

## Deploy Notes


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_07_11_143408) do
+ActiveRecord::Schema.define(version: 2019_09_05_151355) do
```
mheap: This doesn’t look like it should be in here

benhayehudi: Resolved this, thanks!

PR #2026, opened 2019-10-02
Upgrades to Node 12 to fix Webpacker error

conshus:
## Description

When following the instructions, I got the following error when trying to view the site:
image
which I found was due to this error:
image
Upgrading Node fixes the problem, but had to change the package.json to allow for versions larger than 10.

## Deploy Notes


package.json
```diff
@@ -68,7 +68,7 @@
   },
   "version": "1.1.0",
   "engines": {
-    "node": "> 8.0.0 <= 10",
```
mheap: Could we explicitly pin to versions <= 12 instead of removing the constraint?

When new versions are released it breaks out build process unless we pin to a known working version

conshus: Got it. Makes sense. Keep it to what you know works.

package.json
```diff
@@ -68,7 +68,7 @@
...
+    "node": "<=12",
```
mheap: Do you think we should update this to `>= 12.0.0 < 13`? That gives us an explicit range that we know works.

The current change could install node `0.6` which would satisfy the criteria but wouldn't work for NDP

conshus: Made the change.

PR #2029, opened 2019-10-03
Search NDP on request query path to improve 404 results

benhayehudi:
## Description

This PR adds a `find_by_search` method to the `Redirector` service, wherein if the request path is not found in the environment redirects, it initiates a search query using Algolia, as recommended in the related Jira ticket.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,6 @@
 class ApplicationController < ActionController::Base
+  include ApplicationHelper
+
   rescue_from Errno::ENOENT, with: :not_found
   rescue_from ActiveRecord::RecordNotFound, with: :not_found
   protect_from_forgery with: :exception
@@ -16,6 +18,7 @@ class ApplicationController < ActionController::Base
       redirect_to redirect
     else
       NotFoundNotifier.notify(request)
+      not_found_search_results if search_enabled?
       render_not_found
     end
   end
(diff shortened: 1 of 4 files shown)


benhayehudi:
@mheap - As you review this, I have an implementation question for you. Right now, the existing test is failing, because as I have currently built this new method into the `Redirector` class, it will always default to a search if it can't be found in an environment redirect file. However, I am not sure if that is how we want the behavior to act. If so, it changes the existing behavioral assumptions for the class, and for it's accompanying existing tests. I can make an argument that this is desired but would like to hear your thoughts as well.

mheap:
I'm not sure that this is the behaviour that we want (at least, automatically redirecting). I think I'd rather see this behaviour in the 404 handler, listing (maybe five?) potential pages for the user to choose from rather than automatically redirecting them.

This means that the page will continue to return a 404, rather than every URL on NDP returning a 200

benhayehudi:
> I'm not sure that this is the behaviour that we want (at least, automatically redirecting). I think I'd rather see this behaviour in the 404 handler, listing (maybe five?) potential pages for the user to choose from rather than automatically redirecting them.
> 
> This means that the page will continue to return a 404, rather than every URL on NDP returning a 200

Thanks, Michael. Maintaining a 404 page, and not making every URL "work" makes more sense, upon further reflection. I refactored my code by moving the new functionality into the 404 processes, and out of the `Redirector`. The 404 view now presents 5 options for a user based on their query, as opposed to a link back to the home page like previously.

PR #2035, opened 2019-10-07
Add a list of language definitions for Verify API

lornajane:
## Description

We have a list of allowed values as an enum in the API reference, but no lookup for what the language codes mean - so I'm adding this. Currently a work in progress as we don't have definitions for all the languages.


lornajane:
Guillermo is happy so as long as one of the rest of you is, this is good to go. Thanks :)

PR #2039, opened 2019-10-07
Adding link to bios from image in sessions

leggetter:
## Description

Adding link to bios from image in sessions


leggetter:
Seems that Vamis' bio is missing. Would also be great to have the over links appear one sessions headers/titles.

PR #2040, opened 2019-10-08
Make Contact Center Guide a Use Case

marklewin:
## Description

As this is described as a use case, I've had a go at turning it into one.


tbedford:
Needs a suitable redirect.

PR #2046, opened 2019-10-10
I18n v1

fabianrbz:
## Description

This branch was updated with master, and a few things needed to be addressed, e.g.: adding support for i18n to use_cases and tutorials

## Notes

• there is a url (probably more than one) that’s not being generated correctly (but still works)
• need to make sure that tutorials/use-cases are working correctly
• need to make those things work with the fallback mechanism in case the language is not available (could be done after we merge if nothing is broken)
• need to redirect those to /:locale/url  (could be done after we merge if nothing is broken)


mheap:
@fabianrbz I'd expect that for the default language (`en`) I wouldn't see it in the URL. This would be true for `cn` when visiting `developer.nexmocn.com` too

mheap:
Choosing a tutorial on https://nexmo-developer-i18n.herokuapp.com/use-cases leads to a 404

mheap:
Changing product on https://nexmo-developer-i18n.herokuapp.com/en/client-sdk/tutorials leads to a 404

app/services/tutorial_list.rb
```diff
@@ -28,17 +29,20 @@ def self.tasks_for_product(product)
 
   def self.all
     tasks = []
-    Dir.glob("#{Rails.root}/config/tutorials/*.yml") do |filename|
+    # TODO: make this work with I18n fallback
+    Dir.glob("#{Rails.root}/config/tutorials/en/*.yml") do |filename|
```
mheap: Should this use `I18n.default_locale` rather than `/en/`?

spec/presenters/sidenav_resolver_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+    it 'removes the root, language and extension' do
+      expect(subject.path_to_url('_documentation/en/messaging/sms/overview.md')).to eq('messaging/sms/overview')
+    end
+  end
+
+  describe '#item_navigation_weight' do
+    context 'with given navigation weight' do
+      it 'returns the corresponding one' do
+        item = { is_task?: true, title: 'messages' }
+        expect(subject.item_navigation_weight(item)).to eq(1)
```
mheap: Should we be mocking the call to `config/navigation.yml` so that navigation weight changes don't impact the tests?

app/presenters/title_normalizer.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def initialize(folder)
+    @folder = folder
+  end
+
+  def normalize
+    if @folder[:is_task?] || @folder[:is_tabbed?]
+      @folder[:title]
+    elsif @folder[:is_file?]
+      Tasks.document_meta(@folder[:path])['navigation'] || Tasks.document_meta(@folder[:path])['title']
```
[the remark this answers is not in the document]
fabianrbz: @mheap what do you mean? the check in the `elsif` is for `is_file?`

spec/presenters/sidenav_resolver_spec.rb
fabianrbz: @mheap definitely, can we do that as part of a separate PR?

PR #2053, opened 2019-10-16
Check for `meta_title` and `description` Rake task for documentation

benhayehudi:
## Description

This PR adds a rake task to check for the existence of `meta_title` and `description` in every document in `/_documentation`.

## Deploy Notes

### DIFF
--- a/lib/tasks/documentation.rake
+++ b/lib/tasks/documentation.rake
@@ -3,11 +3,11 @@ namespace :documentation do
   task 'check_keys': :environment do
     documentation_path = "#{Rails.root}/_documentation/**/*.md"
     documents = []
-    
+
     Dir.glob(documentation_path).each do |filename|
       document = YAML.safe_load(File.read(filename))
-      
-      if (!document.key?('meta_title') == false || !document.key?('description') == false)
+
+      if !document.key?('meta_title') == false || !document.key?('description') == false
         documents.push(filename.split('/_documentation')[1])
       end
     end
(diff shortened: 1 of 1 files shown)


lib/tasks/documentation.rake
```diff
@@ -0,0 +1,16 @@
...
+    documents = []
+
+    Dir.glob(documentation_path).each do |filename|
+      document = YAML.safe_load(File.read(filename))
+
+      if !document.key?('meta_title') == false || !document.key?('description') == false
+        documents.push(filename.split('/_documentation')[1])
+      end
+    end
+    puts "The following #{documents.count} documents are missing either a 'meta_title' or 'description' key:\n#{documents.join("\n")}"
```
mheap: I'd expect no output if the check passes successfully, and for this to `raise` if there are any entries in `documents`

lib/tasks/documentation.rake
```diff
@@ -0,0 +1,16 @@
+namespace :documentation do
+  desc 'Verify all pages have meta_title and description keys'
+  task 'check_keys': :environment do
+    documentation_path = "#{Rails.root}/_documentation/**/*.md"
+    documents = []
+
+    Dir.glob(documentation_path).each do |filename|
+      document = YAML.safe_load(File.read(filename))
+
+      if !document.key?('meta_title') == false || !document.key?('description') == false
```
mheap: What does the `false` do here? It looks like a logic error

```
document
=> {"foo"=>"bar"}
```

```
!document.key?('meta_title') == false || !document.key?('description') == false
=> false
```

I'd have expected:

```
if !document.key?('meta_title') || !document.key?('description')
```

However, this wouldn't catch instances when the key is set to a false-y value. The following will handle both situations:

```
if document['meta_title']&.empty? || document['description']&.empty?
  documents.push(filename.split('/_documentation')[1])
end
```

PR #2074, opened 2019-10-18
Remove constraints from renderer.

fabianrbz:
## Description

* From now on, constraints are only valid for NDP and should be updated
here.
* Redirects are up to date with what we had in the renderer.
* Constraints now include everything that was defined in the renderer.

## Deploy Notes

We need to release https://github.com/Nexmo/nexmo-oas-renderer/pull/35 first, and then bump its version here.


fabianrbz:
@mheap I;m not sure how do you want to handle the issue with the api-errors, most of those are for   APIs that are in beta.

mheap:
@fabianrbz What's the issue with `api-errors`?

fabianrbz:
@mheap the check that is failing

fabianrbz:
@mheap I ended up skipping the subaccounts from the rake task as it was still failing. It requires the error to be in the page :/

mheap:
@fabianrbz I'll approve this to keep us moving, but we do need to fix those broken links by adding `x-errors` to the subaccounts YAML

PR #2081, opened 2019-10-21
Linking a Facebook Page to a Nexmo application

pardel:
## Description

Added extra step for linking a Facebook Page to a specific Nexmo application.
This was a note before and easily missed by developers.

## Deploy Notes

No changes

### DIFF
--- a/_partials/reusable/link-facebook-to-nexmo.md
+++ b/_partials/reusable/link-facebook-to-nexmo.md
@@ -1,4 +1,4 @@
-## Link your Facebook Page to your Nexmo account
+## Step 1: Link your Facebook Page to your Nexmo account
 
 Linking your Facebook page to your Nexmo account allows Nexmo to handle inbound messages and enables you to send messages from the Nexmo Messages API.
 
@@ -46,7 +46,26 @@ You will see your Facebook Pages listed.
 
 At this point your Nexmo Account and this Facebook Page are linked.
 
-> **NOTE:** If at some point in the future you want to further link a specific application to this Facebook Page you can now do this from the Dashboard on the External Accounts tab for the specific Messages and Dispatch application you want to link.
+
+## Step 2: Link your Facebook Page to your Nexmo application
+
+Once your Facebook page is linked to your Nexmo account, it becomes available for use by any of your applications. Here is how to link the page to a Nexmo application:
+
+**1.** Navigate to the [Messages and Dispatch applications page](https://dashboard.nexmo.com/messages/applications).
+
+**2.** From the list, select the application you want to link to.
+
+**3.** Then, select the "External Accounts" tab from the top navigation.
+
+**4.** Finally, click the "Link" button beside your Facebook Page (the `Provider` should be `messenger`).
+
+You're now ready to receive messages users send on your Facebook Page.
+
+
+> **NOTE:** If at some point in the future you want to link a different application to this Facebook Page, you can simply redo Step 2 for the new application.
+
+
+<hr />
 
 ## Re-linking your Facebook page to your Nexmo account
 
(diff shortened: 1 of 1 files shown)


_partials/reusable/link-facebook-to-nexmo.md
```diff
@@ -1,4 +1,4 @@
-## Link your Facebook Page to your Nexmo account
+## Step 1: Link your Facebook Page to your Nexmo account
```
tbedford: Please avoid numbering in titles. In this case it confuses matters with other steps on the page.

_partials/reusable/link-facebook-to-nexmo.md
```diff
@@ -46,7 +46,26 @@ You will see your Facebook Pages listed.
 
 At this point your Nexmo Account and this Facebook Page are linked.
 
-> **NOTE:** If at some point in the future you want to further link a specific application to this Facebook Page you can now do this from the Dashboard on the External Accounts tab for the specific Messages and Dispatch application you want to link.
+
+## Step 2: Link your Facebook Page to your Nexmo application
```
tbedford: Remove number. As you are adding an extra step here this could affect the list of steps on the page that uses this partial (see review app for effect). It might be better to put this step in its own partial as it is linking a specific app to Facebook. I would also world it as linking your Nexmo application to your Facebook page rather than the other way around as that better matches what you are actually doing.

_partials/reusable/link-facebook-to-nexmo.md
```diff
@@ -46,7 +46,26 @@ You will see your Facebook Pages listed.
...
+
+**4.** Finally, click the "Link" button beside your Facebook Page (the `Provider` should be `messenger`).
+
+You're now ready to receive messages users send on your Facebook Page.
+
+
+> **NOTE:** If at some point in the future you want to link a different application to this Facebook Page, you can simply redo Step 2 for the new application.
+
+
+<hr />
```
tbedford: Remove this HTML.

_partials/reusable/link-facebook-to-nexmo.md
pardel: These 2 parts need to stay together for completeness - doing only Part 1 won't set up the FB account properly and won't be usable.
The numbering seems to be ok: https://nexmo-developer-pr-2081.herokuapp.com/messages/concepts/facebook

As for the direction, the operation happens in the context of a Nexmo application so "adding FB page to Nexmo app" seems to make more sense:

Screenshot 2019-10-22 at 12 36 54

tbedford: Yes, it's confusing, but linking your Nexmo **account** to your Facebook page is a stand-alone operation. You can do it even if you've not created any applications at all - any JWT can be used to validate the process. You are just linking the accounts. The problem with your approach is it assumes an application has been created, which it might not have been. Having said that, in the interests of getting something merged for Campus, I would think delineating as Part 1 and Part2 would be acceptable for now. I will come back and take another look later - I have to convert many of the things using partials to the new step-by-steps anyway.

_partials/reusable/link-facebook-to-nexmo.md
```diff
@@ -46,7 +46,25 @@ You will see your Facebook Pages listed.
...
+**2.** From the list, select the application you want to link to.
+
+**3.** Then, select the "External Accounts" tab from the top navigation.
+
+**4.** Finally, click the "Link" button beside your Facebook Page (the `Provider` should be `messenger`).
+
+You're now ready to receive messages users send on your Facebook Page.
+
+
+> **NOTE:** If at some point in the future you want to link a different application to this Facebook Page, you can simply redo Step 2 for the new application.
```
tbedford: Part 2 to avoid confusion with step 2.

pardel: Oh, sorry. missed that one.

PR #2083, opened 2019-10-21
Add folks to teams page

slorello89:
## Description

Adding Dwane, Chris, Michael, and I to the team page

## Deploy Notes


config/team.yml
```diff
@@ -89,4 +89,20 @@ current:
     email: 'bethany.loft@vonage.com'
     twitter: 'bb_loft'
     role: 'Junior Community Event Manager'
+  - name: 'Steve Lorello'
+    email: 'steve.lorello@vonage.com'
+    twitter: 'slorello'
+    role: '.NET Developer Advocate'
+  - name: 'Chris Tankersley'
+    email: 'chris.tankersley@vonage.com'
```
tbedford: It looks like Chris and Dwane don't have avatars co figured. We use Gravatar. In Gravatar you need to specify the avatar based on email address if memory serves (been a while).

slorello89: @dragonmantank  - @conshus can you update this comment here when you've configured yoru gravatars?

conshus: Had to upload a picture from another computer to get it working. Let me know if it doesn't work.
image

dragonmantank: Capture

I have one set up? Maybe it's a caching issue?

slorello89: @tbedford Looks like they should be set now - can you check again?

thanks,

PR #2087, opened 2019-10-22
Update JS Release Notes for v6.0.2

dimitrisniras:


tbedford:
Looks like there's one more:

```
25 | ved storage of JWT token from localStorage to `NexmoClient` config objec 
```

localStorage -> `localStorage` (variable always need to be in back ticks to avoid spelling errors)

PR #2094, opened 2019-10-22
VNI-229: Add new Verify workflows 6 and 7

AFavreNexmo:
## Description

* Add Verify workflow_id 6 and 7


lornajane:
Thanks this is a great start! Closing since #2109 builds on this and will include a review app and some other recent changes

PR #2101, opened 2019-10-23
Fix incorrect git clone command

marklewin:
## Description

The `git clone` command in the Client SDK Click to Call use case was incorrect. Fixed.


tbedford:
@marklewin something pretty disastrous seems to have happened in the build. I guess you could try rebuilding?

marklewin:
> @marklewin something pretty disastrous seems to have happened in the build. I guess you could try rebuilding?

I think there's a Travis problem. Triggered the rebuild but still failing.

fabianrbz:
@marklewin @tbedford fixed! (it was a caching issue with travis)

PR #2109, opened 2019-10-25
Add documentation for new Verify Workflows

lornajane:
## Description

We have new workflow IDs in the Verify API. This adds the documentation for the new workflows and replaces #2094 because I had to rebase to get some other changes in context.


lornajane:
The problems I saw with this feature were related to a routing problem; the feature itself seems OK to some numbers (just weirdly not my work mobile!)

PR #2111, opened 2019-10-25
API Specification Update

NexmoDev:


marklewin:
Marking as don't merge while @Too-Funky has a look at the review app.

marklewin:
Cancelling PR as more changes required to OAS.

PR #2113, opened 2019-10-25
API Specification Update

NexmoDev:


marklewin:
This is for the soft deprecation of `messageCallbackType` parameter in the Numbers API as suggested by Brennan Topley.

PR #2119, opened 2019-10-29
[DEVX-1375] Populate OAS constraints from files.

fabianrbz:
## Description

Remove the hardcoded list of constraints and use the files inside
OAS_PATH to populate them.


mheap:
Is this something that we want to do? Being able to have OAS docs merged in to master but not showing in NDP feels like a useful feature

On the other hand, maintaining this list is a burden and I can see a lot of arguments for removing it

@lornajane @tbedford @marklewin Any thoughts?

fabianrbz:
i thought about that and we could flag the specs that we don't want to display with a flag in a frontmatter, or we could keep them in a separate folder. otherwise, we can just keep the list.
what are your thoughts?

mheap:
A flag in the OAS itself could work? `x-show-ndp: false` at the top level or similar?

mheap:
We've had this exact problem with reporting. We wanted it on `master` on the API specifications repo and showing in a review app, but not on production

The `x-show-ndp` solution wouldn't work for this use case (so back to the drawing board), but it is a real world use case.

Whether we mind showing an API that is in EA on production NDP or not is another issue. What are the pros/cons to doing so?

lornajane:
for the specific EA reporting API use case, either it's public enough to be in master in which case it shows in the repo and would render on NDP if you guessed the URL correctly. Or it isn't, in which case it can stay in a branch and we can use that branch in the submodule when we pull into EA. I am not sure what master-but-not-NDP looks like since our specs are public anyway but I don't feel strongly about it if you feel it's a problem that needs solving!

mheap:
I'm not sure it's a big enough problem to need solving. 

The use case here was to have an always-online OAS (hence being in master) available at http://nexmo-api-specification.herokuapp.com/ for the team to reference, but it wasn't yet supposed to be in NDP.

fabianrbz:
what about having a staging branch for `api-specification` used for http://nexmo-api-specification.herokuapp.com/ ? That way we can show whatever is in master in NDP

mheap:
That could work, but we've got a lot of automation set up based on the assumption that `master` is live (e.g. automated changelogs).

I'm back to either having an allowlist in the NDP constraint, or populating it from the list in the `definitions` folder

mheap:
I'm going to close this. We can revisit in the future if it becomes an issue

Issue #2015, opened 2019-09-27
Inaccessible tabs


lornajane:
## Description

I'm no longer able to read the documentation on the NCCO Guide page. There are some tabs for "Single Option", "Single Value" and so on, but these seem to be some sort of special span tag, they aren't focusable or interactive.

The same applies to the language tabs (but only on this page, the code snippets work well in general for me at this point).

## What did you expect to happen?

I expected to be able to interact with the user interface

## What actually happened?

Nothing. I can see the tabs, but they're just display, they're not buttons or links or anything else I can focus.

## Steps to reproduce:

Go to https://developer.nexmo.com/voice/voice-api/guides/ncco and try using something that isn't a pointing device to access the tabs.

Please always let me know if you would like me to look at the tabs in any NDP changes, I'm always happy to! Right now, tabs in a couple of places and not being able to scroll the nav bar at all are reasons that I can't use some parts of the documentation and this might exclude our customers as well as our colleagues.

brkn:
I would like to work on this.

PR #2124, opened 2019-10-30
Add accessibilit to tabs

fabianrbz:
## Description

* Updates volta to include a11y features
* Highlights the selected tab in pink just to make it obvious and we can test it (we need to do something else here)
* Adds `role=tab` to tabs

## Note: experimental for now


lornajane:
Oh this is helpful for what I reported in #2015 - I can read the NCCO reference now!

fabianrbz:
great!
@lornajane @mheap I changed the colour to `lighter-blue`, do you think that it is necessary or should I just drop the highlighting?

mheap:
@fabianrbz I see the light blue when focusing with a mouse, but not when using tab to select the item. Am I supposed to see it on the tab?

fabianrbz:
@mheap I ended up removing the colour on focus, so it should use volta's styles

lornajane:
I've approved but it looks like we're pointing to a branch in one of the dependencies so I wasn't sure if I should merge.

fabianrbz:
@lornajane yes, we need to merge https://github.com/Nexmo/nexmo-oas-renderer/pull/40 first.
Can you look at that one?

lornajane:
Great. Approved and merged over there but I assume you need to roll releases and update over here ... let me know if you need me to do anything else?

fabianrbz:
I'll take it from here, thanks!

PR #2133, opened 2019-11-05
Cache sidenav

fabianrbz:
## Description

* Cache the sidenav to ease the memory consumption
* Move the logic of marking the active link in the sidenav to the frontend

## Note:
We need to add memcachier plugin to heroku, it has a free plan that I think will work for us. 

EDIT: looks like we use redis for caching, so we can use that one,

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
-release: bundle exec rake db:migrate && bundle exec rake tmp:clear
+release: bundle exec rake db:migrate && bundle exec rake cache:clear
 web: rails s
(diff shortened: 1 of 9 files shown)


Procfile
```diff
@@ -1,2 +1,2 @@
-release: bundle exec rake db:migrate
+release: bundle exec rake db:migrate && bundle exec rake tmp:clear
```
mheap: It looks as though this will only clear caches that are stored on disk, not in Redis

https://github.com/rails/rails/blob/ef5d85709d346e55827e88f53430a2cbe1e5fb9e/railties/lib/rails/tasks/tmp.rake#L25-L30

fabianrbz: you are right, fixed!

PR #2138, opened 2019-11-06
Move ip-messaging tasks under a subfolder.

fabianrbz:
## Description

Looks like we already have support for this, we just need to rename the tasks and create the corresponding folders.

@marklewin is something like this what you had in mind?  

CC @tbedford


tbedford:
Yes, looks like it works! :)

PR #2144, opened 2019-11-08
Update ncco-reference.md

franziz:
## Description

The explanation for `stream` is incorrect. It explains `talk` action. So, I change the work `talk` to `stream`.


mheap:
Thanks for this @franziz 🙌

PR #2174, opened 2019-11-14
[DEVX-1601] Add visitor id to logs

fabianrbz:
## Description

Set a random 8 character alphanumeric value `visitor_id` and store it in the session.
Use `visitor_id` as part of the logs so we can follow the user's journey through the site.

### DIFF
--- a/app/middleware/nexmo_developer/visitor_id.rb
+++ b/app/middleware/nexmo_developer/visitor_id.rb
@@ -1,0 +1,13 @@
+module NexmoDeveloper
+  class VisitorId
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      req = Rack::Request.new(env)
+      req.session['visitor_id'] ||= "G-#{SecureRandom.alphanumeric(8)}"
+      @app.call(env)
+    end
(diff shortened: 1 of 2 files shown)


fabianrbz:
@mheap we need to do the same in the `renderer` what do you think about a middleware? It might be too much though

mheap:
Middleware sounds good to me. It'd need to be SSO aware in the future, so
perhaps just read from the session and allow NDP to handle populating the
session?

app/middleware/nexmo_developer/visitor_id.rb
```diff
@@ -0,0 +1,13 @@
+module NexmoDeveloper
+  class VisitorId
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      req = Rack::Request.new(env)
+      req.session['visitor_id'] ||= SecureRandom.alphanumeric(8)
```
mheap: It looks like we lose the `G` prefix here.

```ruby
req.session['visitor_id'] ||= "G#{SecureRandom.alphanumeric(8)}"
```

PR #2199, opened 2019-11-20
Reorganize Client SDK tutorial material into sub-folder

tbedford:
## Description

This PR reorganizes tutorial material into sub-folders. This avoid proliferation of non-generic files in the `client-sdk` folder. Some minor changes to titles are also included to avoid use of acronym (PSTN) that might not be widely known.

## Review

1. Go to review app.
2. Go to Client SDK tutorials.
3. Click through tutorials to make sure pages are present and links work.


tbedford:
@pardel FYI - we have some generic topics here, and then some specific to the tutorial, which are kept in their own folder.

config/tutorials/en/phone-to-app.yml
```diff
@@ -30,9 +30,9 @@ tasks:
 - link-nexmo-number
 - client-sdk/create-user
 - client-sdk/generate-jwt
-- client-sdk/create-webhook-server-phone-to-app
-- client-sdk/phone-to-app-code
-- client-sdk/run-phone-to-app
+- client-sdk/phone-to-app/create-webhook-server-phone-to-app
```
fabianrbz: same here

config/tutorials/en/app-to-phone.yml
```diff
@@ -31,9 +31,9 @@ tasks:
 - client-sdk/create-application
 - client-sdk/create-user
 - client-sdk/generate-jwt
-- client-sdk/create-webhook-server-app-to-phone
-- client-sdk/app-to-phone-code
-- client-sdk/run-app-to-phone
+- client-sdk/app-to-phone/create-webhook-server-app-to-phone
```
fabianrbz: @tbedford do we still need to namespace the files?
Given that they are under `app-to-phone/` we could rename them to:
* `create-webhook-server`
* `code`
* `run`

What do you think?

tbedford: Yes I think it's better with the simpler names. I will fix.

PR #2218, opened 2019-11-26

NexmoDev:
## Description

* Auto-generated PR created when Chris updated PHP code snippet repo. 
* Also adds the config to wire in the new snippets.

### DIFF
--- a/_examples/verify/search-verification-request/php.yml
+++ b/_examples/verify/search-verification-request/php.yml
@@ -1,0 +1,15 @@
+---
+title: PHP
+language: php
+dependencies:
+    - nexmo/client
+client:
+    source: .repos/nexmo/nexmo-php-code-snippets/verify/search.php
+    from_line: 5
+    to_line: 6
+code:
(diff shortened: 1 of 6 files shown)


tbedford:
@lornajane as Chris is away for a while now would you be able to do a quick review please? I wired in the missing config, but not 100% on whether my line numbers are suitable. Thanks.

_examples/verify/search-verification-request/php.yml
```diff
@@ -0,0 +1,15 @@
+---
+title: PHP
```
benhayehudi: I don't think defining `title`, `language` is necessary here anymore as it's implicitly configured based on the file endings. I'm also not sure we need the `nexmo/client` as an explicit dependency. Unless it's something besides the SDK, i.e. a 3rd party dependency, I don't think we need to specify the SDK as a dependency anymore.

tbedford: Thanks Ben, looks like you are right and you don't need that config any more. Removed.

PR #2221, opened 2019-11-27
Refactor search terms

fabianrbz:
## Description
Refactor SearchTerms generation and allow to choose which locale to use for the content

### DIFF
--- a/app/services/search/article.rb
+++ b/app/services/search/article.rb
@@ -1,0 +1,39 @@
+module Search
+  class Article
+    delegate :config, :title, :description, :body, :path, :doc_type, to: :@doc
+
+    def initialize(doc, html)
+      @doc  = doc
+      @html = html
+    end
+
+    def relative_path
+      @relative_path ||= "#{config[:base_url_path]}/#{path.relative_path_from(config[:origin])}".gsub('.md', '')
+    end
+
+    def product
+      @product ||= begin
+                     product = relative_path.split('/')[1]
+                     if product == 'messaging'
+                       "#{relative_path.split('/')[1]} > #{relative_path.split('/')[2]}"
+                     else
+                       product
+                     end
+                   end
+    end
+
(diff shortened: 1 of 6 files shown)


app/services/search/article.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    def relative_path
+      @relative_path ||= "#{config[:base_url_path]}/#{path.relative_path_from(config[:origin])}".gsub('.md', '')
+    end
+
+    def product
+      @product ||= begin
+                     product = relative_path.split('/')[1]
+                     if product == 'messaging'
+                       "#{relative_path.split('/')[1]} > #{relative_path.split('/')[2]}"
```
benhayehudi: I am assuming we are assuming the file path will remain like this for the foreseeable future?

spec/services/search/article_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'rails_helper'
+
+RSpec.describe Search::Article do
+  let(:doc_type) { 'documentation' }
+  let(:path) { '_documentation/en/concepts/overview.md' }
```
benhayehudi: Should we also add a test for when the locale is not the default locale?

app/services/search/article.rb
fabianrbz: not really, we update the index every week, so it shouldn't be an issue

spec/services/search/article_spec.rb
fabianrbz: we don't have any file in a different language yet, and also we are mocking `File.read`.
But i can test one with `en` and the other with `cn`

benhayehudi: Since the goal is to ensure compatibility across locales, thought it might be a good idea. At the very least, I know you thought about it! I'll leave it up to you, and approve the PR either way.

PR #2203, opened 2019-11-20
Fix Nexmo Client SDK Documentation

vonageviollete:
## Description
Fix the client SDK documentation for Nexmo.

PR #2225, opened 2019-11-28
Fix sdk docs

fabianrbz:
## Description

Taken from https://github.com/Nexmo/nexmo-developer/pull/2203/


tbedford:
@fabianrbz - thanks for fixing this!

PR #2231, opened 2019-11-29
[DEVX-1882] Add new Management APIs tile.

fabianrbz:
## Description
[Jira ticket](https://nexmoinc.atlassian.net/browse/DEVX-1882)

Create a "proper" tile with the Global and Developer APIs

### DIFF
--- a/app/views/static/_products.html.erb
+++ b/app/views/static/_products.html.erb
@@ -280,26 +280,28 @@
     </div>
   </div>
 
-  <div class="Vlt-col Vlt-col--1of3 Vlt-col--M-1of2 Nxd-col-stretch">
-    <div class="Vlt-card Vlt-article--reverse Vlt-card--center">
-
-      <nav class="Vlt-center" style="width:100%;">
-        <h3 class="Vlt-grey-dark">Developer API</h3>
-        <a href="/api/developer/messages">Messages</a> |
-        <a href="/api/developer/numbers">Numbers</a> |
-        <a href="/api/developer/pricing">Pricing</a>
-      </nav>
-      <hr class="hr--short">
-      <nav class="Vlt-center" style="width:100%;">
-        <h3 class="Vlt-grey-dark">Global APIs</h3>
-        <a href="/api/account">Account</a> |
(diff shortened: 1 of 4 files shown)


app/views/static/_products.html.erb
```diff
@@ -280,26 +280,27 @@
...
+      <div class="Vlt-card__image Vlt-card__image--app Vlt-card__image--green-light">
+        <div class="Vlt-card__image__icon">
+          <svg>
+            <use xlink:href="/symbol/volta-icons.svg#Vlt-icon-laptop"/></svg>
+        </div>
+      </div>
+      <div class="Vlt-card__content">
+        <h2>Management APIs and Services</h2>
+        <nav>
+          <a href="/api/developer/messages">Messages</a>
```
tbedford: This is actually probably better termed "Message Search" rather than "Messages".

mheap:
Screen Shot 2019-11-29 at 13 30 21

Is there a way to keep it the same height as all of the other tiles?

tbedford:
> Is there a way to keep it the same height as all of the other tiles?

Possibly related - could it just be "Management APIs" and lose the "and Services."

tbedford:
Also, I just noticed, above "Account" there is some small text that says "Management". The box on the right does not have that, and also does that mean Account should be under the Management APIs box?

mheap:
Account can be dropped from the Management APIs tile

fabianrbz:
I pushed some changes, the tiles now have the same height.
@tbedford the title was a specific requirement, but we can talk about it in the [Jira ticket](https://nexmoinc.atlassian.net/browse/DEVX-1882)

mheap:
The other tiles look like they're an odd height now with whitespace under `API Reference`. We can change this tile to just `Management APIs` if it helps

fabianrbz:
did some changes, now the tiles have all the same height and it's almost the one they had before

fabianrbz:
@tbedford can you take another look at this?

tbedford:
Still think it's slightly odd that Account is in a management tile and there's a  management API in the next tile over. But....that's another issue...

PR #2273, opened 2019-12-13
Update member.md to reflect Member limit

conshus:
## Description

Adds a note that the limit of Members in a Conversation is 100.


_documentation/en/conversation/concepts/member.md
```diff
@@ -23,3 +23,5 @@ MEM-df8e57d8-1c8e-4573-bf4d-29d5414dcb42
 ```
 
 Memberships can be created in various ways: [via the Conversation API](/conversation/code-snippets/member/create-member), using the functions provided by the [Client SDK](/client-sdk/overview), or using the command line.
+
+> **NOTE:** There is a limit of 100 Members that can be added to a Conversation.
```
marklewin: Better to use active rather than passive voice where possible. How about "You can add a maximum of 100 Members to a Conversation." ?

conshus:
> You can add a maximum of 100 Members to a Conversation.
Updated.
Thanks for the article. Good info. I've got it bookmarked so I can refer to it in the future.

PR #2289, opened 2019-12-18
CodeSnippetFilter spec

benhayehudi:
## Description

This PR adds testing for the previously untested `CodeSnippetFilter`, which generates each individual code snippet on NDP.

### DIFF
--- a/spec/filters/code_snippet_filter_spec.rb
+++ b/spec/filters/code_snippet_filter_spec.rb
@@ -109,6 +109,7 @@ RSpec.describe CodeSnippetFilter do
 
   it 'returns code snippet with application and client configuration instructions and dependencies if they are defined' do
     allow(File).to receive(:exist?).and_call_original
+    expect(SecureRandom).to receive(:hex).at_least(:once).and_return('ID123456')
     expect(File).to receive(:exist?).with(/example_snippet/).and_return(true).twice
     expect(File).to receive(:read).with(/example_snippet/).and_return(example_source_file).twice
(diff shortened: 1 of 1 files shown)


lornajane:
I can't tell from the pull request description but does this fix the missing code snippets on the overview pages? I was just about to open an issue when I saw this come in!

benhayehudi:
> I can't tell from the pull request description but does this fix the missing code snippets on the overview pages? I was just about to open an issue when I saw this come in!

Oops, sorry about forgetting to update the description, just corrected that! This adds testing for the filter on NDP, which previously had no testing associated with it.

spec/filters/code_snippet_filter_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+    output = described_class.call(input)
+
+    expect(output).to include(expected_app_name)
+    expect(output).to include(expected_base_url)
+    expect(output).to include(expected_events_url)
+    expect(output).to include(expected_answer_url)
+  end
+end
+
+def example_source_file
```
fabianrbz: @benhayehudi can you define a `let` inside the describe instead of this method?
You can move this method inside the describe as well, we just want to scope it to this spec.

benhayehudi: @fabianrbz, moved the method into the describe. Let me know what you think. Thanks!

fabianrbz: @benhayehudi can you change it to a [let](https://relishapp.com/rspec/rspec-core/v/3-9/docs/helper-methods/let-and-let)?

PR #2290, opened 2019-12-18
Async Trigger and Callback PHP Examples

NexmoDev:
Adds async trigger and callback examples for PHP

### DIFF
--- a/_examples/number-insight/async-callback/php.yml
+++ b/_examples/number-insight/async-callback/php.yml
@@ -1,0 +1,5 @@
+---
+code:
+    source: .repos/nexmo/nexmo-php-code-snippets/number-insights/advanced-async-callback/index.php
+    from_line: 9
+    to_line: 17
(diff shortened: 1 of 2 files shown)


lornajane:
Tagging @dragonmantank to review as I guess he may already have a branch of matching changes

dragonmantank:
Thanks @tbedford, noting. This quarter all the server SDKs are cleaning up their code snippets, so I'll make sure that gets put on the list. The basic example should be two blocks. I suspect that is just due to the age of that code example.

Is there a better way to help denote when we intentionally do no have a client block, like the async callback example? There is no client involved in that example since it's just reading a request body.

tbedford:
Thanks @dragonmantank. At the moment there's no way to denote that for async callback example and similar, other than to leave out the client section, but we will be revamping code snippet infrastructure in due course. I will make a note to address that.

PR #2310, opened 2020-01-02
Add code snippets for Verify API PSD2 feature

lornajane:
## Description

Adding the code snippets for the PSD2 feature (just waiting a final list of languages before we add this to the API reference so getting some code samples ready to go).

WIP because the spec isn't live but also I've broken something! My local says it can't find the file because it's not looking in the `en/` folder and I'm not sure what to do about that ...


lornajane:
This has been open too long, I'll probably have to start again anyway.

PR #2316, opened 2020-01-07
Experimental PR

tbedford:
## Description

Initial look at guides that have been requested for Client SDK.

This PR has been created to clarify the direction we wish to proceed in - I will do the real work in another PR. 

In this PR: 

1. [Send/Receive DTMF - unchanged from what was](https://nexmo-developer-pr-2316.herokuapp.com/client-sdk/in-app-voice/guides/send-and-receive-dtmf)
2. [Mute - Uses tabbed content, NOT tabbed folder](https://nexmo-developer-pr-2316.herokuapp.com/client-sdk/in-app-voice/guides/mute)
3. [Make a Call - uses tabbed folder](https://nexmo-developer-pr-2316.herokuapp.com/client-sdk/in-app-voice/guides/make-call) 
4. Use Code Snippets infrastructure (yeah, we've been here before but thought I'd mention it - I still think it's valid).

Which option do we prefer? (p.s. 1 and 2 are essentially the same)

I would also do a refactoring of the relevant tabbed content as it could be better organized.


tbedford:
Check redirects failing due to GitHub Action bug. WIP.

pardel:
I prefer 1 to 3. I think having all the languages listed works better - we only have 5 so looks good as well.
3 has issues with the tabs inside folders - eg. can't point straight to Obj-C version.

pardel:
We should also set some guidelines on what constitutes a 'Guide', what a 'Tutorial' and what goes under 'Getting Started'.
ATM, looks like 'Guides' are simple actions (send DTMF, mute). However, the line is very blurry between the other 2 categories.

tbedford:
@pardel yes agree 1 was my preferred option too, particularly as you then don't need to duplicate any boiler plate text. 

With regards guides/tutorials/use cases - exactly why I wanted to avoid guides and focus on tutorials. I prefer to have concepts for conceptual material, code snippets for code building blocks, and then tutorials to provide step-by-step instructions on how to do specific things. For Client SDK code snippets were deemed a no-no, so we are going with guides instead and the idea is to provide more context than you'd usually find in a code snippet.

I have on my backlog to update the contribution guide - I think I will make that a priority. In the meanwhile here's a brief summary for Client SDK:

* Guide - somewhere between a tutorial and a code snippet. Sometimes you want to provide code + more explanation and context and this is where the Guide comes in. You are explaining how to do a finite specific thing. Kind of a *special case for Client SDK* as we really want Code Snippets, but are using Guides to allow for more context. 

* Tutorial - for longer form *step-by-step* how-to. If a guide is getting too long, or describing a more complex procedure, it should probably be a step-by-step. Tutorials are generally single product. 

* Use Cases - Currently we are using these for more complex "use cases" potentially involving multiple products. For example: Sendinblue use case uses Conversation API, Client SDK and Sendinblue API. 

With regards the Getting Started section that was something Britt specifically wanted for clients, happy to revisit it though. Some of the material is covered by tutorials (for JS), we can also look at converting/moving the iOS/Android flavours. Personally I would like to do away with that section and just have Concepts/Guides/Tutorials and Use Cases.

mheap:
@tbedford @pardel With the changes that Paul made last week, is this PR required?

pardel:
@mheap @tbedford This will probably have lots of conflicts once rebased.

tbedford:
I think we can close this now.

PR #2323, opened 2020-01-09
Removed highly specific implementation details from Team feature spec

benhayehudi:
As a follow-up to yesterday's refactor of the test to count the number of team members precisely, this PR removes a lot of the expectations on exact wording within the Team view. We should not be testing for the presence of the word "and", for example.

### DIFF
--- a/spec/features/team_spec.rb
+++ b/spec/features/team_spec.rb
@@ -19,36 +19,11 @@ RSpec.feature 'Team' do
 
     within('.Nxd-landing-main') do
       expect(page).to have_css('h1', text: 'Nexmo Developer Relations Team')
-      expect(page).to have_css('p:nth-of-type(1)', text: 'Our mission is to build a world-class open source documentation platform to help developers build connected products.')
-
-      within('p:nth-of-type(2)') do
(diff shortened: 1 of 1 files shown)


spec/features/team_spec.rb
```diff
@@ -19,36 +19,11 @@
...
       expect(page).to have_link('Contributors', href: 'https://github.com/Nexmo/nexmo-developer/graphs/contributors')
-
-      within('.Vlt-grid:nth-of-type(2)') do
-        within('.Vlt-col:nth-of-type(1)') do
-          expect(page).to have_link(href: '/spotlight')
-        end
-        within('.Vlt-col--3of4') do
-          expect(page).to have_css('h2', text: 'Write for Nexmo')
-          expect(page).to have_content('You don\'t have to be part of the team to write for Nexmo.')
-          expect(page).to have_link('Developer Spotlight', href: '/spotlight')
```
fabianrbz: I would leave the check for the spotlight link

benhayehudi: I took it out because it's a reference to a specific team initiative, which is more likely to change over time than text like "Join the team". It seemed too temporal for me to be part of a long-term test. What do you think?

PR #2336, opened 2020-01-16
Add docs to vbc authentication to append `vbc.prod` to username when using password grant

tbass134:
## Description

When using password grant to login using VBC API's, the domain (`@vpc.prod`) needs to be appended to the user name


sammachin:
I've just tested this and confirm the change works and is nessacery

tbass134:
@mheap its in the /.repos folder.

PR #2343, opened 2020-01-21
API Specification Update

NexmoDev:


tbedford:
@niati : I will meet up with you to explain how we deal with these.

PR #2255, opened 2019-12-09
DEVX-1863: Allow locale in url

fabianrbz:
## Description

Allow locale in url and fallback to `en`.

## Deploy Notes

We need to release `nexmo-markdown-renderer` before merging this.

PR #2344, opened 2020-01-21
Remove MarkdownPipeline from NDP

benhayehudi:
Connected to the publication of https://github.com/Nexmo/nexmo-markdown-renderer, remove the markdown pipeline from NDP and require the the above mentioned markdown renderer gem.


benhayehudi:
@fabianrbz - receiving the following error when attempting to start NDP with the gem locally:

```
/Users/bgreenberg2/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/activesupport-6.0.2.1/lib/active_support/evented_file_update_checker.rb:170:in `initialize': no implicit conversion of Array into String (TypeError)
```

It's referencing this part of `evented_file_update_checker`:

```
 class PathHelper
   def xpath(path)
    Pathname.new(path).expand_path
  end
...
```

mheap:
@benhayehudi Looks like there are some conflicts to be resolved here

benhayehudi:
Yes... there are some changes to the markdown filters also in progress in #2255 that conflict with what's in progress in the abstraction of the markdown pipeline in https://github.com/Nexmo/nexmo-markdown-renderer. 

The more I think about it, I think we ought to close this PR, finish the work on the new markdown renderer, and then re-open it with that finished work incorporated first.

fabianrbz:
@benhayehudi yeah, let's close this one until the pipeline is ready.

PR #2373, opened 2020-01-30
Edit Private Communication Use Case

marklewin:
## Description

- [x] Rewriting and restructuring to make it more consistent with our existing use cases.
- [x] Fix Mermaid diagrams that weren't displaying
- [x] Check code and webhook URLs
- [x] Make code available in GitHub repo
- [x] Link to GitHub repo from use case

## Review

Dwane, would you be able to review Phil's code please.


tbedford:
@niati - could you have a look through please as a final check? Dwane has updated and tested the code, so just looking for any obvious mistakes/typos at this point. Thanks!

PR #2374, opened 2020-01-30
Add Localization plus link to WhatsApp docs.

tbedford:
## Description

See DEVX-1400.

## Review

- [ ] [WhatsApp concept topic]()

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -325,3 +325,4 @@ DocStrap
 whitelisting
 unmuting
 earmuffing
+Localizable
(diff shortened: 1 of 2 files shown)


tbedford:
@hughhopkins is there anything more we should say here?

hughhopkins:
I'm not a fan of this feature from WhatsApp. It’s a bit like fallback locale. It removes control from the developer to be sure what is shown to the end user. 

It does not work in our abstracted template message format. It does work in the custom object but given it's complexity I do not advise this feature at all.

tbedford:
Closing based on Hugh's comments.

PR #2393, opened 2020-02-03
Add support for multi-language tutorials.

fabianrbz:
## Description

Prototype for adding multi-language tutorials.

The way it works is we define an entry point for each tutorial with some metadata (description and product) under `config/tutorials/en/`, e.g.

```
config/tutorials/en/client-sdk/phone-to-app.yml
```

and the different code languages should live in a folder under the same name:

```
config/
  tutorials/
    en/
      client-sdk/
        phone-to-app/
          javascript.yml
          python.yml          
```

the different steps are defined as before in `_tutorials/en/client-sdk` with the caveat that now we can have shared steps across different code languages or have specific definitions for them, .i.e.

```
_tutorials/en/client-sdk/
  phone-to-app/
    code/ <- specific to each code language
      python.md
      javascript.md
    create-application.md <-shared
    create-conversation.md <- shared
```

This is a WIP, and currently only works for `https://nexmo-developer-pr-2393.herokuapp.com/client-sdk/tutorials/phone-to-app/introduction` but if we are ok with it I'll update the rest of the tutorials.

### Note: 
* We could place the metadata inside the tutorial's folder instead.
* The code language list on the TOC needs some styling, highlight current language, etc.
* The other alternative is using tabs like we currently have for ios


mheap:
I like it! One requested change:

`config/tutorials/en/client-sdk/phone-to-app.md` => `config/tutorials/en/client-sdk/phone-to-app/base.md`

Instead of having a file and a folder named the same, could we have a base config in the folder? It doesn't have to be called `base`, but I'd like to keep it all together.

Also, should these configs live in the `config` folder, or somewhere else? Or should that be in a followup PR?

tbedford:
Has the PR description been updated to reflect current state of play? I thought we were planning to keep the config in the same directory as the content so everything was in one place?

fabianrbz:
> I like it! One requested change:
> 
> `config/tutorials/en/client-sdk/phone-to-app.md` => `config/tutorials/en/client-sdk/phone-to-app/base.md`
> 
> Instead of having a file and a folder named the same, could we have a base config in the folder? It doesn't have to be called `base`, but I'd like to keep it all together.
> 
> Also, should these configs live in the `config` folder, or somewhere else? Or should that be in a followup PR?

I'll add the `base.md` and yes, the move from `config` should be done in a follow up PR.



> Has the PR description been updated to reflect current state of play? I thought we were planning to keep the config in the same directory as the content so everything was in one place?

It still matches the current state of the code, see above.

mheap:
@fabianrbz I'm bailing on this merge and handing it back to you! I don't have the required context to make the right choices

fabianrbz:
@mheap rebased done!

PR #2397, opened 2020-02-03
Add Conversion API Reference to TOC

tbedford:
## Description

Adds API Reference to sidebar for Conversion. Adds stub file and redirect.

## Review

In [review app](https://nexmo-developer-pr-2397.herokuapp.com/documentation) open up sidebar and navigate down to Messaging and Conversion. API entry should now be present in sidebar. Click on this to be taken to `/api/conversion`.

### DIFF
--- a/_documentation/en/messaging/conversion-api/api-reference.md
+++ b/_documentation/en/messaging/conversion-api/api-reference.md
@@ -1,0 +1,8 @@
+---
+title: API Reference
+---
+
+# API Reference
+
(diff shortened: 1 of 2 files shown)


tbedford:
The reason I didn't do it for US Short Codes is because it's a bit of a hack involving stub files and redirects. The reason is there are two APIs to reference, rather than one in this case. What I would prefer is to address that under a separate PR if that's OK. I will raise a ticket for that work.

PR #2410, opened 2020-02-04
Add US Short Code API to sidebar

tbedford:
## Description

Add US Short Code API References to sidebar.

## Review

Go to [review app](https://nexmo-developer-pr-2410.herokuapp.com/documentation)

Check Messaging / US Short Codes. Present in Sidebar are the following:

1. Two factor API Reference
2. Alerts subscription API Reference
3. Alerts sending API Reference

### DIFF
--- a/_documentation/en/messaging/us-short-codes/alerts-sending-api.md
+++ b/_documentation/en/messaging/us-short-codes/alerts-sending-api.md
@@ -1,0 +1,8 @@
+---
+title: Alerts - Sending API Reference
+---
+
+# Alerts - Sending API Reference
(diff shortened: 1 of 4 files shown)


tbedford:
Yes, it is because it's done a different way to how we normally do things, where you get the text colouring automatically. However, I'm now looking into a workaround. Stay tuned!

tbedford:
OK so I was hoping to be able to hack this using a navigation override, but it can't be done currently. You can have a link to one API Reference (in blue), or all three as they are! Sorry about that.

PR #2420, opened 2020-02-06
Added Media API

peter-nexmo:
Added Media API under authentication as only the API Key & Secret in the query string is currently supported, but there is a common misconception that JWT is currently supported.


mheap:
Is the confusion that VAPI uses JWTs to download recordings? https://developer.nexmo.com/voice/voice-api/code-snippets/download-a-recording

peter-nexmo:
Quite possible. And everything else in the Conversation stack uses JWTs. Eventually, Media API should support JWTs, but it doesn't currently in Beta.

PR #2423, opened 2020-02-07
Content Disambiguation

benhayehudi:
## Description

Create a `DOCS_BASE_PATH` variable that points to path for all documentation, and get documentation from that path and not `Rails.root` anymore.


benhayehudi:
@fabianrbz - This is the PR that we discussed we'd create that preserves all the filters inside NDP, and does not introduce the new `nexmo_markdown_renderer`, but simply points NDP to an external documentation path by environment variable and successfully renders it, as the next incremental step.

benhayehudi:
This is good! @fabianrbz, can you review (and if good merge, then i can carry this forward to the next step in our list tomorrow?) 🚀 🙏 🙌

PR #2427, opened 2020-02-10
Update call-flow.md

JamieChapman8462:
## Description

Added the notify event to the events.


_documentation/en/voice/voice-api/guides/call-flow.md
```diff
@@ -55,6 +55,7 @@ The following table shows possible values for the `status` field of an event obj
 | `timeout`    | The call timed out before it was answered|
 | `transfer`    | The call has been transferred to a new conversation
 | `unanswered` | The recipient handset was unreachable, or the recipient declined the call | 
+| `notify` | An event to signify we have reached the `notify` action in the NCCO |
```
tbedford: > An event to signify we have reached the `notify` action in the NCCO

Who is 'we'? Perhaps reword as:

An event to signify that processing has reached the `notify` action in the NCCO.

PR #2434, opened 2020-02-12
Fixed the PHP JWT example to something that would work

dragonmantank:
## Description

The original PHP JWT code wasn't valid code, and would never have generated a JWT with our library. This shows a way to generate a JWT which should actually go back at least part way on the 1.x branch even.

## Deploy Notes

There are no additional deploy steps other than build and push.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -325,3 +325,4 @@ DocStrap
 whitelisting
 unmuting
 earmuffing
+Keypair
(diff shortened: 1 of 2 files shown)


fabianrbz:
@dragonmantank looks like the tests are failing because of a spelling error. The spellchecker doesn't like `Keypair` for some reason. Can you add it to `.spelling`?

mheap:
@tbedford @dragonmantank  It looks as though this PR removes information (generating an ACL). Could we issue another PR to put it back in?

dragonmantank:
Will do

PR #2436, opened 2020-02-12
Add new env variable DOCS_BASE_PATH to .env.example and docker-compose.

fabianrbz:
## Description

Add new env variable DOCS_BASE_PATH to .env.example and docker-compose.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,6 +20,8 @@ NEXMO_PASSWORD=abc123
 # Required for nexmo-oas-renderer
 OAS_PATH='_open_api/api_specs/definitions'
 
+DOCS_BASE_PATH='.'
+
 GREENHOUSE_API_TOKEN='greenhouse-api-token'
 
 REDIS_URL=redis://localhost:6379/0
--- a/config/application.rb
+++ b/config/application.rb
@@ -34,6 +34,6 @@ module NexmoDeveloper
       g.orm :active_record, primary_key_type: :uuid
     end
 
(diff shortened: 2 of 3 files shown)


mheap:
Should we also default to the current directory for backwards compatibility?

fabianrbz:
@mheap what do you mean? it's using the current directory, so it should work as before.

mheap:
https://github.com/Nexmo/nexmo-developer/blob/master/config/application.rb#L37

Should this be

```
config.docs_base_path = ENV['DOCS_BASE_PATH'] || '.'
```

So that we don't impact existing review apps / local installations?

fabianrbz:
@mheap review apps pick it up from `app.json` and with this docker will work too.
Only local installs will require to update the `.env`. Anyway, updating the code wouldn't hurt, so I'm on it :D

mheap:
New review apps pick it up from `app.json`, but existing ones don't - plus the impact on existing local installs. We have the opportunity to make this backwards compatible, and we should take it (which I see you're already doing!)

mheap:
@fabianrbz Something I was just hit by - do we expect them to specify the trailing `/` on the `DOCS_BASE_PATH`? We expect them to on `nexmo-oas-renderer` but it looks as though we don't in NDP. Can we make it handle both cases?

fabianrbz:
@mheap you are right, I'm guessing we want to include the trailing `/`. I'll do that as part of a new PR. Given that this is just a fix to allow people to use docker

PR #2437, opened 2020-02-12
Rails Executable with Documentation Path Argument

benhayehudi:
## Description

This adds a new `binstub` of `bin/initiate` to NDP where a documentation path can be specified to boot up NDP with at runtime.

Usage example:

`bundle exec initiate server --docs='../documentation/path'`
or
`bundle exec initiate server --d='../documentation/path'`

Additionally, a help flag can be raised:

`bundle exec initiate --help`


mheap:
If I run gem install `nexmo-developer` I’d expect to have a `nexmo-developer` stub automatically added to my `$PATH`

Looks like we’d need to rename `bin/initiate` to `bin/nexmo-developer` and add the following to the gemspec:

```
s.executables << 'nexmo-developer'
```

via https://guides.rubygems.org/make-your-own-gem/#adding-an-executable

PR #2438, opened 2020-02-12
Added the claims back in to the PHP example

dragonmantank:
## Description

Added the JWT claims back to the PHP example

## Deploy Notes

N/A

### DIFF
--- a/_documentation/en/conversation/guides/jwt-acl.md
+++ b/_documentation/en/conversation/guides/jwt-acl.md
@@ -126,7 +126,22 @@ $keypair = new \Nexmo\Client\Credentials\Keypair(
 );
 $client = new \Nexmo\Client($keypair);
 
-$token = $client->generateJwt();
+$claims = [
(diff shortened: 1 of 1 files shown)


_documentation/en/conversation/guides/jwt-acl.md
```diff
@@ -126,7 +126,22 @@ $keypair = new \Nexmo\Client\Credentials\Keypair(
 );
 $client = new \Nexmo\Client($keypair);
 
-$token = $client->generateJwt();
+$claims = [
+    'acl' => [
+        'paths' => [
+            '/*/users/**' => [],
```
mheap: Won't this serialize as `[]` rather than `{}`?

dragonmantank: You are correct, it does. The API doesn't seem to mind though, but I'll change it to serialize an object.

PR #2441, opened 2020-02-13
Code snippet variables

mheap:
## Description

We don't have much consistency on variable name usage and descriptions. This PR converts the free text table to a list of names that pull the descriptions from a config file.

If we agree that this is a good idea, I'm happy to update all of the existing snippets

## Deploy Notes

N/A

### DIFF
--- a/spec/filters/snippet_variables_filter_spec.rb
+++ b/spec/filters/snippet_variables_filter_spec.rb
@@ -1,0 +1,61 @@
+require 'rails_helper'
+
+RSpec.describe SnippetVariablesFilter do
+  it 'returns whitespace with whitespace input' do
+    input = ' '
+
+    expected_output = ' '
+
+    expect(described_class.call(input)).to eql(expected_output)
+  end
+
+  it 'returns unaltered non-matching input' do
+    input = 'some text'
+
+    expected_output = 'some text'
+
+    expect(described_class.call(input)).to eql(expected_output)
+  end
+
+  it 'outputs multiple lines' do
+    input = <<~HEREDOC
+      ```snippet_variables
+      - NEXMO_API_KEY
+      - TO_NUMBER
+      ```
+    HEREDOC
+
+    expected_output = <<~HEREDOC
+      Key | Description
+      -- | --
+      `NEXMO_API_KEY` | You can find this in your account overview
+      `TO_NUMBER` | The number you are sending the SMS to in E.164 format. For example `447700900000`.
+
+    HEREDOC
+
+    expect(described_class.call(input)).to eq(expected_output)
+  end
+
+  it 'throws with an empty list' do
+    input = <<~HEREDOC
+      ```snippet_variables
+      ```
+    HEREDOC
+
+    expect do
+      described_class.new.call(input)
+    end.to raise_error('No variables provided')
+  end
+
+  it 'throws with the wrong data type' do
+    input = <<~HEREDOC
+      ```snippet_variables
+      Not a list
+      ```
+    HEREDOC
+
+    expect do
+      described_class.new.call(input)
+    end.to raise_error('Must provide a list')
+  end
+end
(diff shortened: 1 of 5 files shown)


benhayehudi:
Would it be possible to make these changes in the new markdown renderer gem as that will soon become the source of the Markdown Pipeline?

mheap:
@benhayehudi I saw that the PR to merge that was closed. I'm happy to move the filter across later when we're ready to revisit that pipeline

benhayehudi:
The PR to merge it into NDP was closed, but the work on the renderer is almost done in its repo.

spec/filters/snippet_variables_filter_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'rails_helper'
+
+RSpec.describe SnippetVariablesFilter do
+  it 'returns whitespace with whitespace input' do
```
fabianrbz: you can place the strings in the method call and expectation and gert rid of the variables

lornajane:
Can we update the instructions as well please since lots of "outsiders" contribute to snippets as well? Also is it possible to override this if something different is needed in one particular code snippet context, or should I use a different variable name in that situation so that I can control the copy?

mheap:
@lornajane What kind of situation are you seeing where we'd need the same variable name but a different description?

lornajane:
We mostly workaround this by using a selection of `NUMBER`, `TO_NUMBER`, `RECIPIENT_NUMBER` and that sort of thing but for example in SMS API we use

```
TO_NUMBER | The phone number you are sending the message to.
```

But in Voice API that is instead

```
TO_NUMBER | The number you would like to call to in E.164 format. For example 447700900001.
```

I quite like that we describe what the number is used for _in this code snippet_ and while I am totally +1 for things like API keys/secrets, for the variables I might like to be descriptive within the specific context where I'm using the field - if that makes any sense!

mheap:
It makes total sense. How about being explicit in what's required, but allowing us to render in a certain way? e.g.

```
SMS_TO_NUMBER:
  description: The number you are sending the SMS to in E.164 format. For example `447700900000`.
  render_as: TO_NUMBER
VOICE_TO_NUMBER:
  description: The number you would like to call to in E.164 format. For example 447700900001.
  render_as: TO_NUMBER
```

This would allow us to have a central source of variable names to improve consistency whilst still allowing for additional information based on the context

lornajane:
That would work - I think it's either some sort of renaming dodge like this, or per-product configuration which would encourage us to stay consistent within each section and not lose our way with what could become quite a long list eventually. I'm for per-product but it's not me that is going to build it so absolutely your call.

mheap:
I could go either way on it.

@tbedford As the main consumer of this, would you prefer one big list with `render_as` or multiple files, and specifying the one to use in the snippet e.g.


````
```snippet_variables
type: voice
variables:
  - TO_NUMBER
  - FROM_NUMBER
````

dragonmantank:
I would say, as someone who has to write the code snippets (and would like to automate running them...) having per-product names would be nice, but I totally understand if that's not maintainable.

tbedford:
@mheap I would prefer one file with everything in it, so the `render_as` option.

mheap:
@dragonmantank Would you be able to run your tooling with a single file and `render_as`?

dragonmantank:
@mheap I'd rather not have to pull down the code standards repo just to run tests, just have a clear naming convention for this (though, honestly, the current runner doesn't care since you can embed environment variables into the test, it's more of an issue if you want to have local values for testing through something like a `.env` file).

mheap:
@dragonmantank If it's contained in a single file we can copy between NDP and the snippet tester. Not an ideal solution, but the rate of change for the variable change is intended to be small so it should work for now.

I'll work on updating this PR to use the new Markdown pipeline and variable names file

PR #2468, opened 2020-02-26
NDP Gemification

benhayehudi:
Making NDP into a bootable gem


lib/nexmo_developer/app/controllers/api_errors_controller.rb
```diff
@@ -63,7 +63,11 @@ def scoped_error(definition, id)
   end
 
   def error_config
-    @error_config ||= YAML.load_file("#{Rails.root}/config/api-errors.yml")
+    if Pathname.exist?("#{Rails.configuration.docs_base_path}/config/api-errors.yml")
```
fabianrbz: @benhayehudi can we do something similar to what i suggested in `nexmo-markdown-renderer`?

lib/nexmo_developer/app/presenters/topnav.rb
```diff
@@ -65,12 +65,12 @@ def configure_item_navigation(name)
 
   def navbar_items_default
     @items ||= [
-      ['Documentation', '/documentation'],
-      ['Use Cases', '/use-cases'],
-      ['API', '/api'],
-      ['Tools', '/tools'],
-      ['Community', '/community'],
-      ['Extend', '/extend']
+      {:name=>"Documentation", :url=>"/documentation", :navigation=>"documentation"},
```
fabianrbz, line 68: @benhayehudi can you make these into objects too? also, navigation could be calculated from the url (they should be the match), something like `'/use-case'.delete('/')`

benhayehudi: I understand the second half of your comment. Can you explicate the first half?

benhayehudi: It seems that the navigation parameter is more complicated than just a deleting of the backslash, hence why I maintained it as a separate parameter in the `navbar_items_default` method. However, when reading from the YAML config file, it calls the `configure_item_navigation` method that invokes `#parameterize.underscore` on the name, which is what `Navigation` is built to expect, it looks like.

lib/nexmo_developer/app/presenters/footer.rb
```diff
@@ -0,0 +1,53 @@
...
+  attr_reader :items
+  def initialize(items: nil)
+    @items = items
+
+    after_initialize!
+  end
+
+  def footer_from_config(config)
+    config = YAML.safe_load(open_config(config))
+    @items = [
```
fabianrbz, line 11: @benhayehudi you are assigning `@items` both here and in https://github.com/Nexmo/nexmo-developer/pull/2468/commits/d7ed1ce6b0eb67d40c7548442b1770f1ee809e0e#diff-3d74b2b92492afe32f017518cd5acb93R51

lib/nexmo_developer/app/presenters/footer.rb
```diff
@@ -0,0 +1,53 @@
...
+        status_link_text: config['footer']['links']['status']['text'],
+        social_links: config['footer']['links']['social'],
+        navigation_links: config['footer']['links']['navigation'],
+        support_links: config['footer']['links']['support']
+      }
+    ]
+  end
+
+  def self.navigation_category_titleize(title)
+    if title.include?('-')
```
fabianrbz, line 26: @benhayehudi what's the purpose of this?

lib/nexmo_developer/app/presenters/footer.rb
fabianrbz: also, why is this an array when it only has one element?

lib/nexmo_developer/app/views/layouts/partials/_footer.html.erb
```diff
@@ -1,70 +1,50 @@
...
+          <h5 class="Vlt-white"><%= item[:name] %></h5>
+          <small><%= item[:subtitle] %></small>
+        </div>
       </div>
-    </div>
 
-      <a href="https://www.nexmostatus.com/" target="_blank" class="Vlt-badge-combined">
-        <span class="Vlt-badge Vlt-white Nxd-api-badge">API Status</span><span class="Vlt-badge Vlt-badge--purple Nxd-api-status">Loading</span>
-      </a>
+        <a href=<%= item[:status_link_path] %> target="_blank" class="Vlt-badge-combined">
```
fabianrbz: can you use objects instead of hashes?

lib/nexmo_developer/app/presenters/footer.rb
benhayehudi: This method upcases the word API but titleizes everything else

lib/nexmo_developer/app/presenters/footer.rb
benhayehudi: You're right, I'll address that.

lib/nexmo_developer/app/views/layouts/partials/_footer.html.erb
benhayehudi: why?

PR #2491, opened 2020-03-03
Update fallback_answer_url behavior

benhayehudi:
## Description

The `fallback_answer_url` has changed in behavior:

```
Name | Type | Description
-- | -- | --
reason | string | Error message, one of“Connection closed.”“Connection reset.”“Socket timeout.”“Invalid NCCO.”“Invalid response code.”
original_request | struct | Original request info.
original_request.url | string | Originally requested URL.
original_request.type | enum | One of  { answer, event }
```

### DIFF
--- a/_documentation/en/application/overview.md
+++ b/_documentation/en/application/overview.md
@@ -66,7 +66,7 @@ The webhook URLs you provide when creating an application depend on the applicat
 
 Capability | API used | Webhooks available
 --- | --- | ---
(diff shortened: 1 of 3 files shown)


mheap:
This functionality isn’t live yet I believe

benhayehudi:
> This functionality isn’t live yet I believe

It will be live on Friday, so we shouldn't merge yet, but the request is to have it live when or as close to when the feature goes live.

vshisterov:
We have it delivered just now.

PR #2471, opened 2020-02-26
update code snippets after node audit

AlexLakatos:
https://github.com/Nexmo/nexmo-node-code-snippets/pull/75 need to be merged before this one.

PR #2498, opened 2020-03-04

NexmoDev:


lornajane:
These changes will come in with #2471, this PR isn't needed (but thanks anyway, robot!)

PR #2499, opened 2020-03-04
Update code snippets after Node audit

AlexLakatos:

PR #2505, opened 2020-03-06

NexmoDev:


AlexLakatos:
Pulled this into #2499 instead

PR #2511, opened 2020-03-10
API Specification Update

NexmoDev:


tbedford:
@niati Just to let you know this is an example where an existing automated PR updates due to a change in the submodule repo.

PR #2520, opened 2020-03-12
Add "Nexmo is now Vonage" image

lornajane:
## Description

We're going to need to add an image to all our GitHub READMEs to aid our brand transition, this is a good and reliable place to host it!


mheap:
@fabianrbz Does this image already exist in the rebrand branch? If so, should we point to that?

fabianrbz:
@mheap @lornajane we don't, we have something similar for the header and footer, but that was built with html and css

lornajane:
OK, Isaac is happy so if we can merge sometime, that would be great!

mheap:
:shipit:

PR #2571, opened 2020-03-23
Updated overview of the Reports API

obodriagovatvonagedotcom:
Added information about Reports SYNC to the overview of the Reports API
Added comparison table between Reports SYNC and ASYNC

### DIFF
--- a/_documentation/en/reports/overview.md
+++ b/_documentation/en/reports/overview.md
@@ -5,7 +5,7 @@ meta_title: The Reports API
 
 
 # Reports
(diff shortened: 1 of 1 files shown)


mheap:
The API reference for this is still pending, so adding `don't merge` (but please do review the content)

tbedford:
@obodriagovatvonagedotcom I am not happy about this PR as it looks like it was made from a very old fork of nexmo-developer. In particular there should be a `_documentation/en` folder. The absence of one means this fork was made from a very old copy of nexmo-developer. I would be nervous about merging this, especially as the review app is failing. Can you redo the PR from a recent fork please?

PR #2589, opened 2020-03-26
Updated community page to change all Nexmo references to Vonage.

ruskibenya:
## Description

Updated community page header, contact email, twitter feed, and slack title.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -24,8 +24,8 @@ ActiveRecord::Schema.define(version: 2019_10_22_100247) do
     t.uuid "resource_id"
     t.string "author_type"
     t.uuid "author_id"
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -24,8 +24,8 @@
     t.uuid "resource_id"
```
fabianrbz: Can you drop the changes to this file? I'll take a closer look to them later.

benhayehudi: @ruskibenya, all you need to do here is revert the change to the schema.rb and set it back to what it was before and push that up to this PR. I think everything else looks 👍

fabianrbz:
@ruskibenya do you need a hand with this?

PR #2590, opened 2020-03-26
Adding Glitch to the App to Phone tutorial

conshus:
## Description
New branch since Octokit bump broke the deploy of the other branch
Making changes to the "Making an in-app voice call" tutorial for the Make a call in < 15 mins goal
Adding a Glitch Webhook server option
Create the plan for the JS internal testing https://nexmoinc.atlassian.net/browse/DEVX-2475

## Review

* [Pages to review](https://nexmo-developer-pr-2590.herokuapp.com/client-sdk/tutorials/app-to-phone/introduction)


config/tutorials/en/app-to-phone.yml
```diff
@@ -3,35 +3,33 @@ description: You make a voice call from a web app to a phone.
...
     phone call from your web app to a <abbr title="Public Switched Telephone Network">PSTN</abbr> phone.
 
     To achieve this, you will need:
 
     * A web application that will make the voice call using the Client SDK
-    * A webhook server that can process an inbound webhook and return an <abbr title="Nexmo Call Control Object">NCCO</abbr> to forward the inbound voice call to the destination phone
+    * An <abbr title="Nexmo Call Control Object">NCCO</abbr> to forward the inbound voice call to the destination phone
 
-    When the call is triggered from your application, Nexmo will make a request to your webhook server, passing the number provided by the application in the `GET` parameters. You will use this to generate the correct NCCO and forward the call to the destination phone number.
+    When the call is triggered from your application, Nexmo will get the NCCO from Glitch and forward the call to the destination phone number.
```
tbedford, line 19: > When the call is triggered from your application, Nexmo will get the NCCO from Glitch. 

I think at the very least we should make "Glitch" a link to the Glitch site. Perhaps a few words about why we are using Glitch to host the NCCO would be useful to clarify why we are taking this approach.

_tutorials/en/client-sdk/app-to-phone/modify-ncco.md
```diff
@@ -0,0 +1,18 @@
...
+description: In this step, you modify your NCCO using Glitch.
+---
+
+# Modify NCCO
+
+A Nexmo Call Control Object (NCCO) is a JSON array that you use to control the flow of a Voice API call. More information on NCCO can be found [here](https://developer.nexmo.com/voice/voice-api/ncco-reference).
+
+The NCCO must be accessible by the internet. To accomplish that, we will be using [Glitch](https://glitch.com).
+
+To get your NCCO:
```
tbedford, line 12: "To get your NCCO:" -> To create your NCCO.

_tutorials/en/client-sdk/app-to-phone/code.md
```diff
@@ -12,31 +12,62 @@ Create an HTML file called `index.html` in your project directory. Add the follo
...
-      .then(app => {
-        callPhoneForm.addEventListener("submit", event => {
-          event.preventDefault();
-          let number = callPhoneForm.children.phonenumber.value;
-          app.callServer(number);
-        });
-      })
-      .catch(console.error);
-  </script>
+    <label for="phone-number">Your Phone Number:</label>
```
tbedford, line 28: > Your Phone Number:

I thought we decided we don't need this? The destination number is set in the NCCO so if you want to simplify things the user does not need to put in another number. I tried putting in another number here and it is of course not used as the NCCO forwards the call.

_tutorials/en/client-sdk/app-to-phone/code.md
```diff
@@ -12,31 +12,62 @@ Create an HTML file called `index.html` in your project directory. Add the follo
...
+    <input type="text" name="phone-number" value="" placeholder="i.e. 14155550100" id="phone-number" size="30">
```
tbedford, line 29: Do we really need this? I think we should only need to Call and Hangup buttons.

_tutorials/en/client-sdk/app-to-phone/code.md
```diff
@@ -12,31 +12,62 @@ Create an HTML file called `index.html` in your project directory. Add the follo
...
+        const hangupButton = document.getElementById("hangup");
+        const statusElement = document.getElementById("status");
+        new NexmoClient({ debug: true })
+            .login(USER_JWT)
+            .then(app => {
+                callButton.addEventListener("click", event => {
+                    event.preventDefault();
+                    let number = phoneNumberInput.value;
+                    if (number !== ""){
+                        app.callServer(number);
```
tbedford, line 47: > `app.callServer(number);`

Is there a version of this method that does not use a number? We shouldn't need to specify a number as the destination phone number is set in the NCCO.

config/tutorials/en/app-to-phone.yml
conshus: So, should I add a link here and in the Modify NCCO step?
I mention why we are using Glitch in the Modify NCCO step. I figured since we go more in-depth on the NCCO in that step, that I would mention it there.

_tutorials/en/client-sdk/app-to-phone/code.md
conshus: With regard to the other issues with the phone number, I'll answer it all here. 
I don't see in the documentation where it states that "number" in app.callServer(number) is optional or any other examples that don't have "number". 

I think it would be more confusing to omit it and then they see it in other places being used. So I decided to keep it how the original tutorial had it including the phone number input element.

_tutorials/en/client-sdk/app-to-phone/code.md
conshus: Please see: https://github.com/Nexmo/nexmo-developer/pull/2590#discussion_r406253267

_tutorials/en/client-sdk/app-to-phone/code.md
conshus: Please see: https://github.com/Nexmo/nexmo-developer/pull/2590#discussion_r406253267

_tutorials/en/client-sdk/app-to-phone/modify-ncco.md
conshus: Should I still have the ":" or replace it with a "." (or were you ending your reply)?

config/tutorials/en/app-to-phone.yml
tbedford: Making it a link then should be enough as you discuss in more detail later. The first time the reader sees "Glitch" they might wonder what is is and what it's for if they've not come across it before.

_tutorials/en/client-sdk/app-to-phone/modify-ncco.md
tbedford: "To get your NCCO:" -> "To create your NCCO:"

_tutorials/en/client-sdk/app-to-phone/code.md
tbedford: Oh dear. That's going to be very confusing for a new developer. It was for me and I'm reasonably familiar with this stuff. If I have the number specified in the NCCO why do I need to go to the trouble of re-entering it in an input box and then passing it via a function call? If that was what was needed then I wouldn't need the NCCO at all. Problem then of course is if we are not using the input box then how do we then populate the number for the method call? I would raise the question of `app.callServer(number)` on Slack with the developers.

conshus: Hmmmm... my worry is that developers not knowing that if they need to programmatically change the call number, they won't know how to pass it because we took it out.

How about if we have a note or something that says in a more complex you would programmatically set the number kind of like we do for using the CLI to generate a JWT?

tbedford: Yes, it is confusing, but apparently there is no version of the method call where number is not required. So, we need to stick with what you have. We need to make it clear that hard coding the number in the NCCO is only for test purposes and unlikely in a production use case. I think the adding of a note as you say is a good idea. Thanks!

fabianrbz:
@tbedford @conshus yeah, I did move some things around, I can help with the rebase if you want.

PR #2592, opened 2020-03-27
WIP: Move tutorials to sublevel in TOC

tbedford:
## Description

Moves tutorials to next level down. In-app messaging tutorials and in-app voice tutorials go down a level.

## Review

You need to click In-app Messaging and In-app Voice in the TOC to see where the Tutorials are now:

* [Review app](https://nexmo-developer-pr-2592.herokuapp.com/client-sdk/overview)


tbedford:
@fabianrbz - I'm not sure about the redirects. Are they required?

tbedford:
@pardel this demonstrates the idea of tutorials being organized into in-app messaging and in-app voice subtopics. We could still have tutorials in the parent level of the TOC for tutorials that use both. Would be interested to know what you think.

tbedford:
@niati - would be interested in your thoughts on this. Tutorials in relevant subtopics, or all at the top level?

niati:
@tbedford I like having relevant tutorials grouped in the TOC with corresponding topics as well as at the top level. It gives the tutorials a lot more exposure and provides an easy launch pad from which to dive into them directly from the relevant content.

tbedford:
OK let's leave them where they are then. Thanks!

PR #2603, opened 2020-03-31
Update VAPI Websocket Docs

benhayehudi:
## Description

This PR addresses DEVX-2552 and DEVX-2553.

### DIFF
--- a/_documentation/en/voice/voice-api/webhook-reference.md
+++ b/_documentation/en/voice/voice-api/webhook-reference.md
@@ -74,6 +74,7 @@ The format of the data included depends on which event has occurred:
 * [`busy`](#busy)
 * [`cancelled`](#cancelled)
 * [`unanswered`](#unanswered)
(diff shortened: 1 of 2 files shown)


_documentation/en/voice/voice-api/webhook-reference.md
```diff
@@ -182,6 +183,21 @@ Field | Example | Description
 
 [Back to event webhooks list](#event-webhook)
 
+### Disconnected
+
+If the WebSocket connection is terminated from the application side by any reason, the disconnected event callback will be sent.
```
tbedford: We don't use camel case for websocket elsewhere in the page. However if `WebSocket` refers to a class it should be enclosed in backticks.

benhayehudi: We don't in this page, but we do elsewhere, see for example: https://developer.nexmo.com/voice/voice-api/guides/websockets#further-reading

PR #2615, opened 2020-04-01
Disable map in community pages if there are no events with locations

fabianrbz:
## Description

## Deploy Notes

### DIFF
--- a/app/admin/event.rb
+++ b/app/admin/event.rb
@@ -2,7 +2,7 @@ ActiveAdmin.register Event do
   # See permitted parameters documentation:
   # https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#setting-up-strong-parameters
   #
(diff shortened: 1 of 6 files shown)


mheap:
@fabianrbz Is this the correct solution? What about when we want to add it back?

How about hiding the map if we have zero locations with an event scheduled?

fabianrbz:
@mheap whatever we do, we'll need to revert the commit in order to add it back.
The issue with the locations is that every event has a country so we would need to check for the presence of a city. however add an `allow_blank` to the country dropdown,

mheap:
If we hide the map when we have no locations we can keep it live forever and it'll appear/hide as required

Have we considered how we handle remote events?

fabianrbz:
@mheap we could add a boolean column and a checkbox to the admin UI

mheap:
I'll take your recommendation how to allow it, but I think it's going to be important in the near future

myrsiningos:
All looks good to the Community team.

PR #2631, opened 2020-04-06
Fixed iOS Github repository URL

igorwojda:


config/tutorials/en/ios-in-app-messaging-chat.yml
```diff
@@ -8,13 +8,13 @@ introduction:
   description: This task shows you how to create a iOS chat application using the iOS Client SDK.
   content: |
     # Introduction
-    In this task, you will learn how to create and configure a Client SDK Application and then code a web app that enables two users to send messages to each other.
+    In this task, you will learn how to create and configure a Client SDK Application and then code the app that enables two users to send messages to each other.
 
-    First, you will create a Conversation and two Users. Then, you will authenticate these Users and add them to the Conversation as Members.
+    First, you will create a `Conversation` and two `Users`. Then, you will authenticate these `Users` and add them to the `Conversation` as `Members`.
 
     Finally, you will implement chat functionality in your application, including the ability to view and send messages and be notified when either party is currently typing.
```
igorwojda, line 15: @pardel is this actually implemented in iOS app?

pardel: @igorwojda not entirely sure - will investigate.

igorwojda: I think regardless of this comment we should merge this PR because it fixes critical part of tutorial.

This potencial "functionality issue" can be addressed separately

config/tutorials/en/ios-in-app-messaging-chat.yml
```diff
@@ -8,13 +8,13 @@ introduction:
   description: This task shows you how to create a iOS chat application using the iOS Client SDK.
   content: |
     # Introduction
-    In this task, you will learn how to create and configure a Client SDK Application and then code a web app that enables two users to send messages to each other.
+    In this task, you will learn how to create and configure a Client SDK Application and then code the app that enables two users to send messages to each other.
 
-    First, you will create a Conversation and two Users. Then, you will authenticate these Users and add them to the Conversation as Members.
+    First, you will create a `Conversation` and two `Users`. Then, you will authenticate these `Users` and add them to the `Conversation` as `Members`.
```
tbedford, line 13: I think we should probably only use backticks around code or a class. Generally, where we are talking about about a Conversation as a container we have used capitalization to indicate it is a Nexmo technical term, rather than a general conversation. For example, "Two users can have a conversation via a Conversation blah..." would be OK. Also, "Instantiate a `Conversation` class..." would be OK.

igorwojda: Fixed

PR #2644, opened 2020-04-07
Update SDK docs

dimitrisniras:


fabianrbz:
@dimitrisniras looks like there is an issue with the symlink, the spec we have for the ios docs is failing

dimitrisniras:
> @dimitrisniras looks like there is an issue with the symlink, the spec we have for the ios docs is failing

Got it, I will close this one and a new PR will be created with the right latest symlink

PR #2652, opened 2020-04-09
[DEVX-2685] Group bugnsag exceptions into `Missing Document` and `404…

fabianrbz:
… - Not Found`.

## Description

This allows to easily identify exceptions releated to missing files.

### DIFF
--- a/app/services/not_found_notifier.rb
+++ b/app/services/not_found_notifier.rb
@@ -1,11 +1,11 @@
 IGNORED_FORMATS = %w[php action swf].freeze
 
 class NotFoundNotifier
-  def self.notify(request)
+  def self.notify(request, exception)
     return if ignored_format?(request.params['format'])
     return if crawler?(request.user_agent)
 
-    Bugsnag.notify('404 - Not Found') do |notification|
+    Bugsnag.notify(notification_name(exception)) do |notification|
       notification.add_tab(:request, {
         params: request.params,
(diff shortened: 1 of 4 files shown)


app/services/not_found_notifier.rb
```diff
@@ -23,4 +23,12 @@ def self.ignored_format?(format)
   def self.crawler?(user_agent)
     Woothee.parse(user_agent)[:category] == :crawler
   end
+
+  def self.notification_name(exception)
+    if [Nexmo::Markdown::DocFinder::MissingDoc, Errno::ENOENT].any? { |e| exception.instance_of? e }
```
mheap, line 28: Would you expect `Missing Document` on `Errno::ENOENT`? This would be misleading for me as it'd trigger on missing assets, YAML files etc

fabianrbz: hmm, do missing assets trigger `Errno::ENOENT`?  I think they are caught [here](https://github.com/Nexmo/nexmo-developer/blob/master/config/routes.rb#L109) and that one doesn't raise an exception. 

I included `Errno::ENOENT` because something tried to read a file and it isn't there, but i can remove it and leave `Nexmo::Markdown::DocFinder::MissingDoc`.

fabianrbz: @mheap removed!

PR #2663, opened 2020-04-15
[DEVX-2718] Bump volta.

fabianrbz:
## Description

Fixes issues with missing programming languages support for code snippets.

## Note

Objective-c code snippets are not being picked up by prism because we need to add some custom code to map the key that prism expects.


igorwojda:
Regarding language support:
- `Kotlin` highlighted properly
- `Objective-C` - still no support

(we should merge but remember about adding `Objective-C` support at some point)

PR #2693, opened 2020-04-17
Add rake task to detect duplicate navigation weights

mheap:
## Description

Adds a `rake` task that raises if any documents have a duplicate navigation weight

Usage: `bundle exec rake ci:duplicate_navigation_weight`

This should be added to the GH Actions CI task once all errors are resolved

## Deploy Notes

N/A

### DIFF
--- a/lib/tasks/ci.rake
+++ b/lib/tasks/ci.rake
@@ -144,4 +144,50 @@ namespace :ci do
       raise "Blocked words found:\n\n#{errors.join("\n")}"
     end
   end
+
+  task 'duplicate_navigation_weight': :environment do
+    markdown_files =
+      [
+        "#{Rails.configuration.docs_base_path}/_documentation/en/**/*.md",
+        "#{Rails.configuration.docs_base_path}/_api/**/*.md",
+        "#{Rails.configuration.docs_base_path}/_tutorials/**/*.md",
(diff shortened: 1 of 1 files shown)


lib/tasks/ci.rake
```diff
@@ -144,4 +144,50 @@ namespace :ci do
...
+      end
+    end
+
+    if errors.length.positive?
+      error = "Duplicate weights found:\n\n"
+      errors.each do |err|
+        error += "#{err['path']} (Weight: #{err['weight']})\n#{err['items'].join("\n")}"
+        error += "\n------------------------------\n"
+      end
+      raise error
```
fabianrbz, line 190: @mheap instead of raising, you can use `abort` it prints to STDERR and also exits with an error status code. See https://ruby-doc.org/core-2.7.1/Kernel.html

benhayehudi: Should we switch to `#abort` in other places as well. Seems like a better method to use. 🎓

tbedford:
@mheap @fabianrbz - wasn't sure looking at the comments whether this should be merged or not. It has been approved but it looks like there may be an issue still to resolve?

mheap:
Updated to use `abort` 👍

PR #2721, opened 2020-04-22
update android release-notes for v2.6.1

vonageCI:


VMadalin:
> * I think we can use the language specifier (`java` or `kotlin`) on all code blocks (triple backticks) in these release notes. Our renderer support syntax colouring for these languages now.

Of course, I absolutely agree @igorwojda @tbedford with that but this should be on the next release because that implies to refactor all the entire file and this release was for a hotfix for customer `Latch`

PR #2756, opened 2020-04-29
Added Vonage Developer Day landing page at /day

phazonoverload:
## Description

Added a page at /day for the Vonage Developer Day event, using the config-based landing pages

### DIFF
--- a/app/views/layouts/partials/_notices.html.erb
+++ b/app/views/layouts/partials/_notices.html.erb
@@ -88,7 +88,7 @@
 
 <div class="banner">
   <div class="Nxd-logo">
-    <a target="_blank" href="https://www.eventbrite.com/e/vonage-developer-day-tickets-102886113134" id="logo" class="Nxd-logo__image">
+    <a href="/day" id="logo" class="Nxd-logo__image">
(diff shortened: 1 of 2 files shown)


tbedford:
Looks really nice!

phazonoverload:
Thanks @tbedford!

phazonoverload:
Yeah, this was something I was uncertain about from looking at the directory. I assumed that first-party events/initiatives would live in landing_pages and third-party events in landing_pages/e. 

Happy to move it - just let me know what you’d like. 

If it’s open for discussion I’d suggest my approach above.

benhayehudi:
> Yeah, this was something I was uncertain about from looking at the directory. I assumed that first-party events/initiatives would live in landing_pages and third-party events in landing_pages/e.
> 
> Happy to move it - just let me know what you’d like.
> 
> If it’s open for discussion I’d suggest my approach above.

I like that approach. When we first initiated this, the thinking was _all_ events in `/e`, but I think making explicit that 3rd party events in `/e`, which then necessitates a bit.ly link to make a simple URL, and our events top level, makes sense to me. We want _our_ events to be developer.nexmo.com/our-event, in any case. 

I'll approve and add a note in the landing page guide about this as well. Thanks!

PR #2761, opened 2020-04-30
Add Reports API Pricing information

tbedford:
## Description

https://nexmoinc.atlassian.net/browse/DEVX-2814

* Add Pricing information.
* Update free pricing period to 1st June as requested in the Slack discussion.

## Review

* [Overview with pricing info](https://nexmo-developer-pr-2761.herokuapp.com/reports/overview)

### DIFF
--- a/_documentation/en/reports/overview.md
+++ b/_documentation/en/reports/overview.md
@@ -13,7 +13,7 @@ You can use the Reports API in a wide variety of use cases, including:
 * Invoice reconciliation - Compare your usage data with the invoice you have received.
 * Monitoring and analytics - Add CDR data to your business intelligence or analytics system to correlate it with other events.
 
-> The Reports API Beta is available free of charge for all customers until the 1st of May 2020. After that date, charges will be applied if the number of requests exceeds the Free Usage Tier. Please use [this form](https://info.nexmo.com/ReportsAPI.html) to request the API pricing.
+> The Reports API Beta is available free of charge for all customers until the 1st of June 2020. After that date, charges will be applied if the number of requests exceeds the Free Usage Tier. Please use [this form](https://info.nexmo.com/ReportsAPI.html) to request the API pricing.
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


tbedford:
@mheap - Is this going to be enough, or were you referring to the API Reference spec in your ticket?

mheap:
I think we should add some pricing info from the PDF on the page too

tbedford:
Could that lead to some issues if the prices change without our notification? Very wary of putting price info in developer docs.

mheap:
In general yes, but there's nowhere else with this pricing available at the moment.

With the migration to Reports, we'll get a lot of support requests if it's not easily accessible

tbedford:
How do people usually get pricing info? I think in some cases we ask them to contact sales for pricing info don't we? Perhaps there needs to be a support page or product page or something. They can then put in all the disclaimers and legal jargon about possible price changes etc. We have to be careful we don't put a price in the docs and then someone gets charged a difference price and then we end up with legal issues.

mheap:
In this case they've got to fill out the linked form. I asked Olek and he said it was fine to put the pricing on the docs page.

Long term, it shouldn't live with us. In the short term, it's a worthwhile trade off

tbedford:
@obodriagovatvonagedotcom - could you check the pricing info please? Thanks

obodriagovatvonagedotcom:
Michael is right, it is only its temporary "dwelling" :)

Currently, the pricing of Reports API is mentioned only here: https://www.vonage.com/communications-apis/enterprise/pricing/
You can see that is says "Contact Sales".

Next step will be to add Reports API pricing here: https://www.vonage.com/communications-apis/pricing/

tbedford:
@obodriagovatvonagedotcom - are you happy with the information contained in this PR? If so I will get it reviewed and merged. Thanks.

obodriagovatvonagedotcom:
Without knowing how API works this pricing might look extremely complex, and the way we present it is backwards: at first we describe pricing and then we describe how it works. I would suggest to add pricing afterwards.

Moreover, we are considering changing billing so that it is possible to create limited number of Reports for free in CD, but make API totally paid.
Thus I would add that prices are subject to change and that they are preliminary.

tbedford:
If you do change the pricing, please let me know. **It is critical that customers are not misinformed on pricing in the docs.** 

I will have a look at positioning of pricing info, but I think the more "up front" it is the better. Yes, it might be complicated, but they will be able to refer back to it. The prominent position will mean they know where to find it. Having said that I will have a look today to see if there's a better option. I also did say in the title "subject to change" but looks like I need to add another note on that to make it even more obvious!

PR #2772, opened 2020-05-04
Update SDK docs

dimitrisniras:


dimitrisniras:
> For future reference the copyright notice should be as I previously specified, that is without the ', Inc'. Also, the LICENSE link does not work. Please fix for next time. Thanks.

Will fix the `Inc` in next release, but as for the LICENCE link, you cannot reach it since JS SDK is still a private repo and not a public one.

PR #2785, opened 2020-05-06
[DEVX-2848] Fix <title> and reword it.

fabianrbz:
## Description

Fix <title> tag to use the `@document_title`
In case of markdown files, it tries to use `meta_title` first, and fallback to `title` if it is not present.

### DIFF
--- a/app/views/layouts/partials/_head.html.erb
+++ b/app/views/layouts/partials/_head.html.erb
@@ -2,10 +2,10 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
-  <title><%= title %></title>
+  <title><%= page_title %></title>
 
-  <% if @frontmatter && @frontmatter['description'] %>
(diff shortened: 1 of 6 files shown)


app/views/layouts/partials/_head.html.erb
```diff
@@ -2,10 +2,10 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
-  <title><%= title %></title>
+  <title><%= page_title %></title>
 
-  <% if @frontmatter && @frontmatter['description'] %>
-    <meta name="description" content="<%= @frontmatter['description'] %>">
+  <% if @frontmatter && @frontmatter['meta_description'] %>
+    <meta name="description" content="<%= @frontmatter['meta_description'] %>">
```
mheap, line 8: Should this default to `description` if `meta_description` doesn't exist? Is something better than nothing?

fabianrbz: @mheap done

PR #2791, opened 2020-05-08
make clear that regional domains are required for SIP

benhayehudi:
## Description

DEVX-2852 Regional domain usage must be made abundantly clear that they are required for programmable SIP and not an option.

### DIFF
--- a/_documentation/en/voice/sip/concepts/programmable-sip.md
+++ b/_documentation/en/voice/sip/concepts/programmable-sip.md
@@ -34,7 +34,7 @@ IP addresses of your devices and endpoints can be configured to reach your Nexmo
 
 ## Domain Based Routing
 
-Calls made to a Programmable SIP domain must be handled at a regional level. You must use a R-URI with a regional domain. Please be aware that a R-URI without a regional component in the domain will fail the call.
+Calls made to a Programmable SIP domain must be handled at a regional level. You must use a Request URI with a regional domain. Please be aware that a Request URI without a regional component in the domain will fail the call.
 
 The following code will indicate to Nexmo that you want this SIP call to be handled in the EU:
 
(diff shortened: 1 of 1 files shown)


_documentation/en/voice/sip/concepts/programmable-sip.md
```diff
@@ -34,7 +34,7 @@ IP addresses of your devices and endpoints can be configured to reach your Nexmo
 
 ## Domain Based Routing
 
-If you require your call to be handled by a specific Nexmo region, then you can specify that by using R-URI with such information in the domain part.
+Calls made to a Programmable SIP domain must be handled at a regional level. You must use a R-URI with a regional domain. Please be aware that a R-URI without a regional component in the domain will fail the call.
```
tbedford, line 37: What is R-URI and should we either spell it out, or link to a definition of it?

benhayehudi: I don't know myself actually. Just getting acquainted with SIP. It seems like R-URI was in the existing documentation as well. The context of the sentence leads me to believe it is short for regional URI. Is that correct, @sammachin?

sammachin: Its nothing to do with regional thats just a Vonage thing,

the R URI is the Request URI its the first line of the SIP request and defines the endpoint being called see https://transnexus.com/whitepapers/sip-invite-header-fields/

tbedford: Thanks @sammachin 

@benhayehudi I think if we just write 'Request URI', rather than R-URI that should be enough. We could also link to the document Sam referenced.

benhayehudi: Thank you, @sammachin. I appreciate you always being available to lend your expertise!

@tbedford, I pushed up the change renaming it to "Request URI" to make it clearer. I chose not to link to the article referenced above, as it's another company and I think we need more process/conversation before linking to external companies and their docs.

PR #2799, opened 2020-05-12
Devx 2299 tutorials

fabianrbz:
## Description

Change the way we write tutorials.
Now, every tutorial must have a folder under `config/tutorials/en` with a file called `base.yml` that acts as entry point, with the option to define programming language specific files.


config/landing_pages/hansel.yml
```diff
@@ -76,12 +76,12 @@ page:
       column:
       - type: tutorial
         tutorial:
-          name: two-factor-authentication-dotnet-verify-api
+          name: two-factor-authentication-dotnet-verify-api/base
```
mheap, line 79: Could we strip the `/base` off this in the URL?

fabianrbz: that's not an url, it's the name of the tutorial.
It's possible, the problem is that sometimes you might want to use a language specific tutorial, we could check if the name has a programming language, and if not, use the default one.

fabianrbz: this is applicable only to landing pages though.

mheap:
@fabianrbz Is this ready for review?

fabianrbz:
@mheap not yet

bencgreenberg:
@fabianrbz Do we still need this PR open?

PR #2801, opened 2020-05-12
Changed standard page title and team page custom title.

pmmcls:
## Description

Alter page title code to use Vonage brand and put the brand after the page title. Changed teams page title.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -11,11 +11,11 @@ module ApplicationHelper
 
   def title
     if @product && @document_title
-      "Nexmo Developer | #{@product.titleize} > #{@document_title}"
+      "#{@product.titleize} > #{@document_title} | Vonage API Developer"
     elsif @document_title
-      "Nexmo Developer | #{@document_title}"
+      "#{@document_title} | Vonage API Developer"
     else
-      'Nexmo Developer'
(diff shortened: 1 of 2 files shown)


mheap:
@fabianrbz You were making changes around `title` recently. Any overlap here?

app/helpers/application_helper.rb
```diff
@@ -11,11 +11,11 @@ def theme
 
   def title
     if @product && @document_title
-      "Nexmo Developer | #{@product.titleize} > #{@document_title}"
+      "#{@product.titleize} > #{@document_title} | Vonage API Developer"
```
fabianrbz, line 14: @pmmcls @mheap yes, this branch might be outdated.
This method was removed and the logic now lives here
https://github.com/Nexmo/nexmo-developer/blob/master/app/presenters/page_title.rb#L1

pmmcls: Interesting - I only branched a couple of days ago. ok please reject the change then

pmmcls: is that on master? I've just done a pull and page_title presenter isn't there

pmmcls: Ok not to worry - I'm obviously not doing something right at this end

PR #2806, opened 2020-05-13
Media template character limit update

slorello89:
## Description

Docs for Messages needs to be updated to reflect the new character restrictions on templates - these restrictions have been lifted from 160 chars total - to 60 chars in header / footer and 1024 chars in the body: https://developers.facebook.com/docs/whatsapp/business-management-api/message-templates/

## Deploy Notes

N/A


slorello89:
Apparently our WhatsApp cluster is not updated yet to account for the lifted restrictions - thus I'm tearing down this PR until it is

PR #2828, opened 2020-05-19
Updates for fallback policy deprecation 

slorello89:
## Description

Updating note to state that fallback policy was deprecated


slorello89:
FYI this goes with https://github.com/Nexmo/api-specification/pull/307

PR #2846, opened 2020-05-25
API Specification Update

NexmoDev:


lornajane:
No spec changes, so we don't need to pull this in.

Issue #2849, opened 2020-05-26
Endpoint Array in connect NCCO not clearly marked


slorello89:
## Description

The endpoint field in the Connect NCCO Says "Connect to a single endpoint. Available endpoint types" - which implies that it's looking for one object. In reality the expected format is an array of objects. The description should be something a bit clearer - even though it does in fact only accept one object in the array

## Related resources

https://developer.nexmo.com/voice/voice-api/ncco-reference#connect

benhayehudi:
Hi there @slorello89,

I don't think a change is necessary for this specific doc. The code example preceding it shows the format it is expecting as an array of objects. The text description in the table is also accurate that it is describing what it does, in that it connects to an endpoint.

slorello89:
I disagree - the wording is very confusing - I know the first time I read it I thought it was referring to a single endpoint object as that's how I originally coded it. I'd say it should read like:

Array of `endpoint` objects to connect to. See [available endpoint types](https://developer.nexmo.com/voice/voice-api/ncco-reference#endpoint-types-and-values)
> Note: Supports a maximum of 1 endpoint object.

benhayehudi:
> I disagree - the wording is very confusing - I know the first time I read it I thought it was referring to a single endpoint object as that's how I originally coded it. I'd say it should read like:

If it's confusing for you then it's going to be confusing to others as well. That's a good enough reason to modify it. I'll raise a pull request this morning with a change as you suggest here. Thanks for raising it, Steve!

> Array of `endpoint` objects to connect to. See [available endpoint types](https://developer.nexmo.com/voice/voice-api/ncco-reference#endpoint-types-and-values)
> 
> > Note: Supports a maximum of 1 endpoint object.

benhayehudi:
Addressing this in #2853

PR #2853, opened 2020-05-28
Define connect ncco action requirement explicitly

benhayehudi:
## Description

re: #2849 

This explicitly mentions in the `connect` action that an array of 1 `endpoint` object is required.

### DIFF
--- a/_documentation/en/voice/voice-api/ncco-reference.md
+++ b/_documentation/en/voice/voice-api/ncco-reference.md
@@ -135,7 +135,7 @@ You can use the following options to control a `connect` action:
 
 Option | Description | Required
 -- | -- | --
-`endpoint` | Array of `endpoint` objects to connect to with support for a **maximum** of 1 `endpoint` objects. [Available endpoint types](#endpoint-types-and-values) | Yes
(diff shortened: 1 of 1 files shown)


_documentation/en/voice/voice-api/ncco-reference.md
```diff
@@ -135,7 +135,7 @@ You can use the following options to control a `connect` action:
 
 Option | Description | Required
 -- | -- | --
-`endpoint` | Connect to a single endpoint. [Available endpoint types](#endpoint-types-and-values) | Yes
+`endpoint` | Array of `endpoint` objects to connect to with support for a **maximum** of 1 `endpoint` objects. [Available endpoint types](#endpoint-types-and-values) | Yes
```
tbedford, line 138: > Array of `endpoint` objects to connect to with support for a **maximum** of 1 `endpoint` objects.

Perhaps better as: 

Array of `endpoint` objects to connect to. Currently supports a **maximum** of one `endpoint` object.

benhayehudi: Thanks, @tbedford! Addressed in latest commit.

PR #2870, opened 2020-06-01
Table outlining difference between sandbox and Messages API

slorello89:
## Description

Adding a table at the end of the Messages API Sandbox page showing the differences in feature support between the Sandbox and the Messages API and the

### DIFF
--- a/_documentation/en/messages/concepts/messages-api-sandbox.md
+++ b/_documentation/en/messages/concepts/messages-api-sandbox.md
@@ -108,9 +108,9 @@ Namespace | Name | Languages
 ----|----|----
 `9b6b4fcb_da19_4a26_8fe8_78074a91b584` | `verify` | English, Korean, Japanese, Italian
 
-## Sandbox Supported Features
+## Sandbox supported features
 
-The table below shows a comparison between the features supported in the Messages Sandbox vs the full Messages API
+The following table shows the supported features of the Messages Sandbox compared to the production Messages API:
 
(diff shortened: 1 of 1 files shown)


_documentation/en/messages/concepts/messages-api-sandbox.md
```diff
@@ -107,3 +107,29 @@ At the moment the following WhatsApp templates can be used with the Messages API
 Namespace | Name | Languages
 ----|----|----
 `9b6b4fcb_da19_4a26_8fe8_78074a91b584` | `verify` | English, Korean, Japanese, Italian
+
+## Sandbox Supported Features
```
tbedford, line 111: Minor, but subheadings would have case like: "Sandbox supported features".

_documentation/en/messages/concepts/messages-api-sandbox.md
```diff
@@ -107,3 +107,29 @@ At the moment the following WhatsApp templates can be used with the Messages API
...
+
+The table below shows a comparison between the features supported in the Messages Sandbox vs the full Messages API
```
tbedford, line 113: Terminate sentence before tables, codeblocks etc. with a colon. Also, avoid "vs" and other such abbreviations especially Latin, as it makes translation less convenient. Also, avoid using 'above' and 'below' as it might not always be the case. You could say something like:

"The following table shows the supported features of the Messages Sandbox compared to the production Messages API:"

PR #2877, opened 2020-06-03
ASR NCCO and Webhook reference updates

benhayehudi:
* Note that either `speech` or `dtmf` is required in the `input` NCCO action
* Include `speech` in example of fields sent back in event webhook for `input`

### DIFF
--- a/_documentation/en/voice/voice-api/webhook-reference.md
+++ b/_documentation/en/voice/voice-api/webhook-reference.md
@@ -360,5 +360,6 @@ The event endpoint will also receive webhooks in the event of an error. This can
 Field | Example | Description
  -- | -- | --
 `reason` | `Syntax error in NCCO. Invalid value type or action.` | Information about the nature of the error
+`error` | `ERR1: Failed to analyze audio` | This field present for automatic speech recognition
 `conversation_uuid` | `CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this conversation
 `timestamp` | `2020-01-01T12:00:00.000Z` | Timestamp (ISO 8601 format)
--- a/_documentation/en/voice/voice-api/webhook-reference.md
+++ b/_documentation/en/voice/voice-api/webhook-reference.md
@@ -308,6 +308,7 @@ Field | Example | Description
 `from` | `447700900000` | The number the call came from
 `to` | `447700900000` | The number the call was made to
 `dtmf` | `42` | The buttons pressed by the user
+`speech` | `sales` | The text spoken by the user, a complete example payload shown in [NCCO Reference](/voice/voice-api/ncco-reference#speech-recognition-settings)
 `timed_out` | `true` | Whether the input action timed out: `true` if it did, `false` if not
 `uuid` | `aaaaaaaa-bbbb-cccc-dddd-0123456789ab` | The unique identifier for this call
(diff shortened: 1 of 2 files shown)


_documentation/en/voice/voice-api/webhook-reference.md
```diff
@@ -308,6 +308,7 @@ Field | Example | Description
 `from` | `447700900000` | The number the call came from
 `to` | `447700900000` | The number the call was made to
 `dtmf` | `42` | The buttons pressed by the user
```
slorello89, line 310: This needs to be updated to reflect the new object - should also be clear that it's an object if you passed in the dtmf object to the ncco

benhayehudi: We don't discuss the data types of any of the parameters in this document. Can you clarify what you are intending here in this specific document?

slorello89: the implication of how this is set in the table is that it's posted as a string - which in the new model of course it's not - if you want to point to the sample payload back in the ncco-reference that may be worth doing - but the way it's laid our here is misleading.

benhayehudi:
> @dragonmantank pointed out [Nexmo/nexmo-dotnet#203 (comment)](https://github.com/Nexmo/nexmo-dotnet/pull/203#discussion_r434283527) that there's also an error in the speech object - I'm assuming this is also possible in the dtmf object? Not sure that the error cases have been fully mapped out - not sure they need to be or not the field ought to be documented though

That's a good point, @slorello89. I'm going to take that aspect out of this specific pull request. As seen at https://developer.nexmo.com/voice/voice-api/webhook-reference#errors we document very broadly errors in the aforementioned document, but not to any degree of specificity. Let's discuss this specific issue outside of the context of this particular PR.

slorello89:
If `error` is a field of the webhook it should be documented - It doesn't necessarily need to be particularly granular - but if it's an optional part of the payload it should be called out.

_documentation/en/voice/voice-api/webhook-reference.md
benhayehudi: Can I encourage us to be careful with our language? Misleading means deliberate deceit, and no one is intending to deliberately deceive anyone, @slorello89.

The second column shows the primary value relevant to the parameter, it does not show the data structure where it is found. For example, the number in the `dtmf` parameter in the document is accessed inside `digits`, but the document's intention is to show you a small sample of what data you are interested in looks like. 

Nonetheless, I can include a link to the example payload as I did for the `speech` parameter.

benhayehudi:
> If `error` is a field of the webhook it should be documented - It doesn't necessarily need to be particularly granular - but if it's an optional part of the payload it should be called out.

That's a fair point. I'm conferring with some other folks to get their insight. The need to document the errors of all of our APIs is a known one and has come up before. We have a lot of work to do to make all the possible errors known to our users. 👍

benhayehudi:
I see another `error` message when audio was unable to be analyzed in addition to the one @dragonmantank uncovered:

```
"speech": {
  "error": "ERR1: Failed to analyze audio"
 }
```

Adding an `error` description to the webhook errors table per @slorello89 suggestion.

tbedford:
It was requested by @benhayehudi that I review and merge. However, I can't as changes requested by @slorello89 have not been re-approved. Looking through the comments it appears that Steve's requested changes have been implemented. So, in order to merge I have to dismiss that review I believe.

PR #2885, opened 2020-06-04
adding special IIS Express instructions for ngrok

slorello89:
## Description

Using IIS express with Ngrok can be a bit hinky if you don't set the host-header option in ngrok - I've added a description for how to properly use ngrok with IIS Express.

### DIFF
--- a/_documentation/en/tools/ngrok.md
+++ b/_documentation/en/tools/ngrok.md
@@ -41,9 +41,9 @@ Note that each time you start up the Ngrok tunnel, you will receive a new URL so
 
 ## Usage with IIS Express
 
-IIS Express rejects inbound requests from ngrok if the host header does not match the `bindingInformation` pattern in its configuration. Therefore you will need to configure ngrok to change the host name in the header to match the IIS Express configuration.
+IIS Express rejects inbound requests from Ngrok if the host header does not match the `bindingInformation` pattern in its configuration. Therefore you will need to configure Ngrok to change the host name in the header to match the IIS Express configuration.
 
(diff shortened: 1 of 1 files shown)


_documentation/en/tools/ngrok.md
```diff
@@ -39,6 +39,33 @@ Connections ttl opn rt1 rt5 p50 p90
...
+
+This can be done using ngrok's `host-header` option see the following command for how this might be done for port 3000:
+
+``` shell
+ngrok http 3000 --host-header="localhost:3000"
+```
+
+If you are doing this for your testing and need the original host for something, you can retrieve it from the `X-Original-Host` header.
+
+### Point IIS Express at the correct port
```
tbedford, line 54: Point -> Configure? Generally avoid 'point' as it might throw the translators. Choose something like 'configure' or 'redirect to' perhaps in such situations?

_documentation/en/tools/ngrok.md
```diff
@@ -39,6 +39,33 @@ Connections ttl opn rt1 rt5 p50 p90
...
+
+Validate that IIS Express is listening on port 3000, or whichever port you've decided to point ngrok at, by opening up your `properties/launchSettings.json` and making sure the port number on `iisExpress` -> `applicationUrl` matches the port you've instructed ngrok to forward to. You should also disable SSL on IIS Express by setting the `sslPort` to 0. See the following valid configuration for IIS Express from `launchSettings.json`:
```
tbedford, line 56: I wasn't sure what "`iisExpress` -> `applicationUrl`" means - is the arrow a special notation in this case? Also 'instructed' could change to 'configured'.

PR #2907, opened 2020-06-11
remove extraneous Copy text in copy to clipboard action

benhayehudi:
## Description

Remove the additional "Copy" word added to each copy to clipboard action

### DIFF
--- a/app/webpacker/javascript/code_snippet_events/index.js
+++ b/app/webpacker/javascript/code_snippet_events/index.js
@@ -5,7 +5,7 @@ export default () => {
     // Track copy to clipboard usage
     var clipboard = new Clipboard('.copy-button',{
         text: function(trigger) {
-            return $(trigger).next().text();
+            return $(trigger).next().text().replace('Copy', '');
         }
     });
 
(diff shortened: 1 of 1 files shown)


app/webpacker/javascript/code_snippet_events/index.js
```diff
@@ -5,7 +5,7 @@ export default () => {
     // Track copy to clipboard usage
     var clipboard = new Clipboard('.copy-button',{
         text: function(trigger) {
-            return $(trigger).next().text();
+            return $(trigger).next().text().replace('Copy', '');
```
fabianrbz, line 8: @benhayehudi I don't think this is the right solution, because you would be stripping the word `Copy` from the snippet which might be part of it.
The issue here is that now Volta includes a new html tag with the word `Copy` in it that acts as a tooltip, and it's being picked up by `$(trigger).next().text()`.

The html looks like this 
Screenshot 2020-06-12 at 12 20 46

We can change the code to consider this by doing:

```javascript
return $(trigger).next().find('.main-code').text();
```

benhayehudi: Thanks, @fabianrbz. You're right that's a more durable solution, I'll amend.

PR #2899, opened 2020-06-09
Patrick emails rebrand

pmmcls:
## Description
Updated some email addresses and reverted changes to support@nexmo.com.
Added meta descriptions and changed Nexmo references to Vonage where appropriate on overview pages.

Once support gives green light to re-brand emails then I will do that in another commit/PR.

Not to be merged till 16th June due to new email addresses like smpp@vonage.com (Previously @ nexmo.com)

PR #2918, opened 2020-06-15
Emails rebrand

fabianrbz:
## Description

Rebased version of https://github.com/Nexmo/nexmo-developer/pull/2899

## Deploy Notes

Not to be merged till 16th June due to new email addresses like smpp@vonage.com (Previously @ nexmo.com)


pmmcls:
Looks good to me

_documentation/en/messages/overview.md
```diff
@@ -84,8 +85,8 @@ In this example you will need to replace the following variables with actual val
 
 Key | Description
 -- | --
-`NEXMO_API_KEY` | Nexmo API key which can be obtained from your [Nexmo Dashboard](https://dashboard.nexmo.com).
-`NEXMO_API_SECRET` | Nexmo API secret which can be obtained from your [Nexmo Dashboard](https://dashboard.nexmo.com).
+`NEXMO_API_KEY` | Vonage API key which can be obtained from your [Vonage API Dashboard](https://dashboard.nexmo.com).
```
benhayehudi, line 88: Should the API key and secret suggested variables also be changed to: `VONAGE_API_KEY` and `VONAGE_API_SECRET`?

pmmcls:
> This pull request looks fine to me, but I do want to just leave a general comment, that from a developer customer experience, this is still confusing. Vonage APIs don't exist as far as the code or experience is concerned:
> 
> * The SDK I use is called Nexmo
> * The class and method names in said SDK are called `Nexmo`
> * The developer portal is called developer.nexmo.com
> * If they are all Vonage APIs, why do I need an entirely separate credentialing and SDK and developer portal for the Vonage Video API?
> 
> I know these changes have their timeline, but I'd like to just say I would have liked to have seen them as part of a complete overhaul and integration, and not piecemeal like this.

Hi Ben, 
The rebrand work is continuing through Q3 for the content and plans are underway to unify the domains.  The code on the other hand is likely to take a lot longer and like everything it is all about priorities and tradeoffs.  The point of this change was to get the main overview pages - these the ones with most traffic and so from an SEO perspective the ones we needed to get changed first.

Patrick

_documentation/en/messages/overview.md
pmmcls: I left these as I didn't want any knock on impact.

benhayehudi:
> Hi Ben,
> The rebrand work is continuing through Q3 for the content and plans are underway to unify the domains. The code on the other hand is likely to take a lot longer and like everything it is all about priorities and tradeoffs. The point of this change was to get the main overview pages - these the ones with most traffic and so from an SEO perspective the ones we needed to get changed first.
> 
> Patrick

Thanks for the reply, Patrick. We can definitely continue this off the GitHub thread, if it makes sense. I'll only add that how you describe it above reinforces my point. We wish to reify the transition Nexmo -> Vonage, so we target the highest SEO content. Then, the developer lands on this newly transitioned SEO content, and wishes to actualize based on what they read, and are left with naming that doesn't match the rest of the experience. Yes, if they think for a few minutes they can probably aggregate the experience cognitively, but most uninvested not-yet-engaged developers would rather not expend that extra bit of cognitive energy to figure out a marketing dissonance, and rather go somewhere else where they don't need to do a "Switch and Replace" mentally every time they try to contrast and compare between the documentation portal and everything else.

Just my two cents and I completely value the work that went in here, it's really well done! I would have just like to have seen it roll out in a more consistent end-to-end fashion.

PR #2921, opened 2020-06-16
Devx 2433 use case new styles

fabianrbz:
## Description

Related [ticket](https://nexmoinc.atlassian.net/browse/DEVX-2433)
Updates the styles for `/use-cases`

## Deploy Notes

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,4 +1,7 @@
 en:
+  any_product: Any Product
+  any_language: Any Language
+  read_more: Read more
 
   menu:
     message-search: Message Search
@@ -83,6 +86,8 @@ en:
         get_an_api_key: Get an API key
         start_building: Start building for free
         use-cases: Use Cases
+      programming_languages_dropdown:
+        any_language: Any language
 
   code_snippets:
     copy_to_clipboard: Copy to Clipboard
--- a/config/products.yml
+++ b/config/products.yml
@@ -7,8 +7,8 @@ products:
     dropdown: true
   - 
     name: 'Subaccounts'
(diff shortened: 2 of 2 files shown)


app/views/tutorial/list.html.erb
```diff
@@ -1,14 +1,12 @@
 <h1>Tutorials</h1>
 
-<div class="Nxd-use-cases--filters">
-  <strong>Filter: &nbsp;&nbsp;</strong>
+<p>Get started with Nexmo with tutorials that will walk you through building a variety of practical applications</p>
```
benhayehudi, line 3: Should this be named to `Vonage` per the other rebrand PR?

fabianrbz: I think that's being done in the other PR, this one focuses on changing the page layout/design

mheap:
A couple of things:

* I expected that clicking anywhere on the card on https://nexmo-develo-devx-2433--m58jpu.herokuapp.com/tutorials would take me to the tutorial
* Clicking on a language on a use case filters to that language, but clicking a language on a tutorial takes you to that implementation

fabianrbz:
@mheap making the card clickable is not that easy, because we have nested elements that can be clickable, we could instead remove the `cursor: pointer` from it

mheap:
Yes please, remove `cursor: pointer`

On tutorials if I filter to Kotlin only, then click a tutorial, I'd expect to see the Kotlin implementation rather than the JS one

fabianrbz:
Included in https://github.com/Nexmo/nexmo-developer/pull/3036

PR #2932, opened 2020-06-18
update char length

hughhopkins:
WhatsApp updated the char length


hughhopkins:
duplicate of https://github.com/Nexmo/nexmo-developer/pull/2930 #sorry

PR #2942, opened 2020-06-22
Add fallback options to WebSockets guide

marklewin:
## Description

This addresses the following JIRA ticket: https://nexmoinc.atlassian.net/browse/DEVX-2554

Note that I have substantially rewritten the text provided by Victor. Please ensure that the meaning is unchanged.

### DIFF
--- a/_documentation/en/voice/voice-api/guides/websockets.md
+++ b/_documentation/en/voice/voice-api/guides/websockets.md
@@ -184,7 +184,9 @@ To receive this event, you must include the `eventType: synchronous` in your `co
 ]
 ```
 
-Our API will `POST` the event to your webhook at `eventURL`. You can optionally return a new NCCO in the response with the required fallback actions. For example, if you cannot establish a connection, you might want to play a message to the caller or transfer the call.
+Our API will send the event object to your webhook at `eventURL`. This is a `POST` request by default, but you can specify the request type in the `eventMethod` parameter of the `connect` action.
+
+You can then return a new NCCO in the response with the required fallback actions. For example, if you cannot establish a connection, you might want to play a message to the caller or transfer the call.
 
 If you do not return an NCCO, the next action in the original NCCO will be processed. If there are no further actions in the original NCCO, the call will complete.
 
@@ -204,9 +206,11 @@ In the event of a connection failure, the event `status` will be one of `failed`
 }
 ```
 
+You can return a new NCCO in your webhook response to handle the failed connection attempt.
+
 ### Websocket disconnected
 
-If the WebSocket connection is terminated by the application, your webhook will receive an event with `status` of `disconnected`:
+If the connection is dropped by your application, you will receive an event on your `eventUrl` webhook with a `status` of `disconnected`:
 
 ``` json
 {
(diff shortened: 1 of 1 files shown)


_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -157,6 +157,126 @@ You can send audio back into the call by writing binary messages to the WebSocke
...
+  "uuid": "aaaaaaaa-bbbb-cccc-dddd-0123456789ab",
+  "conversation_uuid": "CON-aaaaaaaa-bbbb-cccc-dddd-0123456789ab",
+  "status": "unanswered",
+  "timestamp": "2020-03-31T12:00:00.000Z"
+}
+```
+
+### Websocket disconnected
+
+If the WebSocket connection is terminated by the application, your webhook will receive an event with `status` of `disconnected`:
```
vshisterov, line 209: Please add the information that one can respond with NCCO to this `disconnected` event callback to make some fallback action (reconnection, transfer, whatever suitable for the use-case).

marklewin: Victor - I have described the ability to respond to an NCCO for either can't connect/disconnect at the top of this section?

_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -157,6 +157,126 @@ You can send audio back into the call by writing binary messages to the WebSocke
...
+        "type": "websocket",
+        "uri": "wss://example.com/socket",
+        "content-type": "audio/l16;rate=16000"
+      }
+    ]
+  }
+]
+```
+
+Our API will `POST` the event to your webhook at `eventURL`. You can optionally return a new NCCO in the response with the required fallback actions. For example, if you cannot establish a connection, you might want to play a message to the caller or transfer the call.
```
benhayehudi, line 187: Is the default `POST` and is it user definable in their application settings?

_documentation/en/voice/voice-api/guides/websockets.md
vshisterov: Ok, I'm just missing some logical pause/switch between `disconnected` payload (below) and the next phrase "if the disconnection is intentional...". One may understand it as "you should do a `PUT` request after receiving the `disconnected`". But probably it's just me :)

marklewin: I'm sorry @vshisterov , but I'm not sure I understand. Maybe I've got the wrong end of the stick entirely with this and I'm sure I could do a better job of explaining it with your help!

This is how I understand it:

- If the connection is dropped by the app, you get a `disconnected` event and you can fallback by providing a new NCCO.
- However, this happens whether the disconnection was unintentional (due to some problem) or you intentionally disconnected the websocket (as part of your business logic)
- You want to be able to fallback _only when the disconnect is unintentional_ , so a better approach to forcing a disconnect is to do a `PUT` request to hang up. If you do that, you can rely on the `disconnect` event being an unintentional disconnection and responding appropriately.

Is my understanding correct? And, if so, would something like the above better explain it?

_documentation/en/voice/voice-api/guides/websockets.md
marklewin: Is this something you can advise on please, @vshisterov ?

_documentation/en/voice/voice-api/guides/websockets.md
vshisterov: @marklewin yes it's absolutely correct 👍 
And I guess it'd be very helpful to add this explanation, it sounds super clear to me.

_documentation/en/voice/voice-api/guides/websockets.md
vshisterov: Yes, default `eventMethod` is POST and can be overridden in the connect action, see 
https://developer.nexmo.com/voice/voice-api/ncco-reference#connect -> `eventMethod`.

benhayehudi: That's what I thought :) In that case, I think we should add a caveat that Our API will by default `POST` the event, unless the HTTP verb was overridden in the `connect` action.

_documentation/en/voice/voice-api/guides/websockets.md
marklewin: Many thanks!

_documentation/en/voice/voice-api/guides/websockets.md
marklewin: 👍

benhayehudi:
I think this is a really good shape. @marklewin, thank you! @vshisterov, can you give it another look through before we merge?

PR #2944, opened 2020-06-22
Fix ASR Webhook Description

vshisterov:
## Description

Input event callback payload structure adjusted.

## Deploy Notes

### DIFF
--- a/_documentation/en/voice/voice-api/webhook-reference.md
+++ b/_documentation/en/voice/voice-api/webhook-reference.md
@@ -329,7 +329,7 @@ Field | Example | Description
 `error` | `ERR1: Failed to analyze audio` | Error/status message: `Speech was not enabled` (status) - input action was configured to capture only DTMF input; `Speech overridden by DTMF` (status) - input action was configured to accept both speech/DTMF and user pressed keys; `ERRX: ...` (error) - user speech could not be recognized.
 
 ##### Transcript text
-Name	Description
+Field | Example | Description
 -- | -- | --
 `text` | `sales` | Transcript text representing the words that the user spoke.
 `confidence` | `0.9405097` | The confidence estimate between 0.0 and 1.0. A higher number indicates an estimated greater likelihood that the recognized words are correct.
--- a/_documentation/en/voice/voice-api/webhook-reference.md
+++ b/_documentation/en/voice/voice-api/webhook-reference.md
@@ -324,7 +324,7 @@ Field | Example | Description
 
 Field | Example | Description
 -- | -- | --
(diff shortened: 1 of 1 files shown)


_documentation/en/voice/voice-api/webhook-reference.md
```diff
@@ -360,6 +382,5 @@ The event endpoint will also receive webhooks in the event of an error. This can
 Field | Example | Description
  -- | -- | --
 `reason` | `Syntax error in NCCO. Invalid value type or action.` | Information about the nature of the error
-`error` | `ERR1: Failed to analyze audio` | This field present for automatic speech recognition
```
benhayehudi, line 363: Why are we removing this `error` field from the example? Does it no longer exist?

_documentation/en/voice/voice-api/webhook-reference.md
```diff
@@ -307,13 +307,35 @@ Field | Example | Description
...
+Field | Example | Description
+ -- | -- | --
+`digits` | `42` | The buttons pressed by the user 
+`timed_out` | `true` | Whether the DTMF input timed out: `true` if it did, `false` if not
+
+#### Speech Recognition Results
+
+Field | Example | Description
+-- | -- | --
+`timeout_reason` | `end_on_silence_timeout` | Indicates if the input ended when user stopped speaking (end_on_silence_timeout), by max duration timeout (max_duration) or if the user didn't say anything (start_timeout)
```
benhayehudi, line 327:
```suggestion
`timeout_reason` | `end_on_silence_timeout` | Indicates if the input ended when user stopped speaking (`end_on_silence_timeout`), by max duration timeout (`max_duration`) or if the user didn't say anything (`start_timeout`)
```

_documentation/en/voice/voice-api/webhook-reference.md
vshisterov: Anyway it's not in the root of the payload, it's nested to `speech` object:
```
{
    "speech": {
        "error": "Speech was not enabled"
    }, ...
```
so now it's confusing. Also, it's inconsistent with other "error callbacks" (with `reason` param) which should be someday also covered in more details.

Let's keep it in `input` callback description (which I adjusted in this PR).

benhayehudi: Okay, I'm fine with that, just as long as we document it somewhere. I believe our new orientation should be every response in a payload that's discovered should be documented in our platform. We need to account for developers working in strongly typed paradigms, where these things unaccounted for can throw them off.

PR #2961, opened 2020-06-29
Add more content to Websockets guide

benhayehudi:
## Description

As Websockets prepare to go GA, this PR adds more content to the Websockets guide on NDP based on internal feedback.

### DIFF
--- a/_documentation/en/voice/voice-api/guides/websockets.md
+++ b/_documentation/en/voice/voice-api/guides/websockets.md
@@ -81,7 +81,7 @@ You can find all the data fields for an NCCO at the [NCCO Reference Guide](/voic
 
 ### First message
 
-The initial message sent on an established WebSocket connection will be text-based and contain a JSON payload, it will have the `event` field set to `websocket:connected` and detail the audio format in `content-type`, along with any other metadata that you have put in the `headers` property of the WebSocket endpoint in your NCCO `connect`, for example:
+The initial message sent on an established WebSocket connection will be text-based and contain a JSON payload, it will have the `event` field set to `websocket:connected` and detail the audio format in `content-type`, along with any other metadata that you have put in the `headers` property of the WebSocket endpoint in your NCCO `connect` action, for example:
 
 ``` json
 {
@@ -164,7 +164,7 @@ You will receive one event for each keypress and each event will contain only on
 
 You can send audio back into the call by writing binary messages to the WebSocket. The audio must be in the same format as described in the previous section. It is important that each message is 320 or 640 bytes (depending on sample rate) and contains 20ms of audio.
 
-You can send the messages at a faster than real-time rate and they will be buffered for playing at the Nexmo end. So for example, you can send an entire file to the socket in one write, providing the 320/640 byte per message restriction is observed. Nexmo will only buffer 1024 messages which should be enough for around 20 seconds of audio, if your file is longer than this you should implement a delay of 18-19ms between each message, or consider using the [REST API to play an audio file](/voice/voice-api/code-snippets/play-an-audio-stream-into-a-call/).
+You can send the messages at a faster than real-time rate and they will be buffered by our API platform for later playback. So for example, you can send an entire file to the socket in one write, providing the 320/640 byte per message restriction is observed. Nexmo will only buffer 1024 messages which should be enough for around 20 seconds of audio, if your file is longer than this you should implement a delay of 18-19ms between each message, or consider using the [REST API to play an audio file](/voice/voice-api/code-snippets/play-an-audio-stream-into-a-call/).
 
 
 ## Websocket Event Callbacks
--- a/_documentation/en/voice/voice-api/guides/websockets.md
+++ b/_documentation/en/voice/voice-api/guides/websockets.md
@@ -72,7 +72,7 @@ Field | Example | Description
  -- | -- | --
 `uri` | `wss://example.com/socket` | The endpoint of your WebSocket server that Nexmo will connect to
 `content-type` | `audio/l16;rate=16000` | A string representing the audio sampling rate, either `audio/l16;rate=16000` or `audio/l16;rate=8000`
-`headers` | { 'name': 'J Doe', 'age': 40 } | An object of key/value pairs with additional optional properties to send to your Websocket server, with a maximum length of 512 bytes.
+`headers` | `{ 'name': 'J Doe', 'age': 40 }` | An object of key/value pairs with additional optional properties to send to your Websocket server, with a maximum length of 512 bytes.
 
 You can find all the data fields for an NCCO at the [NCCO Reference Guide](/voice/voice-api/ncco-reference).
 
--- a/_documentation/en/voice/voice-api/guides/websockets.md
+++ b/_documentation/en/voice/voice-api/guides/websockets.md
@@ -66,11 +66,16 @@ To instruct Nexmo to connect to a WebSocket your application server must return 
 ]
 ```
 
-The `uri` identifies the endpoint of your WebSocket server that Nexmo will connect to.
+The specific data fields for webhooks are the following:
 
-To choose the sampling rate set the `content-type` property to `audio/l16;rate=16000` or `audio/l16;rate=8000` depending on if you need the data at 16kHz or 8kHz. Most real-time transcription services work best with audio at 8kHz.
+Field | Example | Description
+ -- | -- | --
+`uri` | `wss://example.com/socket` | The endpoint of your WebSocket server that Nexmo will connect to
+`content-type` | `audio/l16;rate=16000` | A string representing the audio sampling rate, either `audio/l16;rate=16000` or `audio/l16;rate=8000`
+`headers` | { 'name': 'J Doe', 'age': 40 } | An object of key/value pairs with additional optional properties to send to your Websocket server, with a maximum length of 512 bytes.
 
-You can send additional optional properties to your WebSocket server by adding key/value pairs to a `headers` property. The maximum length of the `headers` data is 512 bytes.
+You can find all the data fields for an NCCO at the [NCCO Reference Guide](/voice/voice-api/ncco-reference).
+
 
 ## Handling incoming WebSocket messages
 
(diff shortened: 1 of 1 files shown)


_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -66,11 +66,16 @@ To instruct Nexmo to connect to a WebSocket your application server must return
 ]
 ```
 
-The `uri` identifies the endpoint of your WebSocket server that Nexmo will connect to.
+The specific data fields for webhooks are the following:
 
-To choose the sampling rate set the `content-type` property to `audio/l16;rate=16000` or `audio/l16;rate=8000` depending on if you need the data at 16kHz or 8kHz. Most real-time transcription services work best with audio at 8kHz.
```
mheap, line 71: > Most real-time transcription services work best with audio at 8kHz.

This is valuable information that we should keep

_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -55,35 +55,42 @@ To instruct Nexmo to connect to a WebSocket your application server must return
...
+
+You can find all the data fields for an NCCO at the [NCCO Reference Guide](/voice/voice-api/ncco-reference).
 
-You can send additional optional properties to your WebSocket server by adding key/value pairs to a `headers` property. The maximum length of the `headers` data is 512 bytes.
 
 ## Handling incoming WebSocket messages
 
 ### First message
 
-The initial message sent on an established WebSocket connection will be text-based and contain a JSON payload, it will have the `event` field set to `websocket:connected` and detail the audio format in `content-type`, along with any other metadata that you have put in the `headers` property of the WebSocket endpoint in your NCCO `connect`. The `headers` property is not present on the JSON payload so the properties are at the top-level of the JSON. For example:
```
marklewin, line 79: Can we say "in your NCCO's `connect`action", just to provide clarity? Same with the bullet that follows this.

_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -158,7 +167,7 @@ You can send audio back into the call by writing binary messages to the WebSocke
 You can send the messages at a faster than real-time rate and they will be buffered for playing at the Nexmo end. So for example, you can send an entire file to the socket in one write, providing the 320/640 byte per message restriction is observed. Nexmo will only buffer 1024 messages which should be enough for around 20 seconds of audio, if your file is longer than this you should implement a delay of 18-19ms between each message, or consider using the [REST API to play an audio file](/voice/voice-api/code-snippets/play-an-audio-stream-into-a-call/).
```
marklewin, line 167: Can we find another way of saying "at the Nexmo end"? I don't think it's super clear and, of course, there's the whole Vonage thing ;)

Maybe something like:

"You can send messages faster than they can be processed in real time and they will be buffered by our API platform for later playback"

_documentation/en/voice/voice-api/guides/websockets.md
benhayehudi: 👍

_documentation/en/voice/voice-api/guides/websockets.md
benhayehudi: 👍

_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -55,35 +55,42 @@ To instruct Nexmo to connect to a WebSocket your application server must return
                     "address": {
                         "line_1": "Apartment 14",
                         "line_2": "123 Example Street",
-                        "city": "New York City"
-                    },
-                    "system_roles": [183493, 1038492, 22],
-                    "enable_auditing": false
+                        "city": "New York City",
+                        "system_roles": [183493, 1038492, 22],
```
vshisterov, line 59: I guess logically both `system_roles` and `enable_auditing` should be out of `address` object (as it was before the change)

_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -55,35 +55,42 @@ To instruct Nexmo to connect to a WebSocket your application server must return
...
 ## Handling incoming WebSocket messages
 
 ### First message
 
-The initial message sent on an established WebSocket connection will be text-based and contain a JSON payload, it will have the `event` field set to `websocket:connected` and detail the audio format in `content-type`, along with any other metadata that you have put in the `headers` property of the WebSocket endpoint in your NCCO `connect`. The `headers` property is not present on the JSON payload so the properties are at the top-level of the JSON. For example:
+The initial message sent on an established WebSocket connection will be text-based and contain a JSON payload, it will have the `event` field set to `websocket:connected` and detail the audio format in `content-type`, along with any other metadata that you have put in the `headers` property of the WebSocket endpoint in your NCCO `connect` action, for example:
 
 ``` json
 {
+    "headers": {
```
vshisterov, line 88: I'm not sure about that, in the event callback "headers" are definitely wrapped in the `headers` attribute, but in the WebSocket message, we didn't change it so I'd expect it to be on the top-level as it was described in the docs. Probably better to double check with experiment.

_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -114,8 +121,10 @@ This results in the following JSON in the first message on the WebSocket:
 {
     "event":"websocket:connected",
     "content-type":"audio/l16;rate=16000",
-    "language": "en-GB",
-    "callerID": "447700900123"
+    "headers": {
```
vshisterov, line 124: Same thing regarding `headers` param here (don't think it's presented in WebSocket message)

_documentation/en/voice/voice-api/guides/websockets.md
benhayehudi: Since `system_roles` and `enable_auditing` are examples of `headers` as we discussed, would you prefer just to remove them from the example?

_documentation/en/voice/voice-api/guides/websockets.md
benhayehudi: I just verified this again, and I got back the custom data sent in the initial `connect` action wrapped in a `headers` object:

```
{"headers"=>"{\"name\":\"Ben Test\",\"addl_data\":{\"line_1\":\"More info\",\"line_2\":\"And more info\",\"line_3\":\"Last line\"}}", "from"=>"Unknown", "to"=>"wss://redacted/cable", "uuid"=>"redacted", "conversation_uuid"=>"redacted", "status"=>"unanswered", "direction"=>"outbound", "timestamp"=>"2020-07-01T07:35:47.394Z"}
```

_documentation/en/voice/voice-api/guides/websockets.md
```diff
@@ -155,12 +157,16 @@ You will receive one event for each keypress and each event will contain only on
...
 
-You can send the messages at a faster than real-time rate and they will be buffered for playing at the Nexmo end. So for example, you can send an entire file to the socket in one write, providing the 320/640 byte per message restriction is observed. Nexmo will only buffer 1024 messages which should be enough for around 20 seconds of audio, if your file is longer than this you should implement a delay of 18-19ms between each message, or consider using the [REST API to play an audio file](/voice/voice-api/code-snippets/play-an-audio-stream-into-a-call/).
+You can send the messages at a faster than real-time rate and they will be buffered by our API platform for later playback. So for example, you can send an entire file to the socket in one write, providing the 320/640 byte per message restriction is observed. Nexmo will only buffer 1024 messages which should be enough for around 20 seconds of audio, if your file is longer than this you should implement a delay of 18-19ms between each message, or consider using the [REST API to play an audio file](/voice/voice-api/code-snippets/play-an-audio-stream-into-a-call/).
 
 
-## Websocket fallback options
+## Websocket Event Callbacks
```
vshisterov, line 163:
```suggestion
## WebSocket Event Callbacks
```

marklewin: Thanks for the suggestion @vshisterov , but as a general rule when writing docs I prefer to highlight what something _does_ rather than what it _is_. This helps people find solutions to what they want to achieve (in this case, provide fallback options) without necessarily knowing what the technical implementation of that solution is (webhook events).

Can you think of a better heading to subscribe what this is doing?

benhayehudi: @marklewin I think your suggestion makes sense to order it on what it does as something more discoverable. Since this section contains a bit on event callbacks in general, and mostly on fallback options, I broke it out with a subheader for the fallback options. I believe this accomplishes both further clarity and more discoverability.

vshisterov: Right, the idea was actually to add some info on event callbacks in general, not only "fallback" ones (we missed it). Btw, I guess we might even list them in the order they sent in case of "normal flow": `started` -> `ringing` -> `answered` -> `completed` (one trick here, `started`, `ringing`, `answered` for WS triggered almost simultaneously, so in some cases, they might be messed up in order)

benhayehudi: The idea from what I recall was to add some info on WebSockets call back info particularly here, and link back to the event webhook guide for more information on general callbacks. 

Which reminds me... we need to open a separate PR to add more info on what the event callback payload looks like for the webhook guide, I don't think it's as clear as it can be on that guide. Separate conversation though.

marklewin: > @marklewin I think your suggestion makes sense to order it on what it does as something more discoverable. Since this section contains a bit on event callbacks in general, and mostly on fallback options, I broke it out with a subheader for the fallback options. I believe this accomplishes both further clarity and more discoverability.

@vshisterov This doesn't make sense to me. You now have an H2 for "WebSocket Event Callbacks" with an H3 for "Fallback options", but all the headers that follow discuss fallback options but are at the same level as the "Fallback options" header. The basic event stuff could be linked to elsewhere or given its own section in the page.

benhayehudi: @marklewin and @vshisterov I tried to implement based on our conversation this morning. Please feel free to directly amend if I missed something. 🙏

vshisterov: 👍

marklewin: LGTM @benhayehudi !

benhayehudi: Waiting for your checkmark! 😄

PR #2962, opened 2020-06-29
Upgrade to new OAS renderer

mheap:
## Description

Update to the new OAS renderer

## Deploy Notes

This currently points at a branch. We should tag and release a new version before this is merged


app/views/layouts/open-api.html.erb
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html lang="en">
+  <%= render partial: 'layouts/partials/head' %>
```
fabianrbz, line 3: @mheap do we need this new layout?

mheap: Yes, it adds a body class for scoping some selectors that are only required when running as part of NDP (not standalone)

Gemfile.lock
```diff
@@ -316,7 +321,7 @@ GEM
     notiffany (0.1.3)
       nenv (~> 0.1)
       shellany (~> 0.0)
-    oas_parser (0.25.1)
+    oas_parser (0.25.2)
```
fabianrbz, line 324: @mheap can you lock this to `0.25.1`

fabianrbz:
@mheap it looks good. We need to add `Tooltip` to Volta here though because I'm seeing an error in the console.
The initialization is done [here](https://github.com/Nexmo/nexmo-developer/blob/master/app/webpacker/packs/application.js#L107) but the extra libraries are [missing](http://volta-core.vonage.com/storybook/?path=/story/setup-javascript--initializing).

mheap:
@fabianrbz All addressed and ready for another review

PR #2983, opened 2020-07-06

NexmoDev:


marklewin:
Change to be made as part of config changes in NDP

PR #2984, opened 2020-07-06
Inclusive language

tbedford:
## Description

Make language more inclusive.


tOverney:
Sorry, I'm coming out of nowhere but honest question, why is `native` a non-inclusive word?

tbedford:
We generally try to avoid it because it's potentially divisive and socially charged. Further info here: https://developers.google.com/style/inclusive-documentation

PR #3003, opened 2020-07-09
Update clientsdk image paths

fabianrbz:
## Description

For screenshots location of files has moved from `public/assets/images` to `public/meta`.


tbedford:
It does appear that some images are completely missing!

PR #3036, opened 2020-07-17
Bump station

fabianrbz:
## Description

## Deploy Notes


config/business_info.yml
```diff
@@ -15,12 +16,9 @@ header:
     sign-up:
       path: https://dashboard.nexmo.com/sign-up
       text:
-        - '{ Try it free }'
-        - '< Try it free />'
-        - '{ Get an API key }'
-        - '< Get an API key />'
-        - '{ Start building for free }'
-        - '< Start building for free />'
+        - try-it
```
bencgreenberg, line 19: Are we removing all the other texts here on purpose?

fabianrbz: yeah, these are the keys for the actual texts in `config/locales/<language>.yml`

mheap: This isn't a like for like change. Part of the A/B test is the surrounding characters

fabianrbz: @mheap @bencgreenberg that's being handled in `station`

PR #3052, opened 2020-07-22
SEO  https://jira.vonage.com/browse/CON-611

pmmcls:
To main index page: Add meta description, title and H1 that are SEO optimised. https://jira.vonage.com/browse/CON-611
Google indexing currently shows random content from the page.

### DIFF
--- a/_documentation/en/index.md
+++ b/_documentation/en/index.md
@@ -1,5 +1,7 @@
 ---
-title: Documentation
+title: API Development Tools
+h1: Vonage API Development Tools
+description: Get the API development tools you’re looking for from Vonage. You’ll find guides, code snippets, and more to develop connected applications with Vonage APIs.
 ---
 
 # Documentation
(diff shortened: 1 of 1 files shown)


bencgreenberg:
I don't believe we currently render in `station` a SEO `h1` `<head>` tag. @fabianrbz should we add a ticket to support this?

fabianrbz:
@pmmcls @bencgreenberg @i5aac these changes actually apply to `developer.nexmo.com/documentation` and not the landing page which is the one with the issue.
I think we need to do some changes in `station` in order to support this.

Regarding the `<h1>`, We currently have `Connected Code` and no way to customize it.
I'll move the landing page to NDP so we can customize it and add support for the meta description.

pmmcls:
> @pmmcls @bencgreenberg @i5aac these changes actually apply to `developer.nexmo.com/documentation` and not the landing page which is the one with the issue.
> I think we need to do some changes in `station` in order to support this.
> 
> Regarding the `<h1>`, We currently have `Connected Code` and no way to customize it.
> I'll move the landing page to NDP so we can customize it and add support for the meta description.

ok please revert this then.  I'll put the required text in the ticket you created.

PR #3068, opened 2020-07-29
Record a Conversation Can Have a Custom eventUrl and eventMethod

bencgreenberg:
As shown in all of our record a conversation code snippets, the `conversation` `record` option can also accept a custom `eventUrl` and `eventMethod` to send the recording URL to. We have previously not made that explicit in the NCCO guide to match the code snippets on NDP. This addresses that.


bencgreenberg:
@marklewin I meant to tag you as a reviewer, not an assignee, now I seem not to be able to untag you in that category! 😁

PR #3084, opened 2020-08-07
Remove deprecated fields from Numbers API update snippets

marklewin:
## Description

Rewrote the [code snippet spec](https://github.com/Nexmo/code-snippet-specs/blob/master/numbers/update-number.md) and updated the associated snippets to remove `MESSAGES_APPLICATION_ID` option (deprecated) and use `app_id` instead, as per: https://nexmoinc.atlassian.net/browse/DEVX-3431.

The only snippet that does not yet demonstrate using `app_id` is the Java one, because this is dependent on work in the Java SDK described [here](https://nexmoinc.atlassian.net/browse/DEVX-3609). However, I have raised this PR because I'd like to the majority of the snippets updated and the java one, while incomplete, is not incorrect. 

When this is approved I will raise a ticket to deal with the Java snippet update.

Please ensure that all snippets in `/numbers/code-snippets/update` display correctly and the variable placeholders are described correctly.


tbedford:
Any idea why so many PHP code snippets have changed?

lornajane:
I think those are updates for the v2.2 PHP SDK release but check the git history on that repo, it's probably a more reliable information source than me

PR #3086, opened 2020-08-10
Add lighthouse action.

fabianrbz:
## Description

## Deploy Notes


mheap:
@fabianrbz Do we want this to run on every PR?

fabianrbz:
@mheap Ideally we should run this in `station` - for every PR first, we can scope it later -. I added it here because we don't have review apps in `station` yet.

mheap:
Let's reduce how often they run as most markdown changes won't have an impact.

Running when the `Gemfile.lock` is updated makes sense to me

PR #3088, opened 2020-08-10
add developer preview to navigation for programmable sip

bencgreenberg:
## Description

## Deploy Notes


bencgreenberg:
@tbedford I also now noticed that the `Redact` label is called `Dev Preview` not `Developer Preview`. It does not have a color associated with it. I have just changed the new one to `Dev Preview` as well. Let's see what color it renders in.

PR #3099, opened 2020-08-13
Rename Nexmo Client SDK

tbedford:
## Description

Rename Nexmo Client SDK to Client SDK.

### DIFF
--- a/_tutorials/en/client-sdk/install-node-client-sdk.md
+++ b/_tutorials/en/client-sdk/install-node-client-sdk.md
@@ -1,9 +1,9 @@
 ---
-title: Install the Client SDK
+title: Install the Client SDK for JavaScript
 description: Install the Client SDK to get the latest functionality for building Programmable Conversation apps.
 ---
 
-If you are planning to use JavaScript to develop your application, you'll need to install (or update) the latest version of the Client SDK.
+If you are planning to use JavaScript to develop your application, you'll need to install (or update) the latest version of the Client SDK for JavaScript.
 
 Create a new project directory and change into it.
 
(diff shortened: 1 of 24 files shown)


_tutorials/en/client-sdk/install-node-client-sdk.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: Install the Nexmo Client SDK
-description: Install the Nexmo Node Client SDK to get the latest functionality for building Programmable Conversation apps.
```
marklewin, line 3: Did you mean to exclude 'Node' here and just refer to the Client SDK rather than the Node Client SDK?

config/tutorials/en/app-to-app/javascript.yml
```diff
@@ -5,7 +5,7 @@ products:
 
 introduction: 
   title: Introduction to this task
-  description: This task shows you how to use make a voice call from a web app to another web app using the JavaScript Client SDK.
```
marklewin, line 8: Same with this: did you mean to refer to the language too?

_tutorials/en/client-sdk/install-node-client-sdk.md
tbedford: Well spotted. It should be Client SDK for JavaScript to be correct.

config/tutorials/en/app-to-app/javascript.yml
tbedford: Well spotted. It should be Client SDK for JavaScript to be correct. However, as long as we're explicit at least once most of the time it can just be Client SDK.

PR #3130, opened 2020-08-24
Removed javascript.md from Make a phone call Getting Started

conshus:
## Description

Removed javascript.md from Make a phone call Getting Started

## Deploy Notes


pardel (approved):
LGTM

PR #3135, opened 2020-08-26
Devx 3868 remove receive a call getting started js

conshus:
## Description

Updated code and documentation to include hang-up and reject call actions and to include Alice as the user.
Removed  In-App Voice - Getting Started - Receive a Phone Call - JS Version

## Deploy Notes


igorwojda:
@conshus You need to add redirect to `config/stitch-redirects.yml` to fix the `check-redirects` (check around line 90 in this file to see how iOS does it)

conshus:
> @conshus You need to add redirect to `config/stitch-redirects.yml` to fix the `check-redirects` (check around line 90 in this file to see how iOS does it). When we are deleting pages we have to make sure that old links are redirected instead of showing 404.

Looked like everything passes now, but the redirect doesn't look like it's working: https://nexmo-developer-pr-3135.herokuapp.com/client-sdk/in-app-voice/getting-started/receive-phone-call/javascript. 

Thanks @igorwojda.

tbedford (approved):

PR #3149, opened 2020-09-01
Devtv week of aug31

ruskibenya:
## Description

There were some updates to the content for the New Stream, Who Dis? episode for this Thursday. Needed to add that information.

## Deploy Notes


ruskibenya:
I've updated to include the accessibility issues brought up by @lornajane when using [Vimium Chrome Extension](https://vimium.github.io/). Solution was to add ARIA role="tab" to each tab. Thanks @mheap <3

PR #3159, opened 2020-09-04
Fixed bad version in PHP dependencies

dragonmantank:
## Description

During the PHP v2.2.0 release, two files got left referencing `nexmo/client:^2.2`, which was never released (it never needed to be). This changes those files back to just a generic `nexmo/client` reference like all the other files. 

## Deploy Notes

No special deploy notes, this is a typo fix


slorello89 (approved):
LGTM

PR #3160, opened 2020-09-04
Cleaned up various run calls, line number fixes

dragonmantank:
## Description

Based on a support request ticket, it was noticed some of the run commands for various PHP scripts had not been updated. This makes them also match where possible.

There's also a line fix, because one of the snippets was offset. 

## Deploy Notes

No special notes, mostly text changes.


marklewin (approved):
Checked in demo app and LGTM

PR #3173, opened 2020-09-10
Add missing redirect

marklewin:
## Description

Add missing redirect resulting from use case being converted into a tutorial


fabianrbz (approved):
:shipit:

PR #3193, opened 2020-09-17
updates on the event cards to reflect new name and timezone style dec…

ruskibenya:
…isions

## Description

Some new style choices from the Community team regarding naming of events and the display of timezones. Also updated to incorporate the location specific logos we received to differentiate the events.


fabianrbz (approved):
looks good to me

PR #3232, opened 2020-10-02
Improving docs - giving an example to list available conversations

srenauld:
## Description

While chatting with Luke over Discord, the idea to expand the docs a tiny bit came up, for two reasons:
- There is no concrete example on fully traversing a page iterator and gathering its resource objects (the closest there is is the [pagination](https://developer.nexmo.com/client-sdk/in-app-messaging/guides/handling-pagination) page, which does not actually show that much)
- Creating, joining and leaving conversations have examples, but listing available conversations, something that would be useful when re-syncing app state after a restart (for example) is not mentioned anywhere but in the actual SDK doc if you know where to look.

I've added a small snippet able to gather all available conversations and return them as the success path of a promise, without `async`/`await` (because not everybody has access to the fancy toys ;-) ).

## Deploy Notes

No fundamental changes (it's all markdown!)

### DIFF
--- a/_documentation/en/client-sdk/in-app-messaging/guides/inviting-members.md
+++ b/_documentation/en/client-sdk/in-app-messaging/guides/inviting-members.md
@@ -35,7 +35,7 @@ source: _tutorials_tabbed_content/client-sdk/guides/messaging/inviting-members/i
 In order to list all conversations available to a user, we need to make use of the various methods aimed at [pagination](/client-sdk/in-app-messaging/guides/handling-pagination), as the 
 conversations collection is paginated. A simple example to acquired this complete set as a list is available below:
 
-```tabbed_Content
+```tabbed_content
 source: _tutorials_tabbed_content/client-sdk/guides/messaging/inviting-members/listing
 ```
 
--- a/_documentation/en/client-sdk/in-app-messaging/guides/inviting-members.md
+++ b/_documentation/en/client-sdk/in-app-messaging/guides/inviting-members.md
@@ -30,6 +30,14 @@ This guide will introduce you to the following concepts.
 source: _tutorials_tabbed_content/client-sdk/guides/messaging/inviting-members/invited
 ```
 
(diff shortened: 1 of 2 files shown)


ruskibenya:
Maybe @conshus @kellyjandrews can take a look at this?

kellyjandrews:
Hey @conshus - can you take a look at this one?  It's a great client-sdk example.

conshus:
Hey @srenauld. Thanks for your contribution. Really great.
Couple of things:
- I think this may be better on it's own page. This page is for items around inviting users to a conversation.
- We try to avoid as much as possible to writing things like "simple" or "easy" because someone may look at the code and feel discouraged because it looks challenging to them. So something like, "Below is sample code on how to retrieve a complete list of conversations the logged-in user is a member (JOINED or INVITED state)." may work better.

- In the code: 
1. It looks like the initial call to get the Conversations is set to only get a single conversation (`page_size: 1,`). I'm curious as to why. I'm pretty sure you can set that to your `maxTotalRecords` and could probably remove all the current, total stuff and have .hasNext() dictate whether or not to do .getNext(). Little less logic a person has to maintain. Would be interested in hearing your thoughts.
2. I see you are creating an Array from the Map. Is that so you gather the conversations into one thing to send back? 
3. Once all the conversations have been retrieved, you send them or any errors back, but only have a "console.log". Could you expand that out a little. For example:
`
.then((conversations)=>{
     renderConversations(conversations);
})
.catch((error) => {
     console.error('Error in retrieving conversations: ', error);
});
`
Thanks again for the submission. Great job!

bencgreenberg:
What's the status of this? Is further progress happening here or shall we close?

srenauld:
Hey @bencgreenberg !

I forgot about this :-( I've been sidetracked and swamped on projects. I'll get to it this weekend!

bencgreenberg:
@kellyjandrews @conshus Circling back to this. Are we moving forward here or shall we close?

PR #3243, opened 2020-10-06
Add cURL code snippet for transferring a call with inline NCCO

juliacodez:
## Description

Add cURL code snippet for transferring a call with inline NCCO


lornajane (approved):
Yay! Thanks for adding this

PR #3256, opened 2020-10-09
Add current java version to code_languages.yml

mheap:
## Description

The latest version of the markdown renderer reads the Java version from `code_languages.yml` rather than hardcoding it

## Deploy Notes

N/A


mheap:
@fabianrbz I think we're in a catch 22 here. This needs merging for Station to pass, but needs the latest Station to recognise the new key

fabianrbz (approved):

PR #3258, opened 2020-10-09
fixing dependency names for .NET PHP and python

slorello89:
## Description

Some of the code snippets for .NET / PHP / and Python were displaying the wrong dependency - updating


dragonmantank:
PHP ones look good after pushing a fix that missed a few. Do we need to check the other three languages as well?

slorello89:
Spot check only showed those three with issues - worth a look.

slorello89:
@dragonmantank - I did a search of the entire `_examples` directory and replaced every reference to a nexmo package inside of a yml file

mheap:
Is https://github.com/Nexmo/nexmo-developer/blob/master/config/code_languages.yml#L55 still correct? Or is the new package name `Vonage`?

_examples/messaging/sms/send-an-sms-with-unicode/php.yml
```diff
@@ -2,7 +2,7 @@
 title: PHP
 language: php
 dependencies:
-    - nexmo
+    - vonage
```
dragonmantank, line 5: This should be `vonage/client`

dragonmantank (changes requested):
Looks like PHP just had one incorrect one, otherwise the PHP snippets look good

_examples/messaging/sms/send-an-sms-with-unicode/php.yml
slorello89: fixed

dragonmantank (approved):
Good for my snippets

tbedford:
@dragonmantank Can we merge?

tbedford:
@lornajane @superdiana - are we good to go on this one?

lornajane:
I'm not sure what I'm doing here :) I can't see that these changes would cause a problem, but I don't think they caught everything.

lornajane (approved):
Definitely a bunch of improvements here! Thanks people

PR #3264, opened 2020-10-12
Added additional spacing between Voyagers profiles

phazonoverload:
## Description

A small amount of additional CSS on the Voyagers landing page. As a result, new Voyagers profiles don't need to sit across multiple rows.


ruskibenya (approved):
✅ Looks good to me

PR #3271, opened 2020-10-13
pointing at the correct badges

slorello89:
## Description

needed to fix the paths for a couple of the badges to compensate for the rebrand as some of them had new client names.


marklewin (approved):
Checked in review app and looks good :)

PR #3288, opened 2020-10-18
Optimize images

JeffResc:
## Description

This pull request optimizes the images found in the repository using lossless compression. This means that the image quality will not be degraded. This will mean less space taken up by images along with faster loading times when viewing them. The overall file size has been reduced by around 23%.

## Deploy Notes

N/A


tbedford:
@JeffResc - Hi Jeff and thanks for the PR. Could you give us a little bit of detail on what you actually did? Did you run the images through a tool? Thanks.

tbedford:
@JeffResc - having had a look I think we can't accept this PR. The thing is a lot of the images that have been changed are actually on submodule repos. Also, for things like our SDK docs/conversation docs those docs are generated automatically from source code, and I think what's going to happen is your changes would be overwritten by our automated process anyway. Anyway, thanks for the PR. Best regards, Tony

PR #3291, opened 2020-10-19
Smartling Translations Workflow

bencgreenberg:
## Description

Check for new Smartling completed translations on a schedule, download them and raise a pull request with the new files.


bencgreenberg:
@fabianrbz, Something to consider when reviewing this is the way we currently initialize `TranslatorCoordinator` now with passing in a `frequency` integer, does that make sense? Shouldn't we dynamically generate it from either the number provided in the markdown file or in the product config, if the former is missing? This is currently unresolved and right now the Rake task will not work because it expects a manual `frequency` input, but considering it this afternoon, I think this needs to be refactored. What do you think?

.github/workflows/smartling_uploader.yml
```diff
@@ -0,0 +1,85 @@
...
+    strategy:
+      matrix: ${{needs.check-changes.outputs.matrix}}
+    steps:
+    - name: Run rake task
+      env:
+        POSTGRES_USERNAME: postgres
+        OAS_PATH: "_open_api/api_specs/definitions"
+        FORCE_COLOR: true
+      run: |
+        RAILS_ENV=test bundle exec nexmo-developer --docs=. --rake-smartling-upload ${{matrix}}
```
bencgreenberg, line 85: @fabianrbz - In the [upload rake task](https://github.com/Nexmo/station/pull/277/files#diff-e8238f8b64f6a3f747c678954ba51f53e0a2ef22d7d494528228e881f18f81f7R56) there is an argument for frequency, which as we know is used to find files with that translation frequency either specified in the markdown metadata for that file and if not there, in the products config.

How might we specify that in the workflow here?

.github/workflows/smartling_uploader_15_days.yml
```diff
@@ -0,0 +1,88 @@
+name: Smartling Uploader for Files Changed within 1 Day
```
bencgreenberg, line 1:
```suggestion
name: Smartling Uploader for Files Changed within 15 Days
```

.github/workflows/smartling_uploader_30_days.yml
```diff
@@ -0,0 +1,88 @@
+name: Smartling Uploader for Files Changed within 1 Day
```
bencgreenberg, line 1:
```suggestion
name: Smartling Uploader for Files Changed within 30 Days
```

.github/workflows/smartling_uploader_180_days.yml
```diff
@@ -0,0 +1,83 @@
+name: Smartling Uploader for Files Changed within 180 Days
+on:
+  push:
+    branches:
```
fabianrbz, line 4: @bencgreenberg this should run on a schedule, every 180 days

.github/workflows/smartling_uploader_5_days.yml
```diff
@@ -0,0 +1,83 @@
+name: Smartling Uploader for Files Changed within 5 Days
+on:
+  push:
+    branches:
+      - master
```
fabianrbz, line 5: same here but 5 days

fabianrbz (changes requested):

.github/workflows/smartling_uploader_5_days.yml
bencgreenberg: Done

.github/workflows/smartling_uploader_180_days.yml
bencgreenberg: Done

fabianrbz (approved):

PR #3321, opened 2020-10-29
Corrected typo

Annamaliakal:
## Description

Corrected a typo:
 is was changed to are; since the singular verb is, does not agree with the plural compound subject bug reports, bug fixes and feedback on the library.


lornajane (approved):
The small details make the difference! Thanks!

PR #3335, opened 2020-11-02
added week.init()

ruskibenya:
## Description
 Added Week.init() show to the schedule for the next few weeks. 

Thanks 💕


pardel:
@ruskibenya We'll need to have an additional show next week so that will push the 9th & 16th scheduled episodes a week later.

pardel (approved):

PR #3355, opened 2020-11-11
Update node.yml

shreyas44:
## Description
Update the dependencies for the node example to send a message on whatsapp in the `Messages API` section to include `@vonage/server-sdk`. The example uses `@vonage/server-sdk` however it's not mentioned in the prerequisites section.
Link to the page with example: https://developer.nexmo.com/messages/code-snippets/whatsapp/send-text

### DIFF
--- a/_examples/messages/whatsapp/send-text/node.yml
+++ b/_examples/messages/whatsapp/send-text/node.yml
@@ -1,6 +1,7 @@
 ---
 dependencies:
     - nexmo@beta
+    - @vonage/server-sdk
 code:
     source: .repos/vonage/vonage-node-code-snippets/messages/whatsapp/send-text.js
     from_line: 12


tbedford:
@shreyas44 - thanks for this and you are right - there is an inconsistency here, but this is not the right fix for now, although I totally see where you are coming from.

This problem would be true for all Messages and Dispatch API Node code snippets, so the PR would ideally deal with all of them. *However*, you still need to install `nexmo@beta` as Messages API (and Dispatch API) are still Beta. 

Having spoken to the developer I'll raise a ticket to capture this so we can address it as soon as possible.

tbedford:
Ticket raised here: #3361

shreyas44:
@tbedford As i continued exploring the Messages API, I came to realize that. A member in the Slack Community chat also mentioned that the messages sandbox currently only works with `nexmo@beta` 🙂

tbedford:
Yes, sorry about that, and thanks again for raising this issue. Hopefully we will have a fix soon.

PR #3358, opened 2020-11-12
Rebrand 2FA API diagram

marklewin:
https://nexmoinc.atlassian.net/browse/DEVX-4809

Partial PR - wanted to get your opinion on this before I update the other Short Code diagrams

## Review

* [Review page](https://nexmo-developer-pr-3358.herokuapp.com/messaging/us-short-codes/guides/2fa)


tbedford:
Love the new look. Couple of thoughts:

1. Split (1) and (2) into separate uni-directional arrows. 
2. (3) is actually two steps, one outward arrow from Vonage to handset and then a return of the PIN code from user to Vonage. You could split those out. I had to think twice when I first saw it combined.

Otherwise looks very useful.

marklewin:
Good suggestions, which I've tried to apply to the other diagrams in this section. This turned into a bit of an epic!

> Love the new look. Couple of thoughts:
> 
> 1. Split (1) and (2) into separate uni-directional arrows.
> 2. (3) is actually two steps, one outward arrow from Vonage to handset and then a return of the PIN code from user to Vonage. You could split those out. I had to think twice when I first saw it combined.
> 
> Otherwise looks very useful.

tbedford (approved):
Looks very smart!

PR #3373, opened 2020-11-17
remove beta label for ASR

bencgreenberg:
## Description

Auto Speech Recognition has now reached GA status and as such the Beta label is being removed from the docs.

## Deploy Notes


tbedford (approved):
Looks good in review app

PR #3392, opened 2020-11-20
devx-4905

tbedford:
## Description

Adds some words on `connection_timeout` and `socket_timeout` for voice webhooks in the Application overview.

## Review

* [Timeouts](https://nexmo-developer-pr-3392.herokuapp.com/application/overview#webhook-timeouts)

### DIFF
--- a/_documentation/en/application/overview.md
+++ b/_documentation/en/application/overview.md
@@ -126,13 +126,13 @@ When creating or updating an application these can be set directly, or updated a
 
 If these values are not specified when creating or updating the application, then default values are applied. The default values for these timeouts depend on the webhook concerned, as shown in the following table:
 
-Webhook | `connection_timeout` | `socket_timeout`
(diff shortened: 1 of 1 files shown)


tbedford:
@vshisterov @JamieChapman8462 - your feedback most welcome! Thank you!

JamieChapman8462 (approved):

marklewin (changes requested):
I wonder if we should say what these parameters are expecting, otherwise someone might think they are just booleans? Something like:

"If Vonage can't connect to the webhook URL, a timeout will occur after the specified time in (seconds?). Vonage then makes one additional attempt to connect to the webhook endpoint."

And is it really a timeout if "Vonage then makes one additional attempt to connect to the webhook endpoint."?

tbedford:
@marklewin - it was a bit rushed. Hopefully better now? Let me know. Thanks.

marklewin (approved):
Great!

PR #3395, opened 2020-11-23
Update SDK Docs

vonageCI:


tbedford (approved):
Just noting again that the branding and copyright notices are way out of date. I believe there was a ticket to correct this some time ago, but thought I'd mention it again. Thanks.

PR #3397, opened 2020-11-26
Add redirect to store.

fabianrbz:
## Description

Redirect `/store` to `https://apimanager.uc.vonage.com/store/`


mheap (changes requested):
This needs a test in ADP to make sure that it's never removed too

bencgreenberg:
Can we merge this here? The proposed test in Station will only pass once this redirect is in place.

fabianrbz:
Merged as part of #3402

PR #3410, opened 2020-12-01
Add redirects for old client-sdk tutorials

marklewin:
[DEVX-4912](https://nexmoinc.atlassian.net/browse/DEVX-4912)

The following are 404s and require redirects:

* /task/client-sdk/ip-messaging/authenticate-users
* /task/client-sdk/ios-in-app-messaging-chat/add-users-to-conversation

Redirected second link to a generic tutorial step, as this could be either Swift or Objective C (for iOS) - let the reader decide 😄


tbedford (approved):
Tested in review app. Looks good. Thanks Mark.

PR #3422, opened 2020-12-04
Add credentials to Smartling workflows

bencgreenberg:
## Description

* Add Smartling credentials to the Smartling GitHub workflows
* Run the download workflow once a week on Monday, not every day


.github/workflows/ci.yml
```diff
@@ -70,15 +70,15 @@ jobs:
         OAS_PATH: "_open_api/api_specs/definitions"
         FORCE_COLOR: true
       run: |
-        ./node_modules/.bin/mdspell -r -n -a --en-us '_documentation/en/**/*.md' '_partials/*.md' '_partials/**/*.md' '_modals/**/*.md' '_tutorials/**/*.md'
+        ./node_modules/.bin/mdspell -r -n -a --en-us '_documentation/en/**/*.md' '_partials/*.md' '_partials/**/*.md' '_modals/**/*.md' '_tutorials/en/**/*.md'
```
fabianrbz, line 73: @bencgreenberg can we add _use_cases/en to these checks too?

fabianrbz (changes requested):

.github/workflows/ci.yml
bencgreenberg: Yes, we can. However, that's a new change from our existing CI in master and running it locally I see multiple spelling and offensive language issues. I'd rather do that in a separate PR so we can address those without over-doing it in this single pull request.

fabianrbz (approved):

PR #3431, opened 2020-12-08
Voice API tutorial tidy up

marklewin:
Moved dotnet VAPI tutorials into `/voice/voice-api` folder, creating tutorial-specific subfolders and sharing common content.

There is no visible difference - this is just to make the tech writers' lives easier - but please check that the following tutorials still work:

- Build a voicemail with .NET
- Pay audio into call with .NET


bencgreenberg (approved):
This is great! 👍

PR #3466, opened 2020-12-22
exclude client SDK docs from Smartling upload workflows

bencgreenberg:
## Description

As the Client SDK docs are beta they are out of scope for auto-translation. This excludes them from the upload workflows.


mheap:
Closing in favour of a new `rake` task that can handle more complex upload rules

PR #3487, opened 2021-01-13
DEVX-5040: Remove extra newline.

fabianrbz:
## Description

It was being added while copying to clipboard.


marklewin (approved):
Tested in review app

PR #3502, opened 2021-01-20
Update text-to-speech.md

vshisterov:
## Description

List of `voiceName` (deprecated) parameter values and mapping to new `language`/`style` added for customer convenience.


.spelling
```diff
@@ -477,3 +477,106 @@ Subnet
 NATing
 VMWare
 ESXi
+Voice Name / Code
+Aditi
+Aditi
+Agnieszka
+Alessio
+Alva
+Amy
```
bencgreenberg, line 486: Do we need to add all these names here? I'd imagine some of these don't trigger a spelling error?

_documentation/en/voice/voice-api/guides/text-to-speech.md
```diff
@@ -56,6 +56,11 @@ NCCO example with specific voice language and style:
 
 ### Supported Languages
 
+> **Where are "voice names"?**
+> 
+> Previously, in order to set the desired language and voice style you had to set `voiceName` parameter with certain "name". You can continue using it unless you want to try another style or support new language. For your convenience, see [below](#legacy-voice-names) how voice names are mapped to `language` and `style`. Please note, not all the supported voices have corresponding `voiceName`; thus it's recommended to use `language` and `style` in all new or being updated apps. You can listen to the voice samples in the table below.
```
bencgreenberg, line 61:
```suggestion
> Previously, in order to set the desired language and voice style, you used the `voiceName` parameter with a certain "name" as the value. The `voiceName` parameter has been deprecated, and while you can continue to use it, it will not contain new styles and languages. For your convenience, see [below](#legacy-voice-names) for a mapping of the legacy `voiceName` parameter to `language` and `style`. Please note, not all the supported styles have a corresponding `voiceName`; thus it's recommended to use `language` and `style` in your application or service. You can listen to the voice samples in the table below.
```

bencgreenberg (commented):
Looks good, just left a few comments to review

.spelling
vshisterov: Not sure, to be honest... I was lazy to check, just saw 10+ errors with the names and added all of them together.

bencgreenberg: It's important we don't clutter the `.spelling` document unnecessarily. Would it be too much to ask you to please check? The output tells you specifically which word(s) caused the failure, so hopefully, it won't be too much work. 🙏🏻

bencgreenberg (approved):

PR #3513, opened 2021-01-25
devx-5029: client sdk landing page

abdulajet:
## Description

https://nexmo-developer-pr-3513.herokuapp.com/client-sdk


mheap:
The column span looks off to me. Better to have a consistent width and have a row with 2 entries at the end?

image

Also, Event Flow isn't really a concept, it's a guide

---

This section looks busy to me

image

Would it work better as a list similar to the Opentok docs?

image
 ---

Otherwise, this is looking great

PR #3514, opened 2021-01-25
updating the messenger linkage steps to reflect new process

slorello89:
## Description

Updating the process to link a facebook page. The process was moved into the dashboard in December.  And the old mesenger.nexmo.com URL is being deprecated - necessitating this update.


marklewin (approved):
LGTM

PR #3517, opened 2021-01-26
Update Documentation on legs acl path

JohnPeters7116:
## Description

New upcoming release of JS, Android and IoS sdks will have a new way to setup calls with prewarmed legs. To the client this will be transparent however there is a request now being sent to POST /legs endpoint. This has always been exposed but has not been in the documentation for acl paths so customers may need to add it.

https://nexmoinc.atlassian.net/browse/CSJ-1205

## Deploy Notes

This will need to be released before we can make the official release fo the prewarm leg changes in the sdks.

### DIFF
--- a/_documentation/en/conversation/guides/jwt-acl.md
+++ b/_documentation/en/conversation/guides/jwt-acl.md
@@ -71,6 +71,7 @@ In the previous section, you can see that the `acl` claim has `paths` object con
 | `/*/image/**`| Send and receive images.|
 | `/*/media/**`| Send and receive audio.|
 | `/*/knocking/**`| Start phone calls.|
+| `/*/legs/**`| Prewarm phone calls. |
 | `/*/push/**`| Receive push notifications|
 | `/*/devices/**`| Send push notifications.|
 | `/*/applications/**`| Upload push notification certificate.|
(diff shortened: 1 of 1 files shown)


_documentation/en/conversation/guides/jwt-acl.md
```diff
@@ -71,6 +71,7 @@ In the previous section, you can see that the `acl` claim has `paths` object con
 | `/*/image/**`| Send and receive images.|
 | `/*/media/**`| Send and receive audio.|
 | `/*/knocking/**`| Start phone calls.|
+| `/*/legs/**`| Prewarm phone calls. |
```
mheap, line 74: What does `prewarm` mean here? Can we add details to help customers understand?

JohnPeters7116: Yeah I wasn't sure of the language to use here. It's really about setting up a call (getting a call id) on the backend before exchanging media and handing control over to VAPI with knocking. This aligns PSTN calls with IP calls on the backend (also should also speed up media exchange/connection to freeswitch as well). But as far as the customer is concerned at this point in the sdk, all these things aren't really apparent to the customer (except for the improvement on call establishment time).

JohnPeters7116: Could we just say "Start phone calls" like for knocking?

JohnPeters7116:
Covered by another PR https://github.com/Nexmo/nexmo-developer/pull/3534

PR #3551, opened 2021-02-10
VAPI Detailed Call Status

vshisterov:
## Description

New `detail` attribute of the call status callbacks added.


bencgreenberg:
@vshisterov, to confirm this new field is only present in the specific events enumerated in the doc? I've tested in production and that seems to be the case

vshisterov:
> @vshisterov, to confirm this new field is only present in the specific events enumerated in the doc? I've tested in production and that seems to be the case

exactly

bencgreenberg (approved):
LGTM!

PR #3553, opened 2021-02-15
Adds legs path to jwt:generate CLI commands

conshus:
## Description

Searched through the docs for `jwt:generate` and added legs path.

Not sure how it happened, but it looks like there are some other changes I didn't make in the PR. Lines of text added that look like they were removed? I created my branch from main and pulled locally and checked out my new branch. Anyone have any ideas what happened?

## Deploy Notes


pardel (approved):
LGTM

PR #3556, opened 2021-02-16
Update SDK Docs

vonageCI:


fabianrbz:
@pardel @conshus should I merge this?

pardel:
I don't think this is right. The `this` additions don't make sense.

PR #3558, opened 2021-02-16
Update app-to-phone tutorial to be more explicit of its scope.

conshus:
## Description

Updated tutorial to more clearly state its goal and scope.

## Deploy Notes


pardel:
obsolete now

PR #3569, opened 2021-02-23
Standardised casing on a Voyagers header

phazonoverload:
One header had incorrect casing compared to others - this PR fixes that


myrsiningos (approved):
Change looks good

PR #3588, opened 2021-03-11
vonage-python-code-snippets Update

NexmoDev:
Added a bunch of missing code snippets, and yaml for snippets we had but were never added

### DIFF
--- a/_examples/account/secret-management/fetch-a-secret/python.yml
+++ b/_examples/account/secret-management/fetch-a-secret/python.yml
@@ -1,11 +1,10 @@
 ---
 dependencies:
     - vonage
-    - python-dotenv
 import_dependencies:
     source: .repos/vonage/vonage-python-code-snippets/secrets/fetch-a-secret.py
     from_line: 1
-    to_line: 11
+    to_line: 3
 client:
     source: .repos/vonage/vonage-python-code-snippets/secrets/fetch-a-secret.py
     from_line: 12
--- a/_examples/account/secret-management/fetch-a-secret/python.yml
+++ b/_examples/account/secret-management/fetch-a-secret/python.yml
@@ -1,0 +1,16 @@
+---
+dependencies:
+    - vonage
+    - python-dotenv
+import_dependencies:
+    source: .repos/vonage/vonage-python-code-snippets/secrets/fetch-a-secret.py
+    from_line: 1
+    to_line: 11
+client:
+    source: .repos/vonage/vonage-python-code-snippets/secrets/fetch-a-secret.py
+    from_line: 12
+    to_line: 12
+code:
+    source: .repos/vonage/vonage-python-code-snippets/secrets/fetch-a-secret.py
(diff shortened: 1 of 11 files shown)


dragonmantank:
Marking as draft as I have to set up the code snippet yamls

_examples/account/secret-management/fetch-a-secret/python.yml
```diff
@@ -0,0 +1,16 @@
+---
+dependencies:
+    - vonage
+    - python-dotenv
+import_dependencies:
+    source: .repos/vonage/vonage-python-code-snippets/secrets/fetch-a-secret.py
```
slorello89, line 6: I'd say the secrets & enviornment setup are not in the spirit of the `import_dependencies` section - I'd move that to the code or client sections importing should only be the `import`/`using`/`require` statements. They can even safely be excluded altogether - the variable names from the spec I can be fairly assumed to be fill in the blank.

_examples/number-insight/async-callback/python.yml
```diff
@@ -0,0 +1,13 @@
...
+    - vonage
+    - python-dotenv
+    - flask
+import_dependencies:
+    source: .repos/vonage/vonage-python-code-snippets/number-insight/async-callback/app.py
+    from_line: 1
+    to_line: 2
+code:
+    source: .repos/vonage/vonage-python-code-snippets/number-insight/async-callback/app.py
+    from_line: 4
```
slorello89, line 12: I don't think the snippet needs to print "Hello World" to the console

_examples/number-insight/async-trigger/python.yml
```diff
@@ -0,0 +1,16 @@
...
+    source: .repos/vonage/vonage-python-code-snippets/number-insight/ni-advanced-async-trigger.py
+    from_line: 1
+    to_line: 10
+client:
+    source: .repos/vonage/vonage-python-code-snippets/number-insight/ni-advanced-async-trigger.py
+    from_line: 13
+    to_line: 16
+code:
+    source: .repos/vonage/vonage-python-code-snippets/number-insight/ni-advanced-async-trigger.py
+    from_line: 18
```
slorello89, line 15: @fabianrbz - the code-highlighting looks really bad for this one - any thoughts?

slorello89 (changes requested):

_examples/number-insight/async-callback/python.yml
dragonmantank: Easier way would be to change the snippet after this is merged in, to avoid conflicts with this PR. I thought I had swapped them all the return blank 200's

_examples/account/secret-management/fetch-a-secret/python.yml
dragonmantank: Removed

slorello89 (approved):
LGTM 👍

PR #3673, opened 2021-05-12
Update decode jwt node example

kellyjandrews:
## Description

Updates the Decode JWT example for node using the new code snippet.

## Deploy Notes


dragonmantank (approved):
LGTM and page renders OK

PR #3693, opened 2021-06-04
DEVX-5579 Update Android Client SDK docs for Nexmo Android SDK 3.0.0

igorwojda:
https://jira.vonage.com/browse/DEVX-5579

- Update docs to be compatible with Nexmo SDK 3.0.0
- Update projects to remove Center reference

### DIFF
--- a/_tutorials/en/client-sdk/in-app-messaging/fetch-events/java.md
+++ b/_tutorials/en/client-sdk/in-app-messaging/fetch-events/java.md
@@ -76,7 +76,7 @@ To handle member related events (member invited, joined or left) you need to fil
 
 ```java
 private String getConversationLine(NexmoMemberEvent memberEvent) {
-    String user = memberEvent.getMember().getUser().getName();
+    String user = memberEvent.getEmbeddedInfo().getUser().getName();
 
     switch (memberEvent.getState()) {
         case JOINED:
@@ -97,7 +97,7 @@ Above method converts `NexmoMemberEvent` to a `String` that will be displayed as
 
 ```java
 private String getConversationLine(NexmoTextEvent textEvent) {
-    String user = textEvent.getFromMember().getUser().getName();
+    String user = textEvent.getEmbeddedInfo().getUser().getName();
     return user + "  said: " + textEvent.getText();
 }
 ```
--- a/_tutorials_tabbed_content/client-sdk/guides/messaging/leave-conversation/kotlin.md
+++ b/_tutorials_tabbed_content/client-sdk/guides/messaging/leave-conversation/kotlin.md
@@ -4,7 +4,7 @@ language: kotlin
 ---
 
 ```kotlin
-conversation.kick("memberName", object : NexmoRequestListener<Any> {
+conversation.kick("memberId", object : NexmoRequestListener<Any> {
     override fun onSuccess(p0: Any?) {
         Log.d("TAG", "User kick success")
     }
--- a/_tutorials_tabbed_content/client-sdk/guides/messaging/leave-conversation/kotlin.md
+++ b/_tutorials_tabbed_content/client-sdk/guides/messaging/leave-conversation/kotlin.md
@@ -4,15 +4,13 @@ language: kotlin
 ---
 
 ```kotlin
-conversation?.allMembers?.firstOrNull()?.let {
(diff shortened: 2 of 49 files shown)


_tutorials_tabbed_content/client-sdk/guides/unmute-call-member/kotlin.md
```diff
@@ -4,12 +4,16 @@ language: kotlin
...
-call?.myCallMember?.mute(false, muteListener)
+    override fun onSuccess(result: Void?) {
+        Timber.d("Member unmuted")
+    }
+}
 
-// Unmute whole call
-call?.mute(false)
+val nexmoMember = call?.myMember
+nexmoMember?.disableMute(muteListener)
```
VMadalin, line 18: This comment is a note, we should reconsider mute/earmuff because there is no consistency between the different SDKs especially because those actions are only applicable to the logged user, and the API shows that customers can mute all members, maybe we should expose those methods just on conversation/call. Here is the documentation of JS SDK: 

```
  /**
   * Mute your stream.
   *
   * @param {boolean} [mute] true for mute, false for unmute
   * @param {number} [streamIndex] stream index of the stream
   * @example <caption>Mute audio stream of your Member.</caption>
   * // Mute yourself
   * conversation.me.mute(true);
   *
   * // Unmute yourself
   * conversation.me.mute(false);
   *
   * @returns {Promise}
  */
```

For sure will be nice to be able to mute some else but this should be moderated. That logic, unfortunately isn't in place but could be an interesting feature to consider for Adi

_tutorials_tabbed_content/client-sdk/guides/messaging/leave-conversation/kotlin.md
```diff
@@ -4,15 +4,13 @@ language: kotlin
 ---
 
 ```kotlin
-conversation?.allMembers?.firstOrNull()?.let {
-    conversation.kick(it, object : NexmoRequestListener<Any> {
-        override fun onSuccess(p0: Any?) {
-            Log.d("TAG", "User kick success")
-        }
+conversation.kick("memberName", object : NexmoRequestListener<Any> {
```
VMadalin, line 7: Following the method here should be the `memberId` for the first parameter. 
```java
public void kick(@NonNull final String memberId, @NonNull final NexmoRequestListener<Void> listener)
```

_tutorials/en/client-sdk/in-app-messaging/fetch-events/java.md
```diff
@@ -97,7 +97,7 @@ Above method converts `NexmoMemberEvent` to a `String` that will be displayed as
 
 ```java
 private String getConversationLine(NexmoTextEvent textEvent) {
-    String user = textEvent.getFromMember().getUser().getName();
```
VMadalin, line 100: This part is a bit confusing but we have pending to improve this on the backend on the following ticket ([Member Event `From` should be the person who sent invite](https://jira.vonage.com/browse/APICS-2022)).
 
In order to explain better let's consider the following scenario we are logged with user User1 and our user (User1) gonna be invited by User2. Then we gonna receive a member event where:
- `getEmbeddedInfo`, `memberId`: gonna reference the user who receives the event (receiver) [User1]
- `invitedBy`, `fromMemberId`: gonna reference the user who send the invite, being the `invitedBy` the user name and `fromMemberId` the memberId [User2]

For example, if you want to display `User1 was invited by User2` on screen then you should have something similar to:
```kotlin
"${event.embeddedInfo.user.name} by $invitedBy"
```

As a note `invitedBy` is optional giving that if we are using PSNT then gonna be VAPI who invites the user.

PS: We should consider updating other files where we have the same code

VMadalin: We should consider probably updating our doc here https://github.com/nexmoinc/nexmoclient-sdk-android/blob/master/comms/src/main/java/com/nexmo/client/NexmoEventEmbeddedInfo.java#L14 because `event producer` is  ok for example if you enable media, but in the case of by invited by someone could be very confusing.

VMadalin (approved):

abdulajet (approved):
:shipit:

PR #3701, opened 2021-06-09
Upd messaging tutorial - fix crashes

igorwojda:
- Removes app crash by changing init order and running things on U thread (this is a quick fix, as this tutorials will be simplified)
- removed redundant (unused) step from app-to-app tutorial


abdulajet (approved):
lgtm

PR #3709, opened 2021-06-11
changed innerHTML to innerText to be safer

conshus:
## Description

changed innerHTML to innerText to be safer, just incase someone entered in JavaScript as their username. 😅
## Deploy Notes


abdulajet (approved):
lgtm

PR #3711, opened 2021-06-14
update docs link for iOS

abdulajet:
## Description

Updating the redirect of the iOS SDK docs to point at the symlink so the latest version of the docs is always shown

### DIFF
--- a/spec/requests/ios_spec.rb
+++ b/spec/requests/ios_spec.rb
@@ -8,11 +8,11 @@ RSpec.describe 'iOS docs', type: :request do
       get '/client-sdk/sdk-documentation/ios/ios'
 
       expect(last_response).to be_redirect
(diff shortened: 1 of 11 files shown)


spec/requests/ios_spec.rb
```diff
@@ -8,11 +8,11 @@
       get '/client-sdk/sdk-documentation/ios/ios'
 
       expect(last_response).to be_redirect
-      expect(last_response.headers['Location']).to eq('http://example.org/sdk/stitch/ios/index')
+      expect(last_response.headers['Location']).to eq('http://example.org/sdk/stitch/ios/latest/index')
```
abdulajet, line 11: @fabianrbz I wasnt sure if i needed to update the test file here?

fabianrbz (approved):

PR #3716, opened 2021-06-16
Adds code snippet and description to set a text status to seen.

conshus:
## Description

Had a question from a customer about the `text:seen` event not firing. Found a method on the TextEvent that will do that so I added a code example and some details.

Thing is that it's in a tabbed element so I guess the other platforms will need to be added at some point.

Thanks.

### DIFF
--- a/_documentation/en/client-sdk/in-app-messaging/guides/seen-receipt.md
+++ b/_documentation/en/client-sdk/in-app-messaging/guides/seen-receipt.md
@@ -19,6 +19,13 @@ This guide will make use of the following concepts:
 
 - **Conversation Events** - `text:seen` events that fire on a Conversation, after you are a Member
 
+## Set TextEvent to seen
+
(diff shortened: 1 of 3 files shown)


abdulajet (approved):

_documentation/en/client-sdk/in-app-messaging/guides/seen-receipt.md
```diff
@@ -19,6 +19,13 @@ This guide will make use of the following concepts:
 
 - **Conversation Events** - `text:seen` events that fire on a Conversation, after you are a Member
 
+## Set Text status to seen
+
+There is the `seen()` method on a TextEvent that will set its status to seen. The following code snippet will set a text's status to seen once a text event happens in the conversation.
```
conshus, line 24:
```suggestion
There is the `seen()` method on a `TextEvent` that will set its status to seen. The following code snippet will set a text's status to seen once a text event happens in the conversation.
```

PR #3777, opened 2021-08-12
Correct iOS app to app tutorials

abdulajet:
## Description

The text is no longer correct as the application uses a webhook server so the NCCO is now dynamic


pardel (approved):
LGTM

PR #3779, opened 2021-08-20
add mofi to devrel team

moficodes:
## Description

- Removing Igor from dev rel team list.
- Adding Mofi Rahman to dev rel team list.


pardel (approved):
LGTM

PR #3781, opened 2021-08-23
add git submodule instructions

moficodes:
update docker file to follow docker convention of uppercase instruction

## Description

a. For running locally the docker-compose instruction does not have instruction for git submodule. it is assumed that the user has already run that command from the previous section. But since this instructions are mutually exclusive we should not assume that the user is going to run them in order.

b. In docker file is best practice to have docker instructions in Uppercase.


moficodes:
@pardel

pardel (approved):

PR #3787, opened 2021-09-02
Update SDK Docs

vonageCI:


conshus:
For some reason, the change I made to include custom_data into the description is not showing in the deployment.
Also, I couldn't figure out how to get the header:
image
to include the custom_data field so it looks like:
image

PR #3796, opened 2021-09-07
Update JS Release Notes for v8.1.1

vonageCI:


conshus (commented):
Do you think we should mention the new `audioConstraints` object addition to params? Seems pretty significant.

conshus (approved):
Looks good.

PR #3810, opened 2021-09-21
bundle update

pardel:
## Description

Ran `bundle update` to upgrade gems to please dependabot :) 

## Deploy Notes

No changes


marcoranieri (approved):
Hello - I went throught some of those gem _changelogs_ to check any change and luckly I haven't spotted _any major change_

PR #3812, opened 2021-09-22
Minor textual changes when firing up docker

SecondeJK:
## Description

Slight textual changes to run docker a bit more efficiently for the developer

## Deploy Notes

None, readme only.


README.md
```diff
@@ -122,18 +122,32 @@ $ git clone git@github.com:Nexmo/nexmo-developer.git
...
+
+and then:
+
 ```
-git submodule init && git submodule update` and then `git config --global submodule.recurse true
+git config --global submodule.recurse true
 ```
 
-Start docker-compose:
+Run docker-compose as a daemon to run in the background:
```
pardel, line 137: @SecondeJK I would add running as a daemon as an extra option (beside the normal one) - don't think one would want to run this always in the background.

SecondeJK: @pardel I've added docs so that dev can choose what method to do

pardel: @SecondeJK Perfect! Thank you!

marcoranieri (approved):

PR #3821, opened 2021-09-29
Team updates

pardel:
## Description

Updated entries on the team page.

## Deploy Notes

No changes.


custom/landing_pages/team.yml
```diff
@@ -17,7 +17,7 @@ page:
           content: "<p class='p-large'>Our mission is to build a world-class open source documentation platform to help developers build connected products.</p>"
       - type: html
         html:
-          content: "<p class='p-large'>Here are some of the people behind Nexmo Developer. Oh, and we're hiring for <a href='#join'>the developer relations team</a> and <a href='/careers'>other teams at Vonage</a>."
+          content: "<p class='p-large'>Here are some of the people behind Vonage Developer. Oh, and we're hiring for <a href='#join'>the developer relations team</a> and <a href='/careers'>other teams at Vonage</a>."
       - type: html
```
superchilled, line 21: The `#join` fragment identifier doesn't resolve to a particular element on the page. Although it also didn't prior to this PR, it's probably worth fixing as part of these changes.

config/team.yml
```diff
@@ -1,50 +1,41 @@
...
-    role: 'Community Manager'
+    twitter: 'pardel'
   - name: 'Dwane Hemmings'
     email: 'dwane.hemmings@vonage.com'
-    twitter: 'dwaneio'
+    role: "JavaScript Developer Advocate"
+    twitter: 'lifelongdev'
   - name: 'Nick Iati'
     email: 'nick.iati@vonage.com'
     role: 'Technical Writer'
```
superchilled, line 23: Should Nick be listed as part of the Dev Rel team? If so, do we also want to add Amy and the rest of her team as part of these changes?

superchilled (changes requested):
Looks good overall, though I've noted a couple of questions/ observations.

config/team.yml
pardel: Good catch! Removed now.

custom/landing_pages/team.yml
pardel: Good point - that should have been an anchor to the jobs below. There is another Jira ticket to update other elements of that page so will include it there.

superchilled (approved):
Looks good, Paul.

PR #3829, opened 2021-10-04
VAPI region URL

vshisterov:
## Description

`region_url` description added to Webhook reference and Troubleshooting guide.


amdcavallaro (approved):
✅

PR #3899, opened 2021-12-07
[APICSI-1309] Update ios sdk build docs

JohnPeters7116:
## Description

With release 4.0.0, we have not embedded webrtc binary in the sdk anymore. If customers use cocopods, the client sdk podspec will download fine without customer needing to do anything more. But if they download the client sdk binary directly instead of using cocopods then they will need to download the VonageWebRTC framework and link it as well. 

## Deploy Notes


horothesun (approved):

conshus (approved):
Looks like everything rendered fine to me.

pardel (approved):

PR #3903, opened 2021-12-08
APIDOC-114: Update Sign-in button destination URL

VinayB09:
Update the Sign in button destination URL to https://identity.nexmo.com/login


config/business_info.yml
```diff
@@ -12,7 +12,7 @@ header:
         - get_an_api_key
         - start_building
     sign-in:
-      path: https://dashboard.nexmo.com/sign-in?icid=nexmocustomer_api-developer-adp_nexmodashbdsigin_nav
+      path: https://identity.nexmo.com/login
```
pardel, line 15: @niati @VinayB09 Do we want to remove the tracking information here?

pardel (changes requested):

config/business_info.yml
niati: @pardel you mean from dashboard.nexmo.com? If so, BI told us to add the tracking to the new URL.

niati: I see what you're saying - it appears that we didn't add it to the identity URL.

VinayB09: Added icid to the. new URL.

marcoranieri (approved):

PR #3961, opened 2022-02-11
updated gems

pardel:
## Description

Rails & co 6.1.4.1 => 6.1.4.4
rspec 3.10.0 => 3.11.0

## Deploy Notes

No changes


superchilled (commented):
@pardel just double checking why Gemfile.lock is the only file change. If we need specific minimum versions of certain gems, do we not want add constraints to that effect in the `Gemfile` itself?

pardel:
they are just updates that happen when bundle install is rerun. We don’t pin versions on station

superchilled (approved):
LGTM

PR #3994, opened 2022-03-02
Messages API documentation: WhatsApp conversation-based pricing

superchilled:
## Description

Adding information to Messages API documentation about the conversation-based pricing model now offered by Vonage for sending WhatsApp messages. Information consists of an overview and some links to further info.

## Deploy Notes

Content update only.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -538,3 +538,4 @@ Dev
 U.S.
 passcodes
 OTPs
+templated
--- a/_documentation/en/messages/concepts/whatsapp.md
+++ b/_documentation/en/messages/concepts/whatsapp.md
@@ -24,6 +24,20 @@ If your customer initiates messaging with you, WhatsApp will not charge you for 
 
 > **IMPORTANT**: The WhatsApp Business Solution may not be used to send any messages to or receive messages from the following countries or regions: Crimea, Cuba, Iran, North Korea, and Syria.
 
+## WhatsApp Conversation-Based Pricing
+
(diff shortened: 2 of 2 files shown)


smackellar (changes requested):
a 24-hour period starting from the time the first message is sent **by the business**
need to add and emphasise this as it's a point often missed

superchilled:
> a 24-hour period starting from the time the first message is sent **by the business** need to add and emphasise this as it's a point often missed

Thanks Stuart. I've made a small update to emphasise this. I'm still not sure how the information we've added inter-relates to the section above on the Customer Care Window. Do you think that needs clarifying at all, or should we just merge this?

smackellar (commented):
Small error: "with pricing model" should be "with the pricing model"

smackellar (commented):
Rules for messaging customers (to replace Important WhatsApp rules)

WhatsApp has a concept of a 24 hour customer care window, during which a business can freely message an end user. The 24 hour window can be initiated in two ways:
- An end user sends a message to the business
- A business sends a templated message to the end user. The 24 hour window starts as soon as the end user replies.
Templates must be approved by WhatsApp before they can be used to send messages to an end user. When the 24 hour window expires, a new 24 hour customer care window must be initiated again.
It is important to be aware that 24 hour customer care windows are not the same as the 24 hour billable conversation window.

abdulajet (approved):

PR #4007, opened 2022-03-09
Updates to DevTV Page

ruskibenya:
Parked devtv landing page. Hid content calendar. Added an alert bar to let users know this page is dead.

### DIFF
--- a/config/redirects.yml
+++ b/config/redirects.yml
@@ -336,3 +336,4 @@
 "/application/nexmo-cli": "/application/vonage-cli"
 "/numbers/number-pools-api/api-reference": "/api/number-pools"
 "/meetings/api-reference": "/api/meetings"
+"/devtv": "/community"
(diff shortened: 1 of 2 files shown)


abdulajet:
I think we should maybe use a redirect to another page rather than the banner it seems a bit heavy-handed? same for the other PR.

Otherwise friendlier language would be good. "To get the latest visit xyz" rather than "not maintained"

ruskibenya:
@abdulajet @marcoranieri How about this?
Screen Shot 2022-03-10 at 21 37 57

kellyjandrews:
Well - I wish I'd looked here before the other PRs 🙁

kellyjandrews:
Hey Ben - you can add redirects here I think  - https://github.com/Nexmo/nexmo-developer/blob/main/config/redirects.yml

marcoranieri (approved):

PR #4008, opened 2022-03-09
updated info on the team page

ruskibenya:
## Description

Updated team info. Added section to include our hiring locations.

### DIFF
--- a/config/team.yml
+++ b/config/team.yml
@@ -10,6 +10,7 @@ current:
   - name: 'Garann Means'
     role: "Developer Advocate"
     email: 'garann.means@vonage.com'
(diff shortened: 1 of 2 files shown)


superchilled (approved):
LGTM

abdulajet (commented):
Might wanna double-check that people who didn't include their Twitter want it there, also is there a list of countries we can hire elsewhere that is managed by HR? So we don't have to update and can just link to it.

ruskibenya:
@abdulajet Updated for twitter and new joiners. I've asked HR if there is a public link that we can change this to. In meantime this should work. I don't think the locations change too often.

PR #4046, opened 2022-03-30
revert DC trailing slash change for JS and android

abdulajet:
Only iOS suffers from this issue. Has been raised with eng as an improvement point.


devwithzachary (approved):

conshus (approved):
Looks good. Got all the trailing slashes removed.

PR #4385, opened 2022-07-20
adds documentation and code snippets

conshus:
## Description

adds documentation and code snippets for reconnecting a call and conversation's media.

## Deploy Notes


_documentation/en/client-sdk/in-app-voice/guides/reconnect-call-js.md
```diff
@@ -0,0 +1,44 @@
...
+
+The JavaScript Client SDK will automatically reconnect to a [call](https://developer.vonage.com/sdk/stitch/javascript/Application.html#reconnectCall) or a conversation's [media](https://developer.vonage.com/sdk/stitch/javascript/Media.html#enable) if a connection goes out for a couple of seconds without you needing to do anything. 
+
+The following is for the situations where a browser window/tab is accidentally closed/refreshed or the user wants to switch audio inputs and are still within the 20-second time limit to reconnect.
+
+### Considerations
+
+There are a few things that need to be in place for the reconnection to happen successfully.
+
+1. The reconnecting user must create a new JWT with their unique username to instantiate the Application Object which has the `reconnectCall` and `getConversation` methods.
```
abdulajet, line 23: I would say something like the reconnecting user must login (or I guess create session now) with a valid JWT, doesn't necessarily need to be a new one. Plus I guess the user wouldn't be creating the JWT the server will be vendoring one or something along those lines

conshus: Yeah, I spoke with Dimitris about the JWT and we decided it would be best practice to generate a new JWT just incase the original one had expired. Plus I don't really feel right about storing a JWT

abdulajet (approved):

PR #4612, opened 2022-10-03
Update install-vonage-cli.md

kasas:
@Vonage/cli is not found, we have to use it lowercased.

## Description

## Deploy Notes


pardel (approved):
@kasas Thank you!

PR #4704, opened 2022-11-14
Add throttled as a generic API Error

SecondeJK:
## Description

New API endpoints and responses we're rolling out tend to be more HATEOS / RESTful in design, to use HTTP verbs efficiently. We have now come across a number of cases where we could do with a generic Throttled API error response, as this is something likely to be standard across many products. 

This is now also the case for generic Invalid Parameters.

If this is merged, we'll need to update engineering on our specific products to make sure 429 errors link to it
i.e. `https://www.developer.vonage.com/api-errors#throttled`
or `https://www.developer.vonage.com/api-errors#invalid-params`

Note that due to the html ID elements, this does not impact product-specific links as they contain the headers in the URL.

### DIFF
--- a/config/api-errors.yml
+++ b/config/api-errors.yml
@@ -126,3 +126,7 @@ generic_errors:
   throttled:
     description: Too many requests have been made on this endpoint
     resolution: Wait a moment and try again
+
+  invalid-params:
+    description: The value of one of more parameters suppied in the request is invalid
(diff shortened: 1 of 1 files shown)


maxkahan (commented):
The description message implies that too many requests have been made "in general" to that endpoint, it doesn't mention the rate limit is the problem. Maybe change the message slightly to take the request rate into account?

maxkahan (approved):
Looks great

